PR #86, opened 2023-03-18
add Invalidations workflow

ranocha:
I added our standard workflow checking for new invalidations. We should probably wait a bit since this requires the package to be registered.


sloede (dismissed):
LGTM! I'll leave the merge up to @LasNikas or @efaulhaber

svchb (dismissed):

svchb (approved):

PR #235, opened 2023-10-05
Fix dam break with corrections

efaulhaber:
The particle spacing kwarg was not set correctly, which caused the tests to run for a very long time.
I also changed the state equation to use pressure clipping in order to make the simulations with `SummationDensity` useful.

I added an error check, which makes `trixi_include` fail when kwargs are passed that don't occur in the example file.
This makes it easier to catch errors like this, especially when an example file is changed that is `trixi_include`d somewhere else.


LasNikas (approved):
Nice! Thanks a lot

PR #427, opened 2024-03-05
Fix acknowledgement logos in readme

efaulhaber:


efaulhaber:
For some reason, the links stop working after some time (minutes/hours).
Interestingly, we get `private-user-images.githubusercontent.com` links, whereas in https://github.com/trixi-framework/Trixi.jl/issues/1390, they get `user-images.githubusercontent.com` links. Note that I specifically created a new issue after we made the repo public.

efaulhaber:
Of course, we can just use the Trixi issue to host our images as well, but we also want to host images for the docs, so it would be nice if we could fix that problem here.

sloede:
> For some reason, the links stop working after some time (minutes/hours). Interestingly, we get `private-user-images.githubusercontent.com` links, whereas in [trixi-framework/Trixi.jl#1390](https://github.com/trixi-framework/Trixi.jl/issues/1390), they get `user-images.githubusercontent.com` links. Note that I specifically created a new issue after we made the repo public.

Seems fixed in #438, or not? I now get something like `https://github-production-user-asset-6210df.s3.amazonaws.com/44124897/310171141-05132bf1-180f-4228-b30a-37dfb6e36ed5.png?``` for the hereon logo, for example

efaulhaber:
Okay, I think I know what's going on now. I think GitHub changed the way they store assets. I was looking at the first posts in https://github.com/trixi-framework/Trixi.jl/issues/1390, where the posted link is the same as the "copy image address" URL.

In the later posts, that changed to `private-user-images.githubusercontent.com`, but you guys didn't use that link, but posted the link that is added by GitHub in markdown when you paste the image. Interestingly, these are not the same:
grafik
grafik
Apparently, the `private-user-images.githubusercontent.com` URLs are only temporary.

sloede:
Ah yes, I didn't remember to tell you that I played around with this before and found that each time the picture URL is accessed, it is actually resolved to something else. You need to use the asset URL that is stored in Markdown, not the (temporary) resolved URL when linking to that image.

README.md
```diff
@@ -1,5 +1,7 @@
 # TrixiParticles.jl
 
+[![Docs-stable](https://img.shields.io/badge/docs-stable-blue.svg)](https://trixi-framework.github.io/TrixiParticles.jl/stable)
+[![Docs-dev](https://img.shields.io/badge/docs-dev-blue.svg)](https://trixi-framework.github.io/TrixiParticles.jl/dev)
```
efaulhaber, line 4: I also added the docs badges. Note that the first one doesn't work yet, but will (hopefully) work in 2 days when the package is registered.

README.md
```diff
@@ -1,13 +1,14 @@
...
+[![Slack](https://img.shields.io/badge/chat-slack-e01e5a)](https://join.slack.com/t/trixi-framework/shared_invite/zt-sgkc6ppw-6OXJqZAD5SPjBYqLd8MU~g)
+[![Youtube](https://img.shields.io/youtube/channel/views/UCpd92vU2HjjTPup-AIN0pkg?style=social)](https://www.youtube.com/@trixi-framework)
```
efaulhaber, line 6: And I moved the social badges in front of the CI.

LasNikas (approved):

svchb (approved):

PR #430, opened 2024-03-05
Add compat entry for Julia

efaulhaber:
Is 1.8 good? @sloede


sloede:
> Is 1.8 good?

I would put the version for which the package is actually tested. If you include tests for Julia v1.8, then v1.8 is ok, if you only test for v1.10, I'd put 1.10 here.

svchb:
I think we should only provide compatibility to one previous version i.e. 1.9.

LasNikas (approved):

PR #431, opened 2024-03-05
The grand reopening Part 1: Validation setup for dam break

svchb:


src/general/interpolation.jl
```diff
@@ -29,6 +30,10 @@ See also: [`interpolate_plane_2d_vtk`](@ref), [`interpolate_plane_3d`](@ref),
                       Or, in more detail, when the boundary has more influence than the fluid
                       on the density summation in this point, i.e., when the boundary particles
                       add more kernel-weighted mass than the fluid particles.
+- `clip_negative_pressure=false`: One common approach in SPH models is to clip negative pressure
+                                  values, but this leads to unphysical behavior of particles at
+                                  walls. Instead we clip here during interpolation thus only
```
efaulhaber, line 35:
```suggestion
                                  values, but this is unphysical. Instead we clip here during interpolation thus only
```

validation/dam_break_2d/plot_dam_break_results.jl
```diff
@@ -55,20 +66,23 @@ function plot_results(axs, ax_max, files)
                       normalization_factor_pressure
         pressure_P2 = json_data["pressure_P2_fluid_1"]["values"] ./
                       normalization_factor_pressure
+
+        label_prefix = occursin("reference", json_file) ? "Reference" : ""
+
         lines!(axs[1], time, pressure_P1,
-               label="dp=$(convert_to_float(split(replace(basename(json_file), ".json" => ""), "_")[end]))",
+               label="$label_prefix dp=$(convert_to_float(json_file))",
```
efaulhaber, line 73:
```suggestion
               label="$(label_prefix)dp=$(convert_to_float(json_file))",
```
Otherwise, you're adding a space in front of the non-reference labels.

validation/dam_break_2d/validation_dam_break_2d.jl
```diff
@@ -95,22 +95,26 @@ viscosity_edac = ViscosityAdami(nu=alpha * smoothing_length * sound_speed / 8)
 fluid_system_edac = EntropicallyDampedSPHSystem(tank_edac.fluid, smoothing_kernel,
                                                 smoothing_length,
                                                 sound_speed, viscosity=viscosity_edac,
+                                                density_calculator=SummationDensity(),
```
efaulhaber, line 98: I thought `ContinuityDensity` is supposed to be better, @LasNikas?

validation/dam_break_2d/validation_dam_break_2d.jl
```diff
@@ -95,22 +95,26 @@ viscosity_edac = ViscosityAdami(nu=alpha * smoothing_length * sound_speed / 8)
...
-run_file_edac_name = "out/validation_result_dam_break_edac_0015.json"
+reference_file_edac_name = "validation/dam_break_2d/validation_reference_edac_$formatted_string.json"
+run_file_edac_name = "out/validation_result_dam_break_edac_$formatted_string.json"
 
 reference_data = JSON.parsefile(reference_file_edac_name)
 run_data = JSON.parsefile(run_file_edac_name)
 
+println("error1", reference_data["pressure_P1_fluid_1"]["values"])
+println("error1", run_data["pressure_P1_fluid_1"]["values"])
+
```
efaulhaber, line 117:
```suggestion
```
?

validation/dam_break_2d/validation_dam_break_2d.jl
```diff
@@ -95,22 +95,26 @@ viscosity_edac = ViscosityAdami(nu=alpha * smoothing_length * sound_speed / 8)
 fluid_system_edac = EntropicallyDampedSPHSystem(tank_edac.fluid, smoothing_kernel,
                                                 smoothing_length,
                                                 sound_speed, viscosity=viscosity_edac,
+                                                density_calculator=SummationDensity(),
                                                 acceleration=(0.0, -gravity))
 
 trixi_include(@__MODULE__, joinpath(examples_dir(), "fluid", "dam_break_2d.jl"),
               fluid_particle_spacing=particle_spacing,
-              smoothing_length=3.5 * particle_spacing,
+              smoothing_length=smoothing_length, smoothing_kernel=smoothing_kernel,
```
efaulhaber, line 103: Can you please change the name to something other than `smoothing_length` and `smoothing_kernel`? Like `smoothing_length_` and `smoothing_kernel_` for example. Otherwise, we might run into the same problem as in the dam break corrections example, where we later add a `trixi_include` without setting smoothing length and kernel, which will then overwrite the variables defined here.

validation/dam_break_2d/plot_dam_break_results.jl
efaulhaber: Also, has to be `extract_resolution_from_filename` now.

efaulhaber (changes requested):

validation/dam_break_2d/plot_dam_break_results.jl
efaulhaber: That's probably why the tests are failing.

validation/validation_util.jl
```diff
@@ -42,3 +42,24 @@ function extract_number_from_filename(filename)
...
+    end
+
+    # Adjust string to have a decimal point at the correct position
+    if leading_zeros > 0
+        decimal_str = "0." * "0"^(leading_zeros - 1) * str_non_zero
+    else
+        decimal_str = str_non_zero
+    end
+
+    return string(parse(Float64, decimal_str))
```
efaulhaber, line 64:
```suggestion
    return decimal_str
```
Shouldn't this be the same?

validation/dam_break_2d/validation_dam_break_2d.jl
```diff
@@ -0,0 +1,158 @@
...
+
+# `particle_spacing` in this case is set relative to `H`, the initial height of the fluid.
+# Use H / 80, H / 320 for validation.
+# Note: H / 320 takes a few hours!
+particle_spacing = H / 40
+smoothing_length = 3.5 * particle_spacing
+smoothing_kernel = WendlandC2Kernel{2}()
+
+fluid_density = 1000.0
+sound_speed = 20 * sqrt(gravity * H)
```
efaulhaber, line 21: As discussed in the previous PR, this is not necessary when you just `trixi_include` the WCSPH simulation before the EDAC simulation.

efaulhaber (changes requested):
And the two unresolved comments from the closed PR.

src/schemes/fluid/entropically_damped_sph/system.jl
```diff
@@ -146,6 +146,8 @@ end
 
 @inline system_sound_speed(system::EntropicallyDampedSPHSystem) = system.sound_speed
 
+@inline system_sound_speed(system::EntropicallyDampedSPHSystem) = system.sound_speed
+
```
LasNikas, line 150:
```suggestion
```

validation/dam_break_2d/validation_dam_break_2d.jl
LasNikas: Yes, `density_calculator=ContinuityDensity()` and `pressure_acceleration=nothing`

svchb: As far as I understand the result is not better just faster. But I just didn't want to update the reference files.

validation/validation_util.jl
svchb: This handles a corner case I ran into e.g.:
julia> decimal_str = "3"
"3"

julia> string(parse(Float64, decimal_str))
"3.0"

But did do this differently before.

validation/dam_break_2d/validation_dam_break_2d.jl
svchb: Yes I did not want to discuss that there.
So what I have done before is the following if I just wanted to update one of the reference files:
1) Comment out one of the trixi_includes and run the validation setup with the resolution.
2) Commit the reference file. 
3) Wonder later what happened when this actually fails when running the validation.

So what happened?
I have run the validation setup in a REPL session after running another example, which did use a different value which was not defined... And this took me hours to figure that out since it only happened with the highest resolution reference file... 
So I am really not sure this crap is worth it to remove two lines.

validation/dam_break_2d/validation_dam_break_2d.jl
LasNikas: `SummationDensity` is more exact for closed surface and for long sim times. In the dam break case, there is almost no difference when using `ContinuityDensity` and `SummationDensity` except the  very shorter runtime. 

In steady free surface problems, `ContinuityDensity` gives better results. 

However, I didn't find something in the literature about that, since the edac method for SPH is very young. 
I could produce similar results for `ContinuityDensity` and `SummationDensity` for a closed surface problem when using a different laplace discretization formulation (see #392 ).

validation/dam_break_2d/validation_dam_break_2d.jl
efaulhaber: Good point! Please keep it, then.

validation/dam_break_2d/plot_dam_break_results.jl
efaulhaber: Not resolved?

validation/validation_util.jl
efaulhaber: Then please either remove it or add a comment explaining why this is necessary. Otherwise, it looks redundant.

validation/dam_break_2d/validation_dam_break_2d.jl
svchb: I have updated the reference files. The result is either the same or slightly worse but not better.

validation/dam_break_2d/exp_pressure_sensor_P1.csv
```diff
@@ -0,0 +1,99 @@
+time,P1
```
LasNikas, line 1: IMHO, it might be cleaner to have all the pressure sensors in one file, no?

svchb: They are manually extracted the times are not identical?

LasNikas: ah nvm. There are already multiple columns for the different heights. It would be to complicated, right?

svchb: So this would be time_P1, P1, time_P2, P2, time_P3, P3

validation/dam_break_2d/plot_dam_break_results.jl
```diff
@@ -67,17 +67,17 @@ function plot_results(axs, ax_max, files)
         pressure_P2 = json_data["pressure_P2_fluid_1"]["values"] ./
                       normalization_factor_pressure
 
-        label_prefix = occursin("reference", json_file) ? "Reference" : ""
+        label_prefix = occursin("reference", json_file) ? "Reference " : ""
 
         lines!(axs[1], time, pressure_P1,
-               label="$label_prefix dp=$(convert_to_float(json_file))",
+               label="$(label_prefix) dp=$(extract_resolution_from_filename(json_file))",
```
efaulhaber, line 73:
```suggestion
               label="$(label_prefix)dp=$(extract_resolution_from_filename(json_file))",
```

validation/validation_util.jl
```diff
@@ -61,5 +61,6 @@ function extract_resolution_from_filename(str)
         decimal_str = str_non_zero
     end
 
+    # Handle integer numbers to have consistent floats strings.
```
efaulhaber, line 64:
```suggestion
    # Convert integer strings to float strings
```
Maybe a bit easier to understand?

efaulhaber (changes requested):

validation/dam_break_2d/plot_dam_break_results.jl
```diff
@@ -0,0 +1,132 @@
...
+edac_reference_files = glob("validation_reference_edac*.json",
+                            "validation/dam_break_2d/")
+edac_sim_files = glob("validation_result_dam_break_edac*.json",
+                      "out/")
+
+merged_files = vcat(edac_reference_files, edac_sim_files)
+edac_files = sort(merged_files, by=extract_number_from_filename)
+
+wcsph_reference_files = glob("validation_reference_wcsph*.json",
+                             "validation/dam_break_2d/")
```
efaulhaber, line 29: These also need to use `joinpath` and `validation_dir`. Could that be the reason why CI is failing?

src/neighborhood_search/grid_nhs.jl
```diff
@@ -85,7 +85,7 @@ since not sorting makes our implementation a lot faster (although less paralleli
   In: Astronomy and Computing 34 (2021).
   [doi: 10.1016/j.ascom.2020.100443](https://doi.org/10.1016/j.ascom.2020.100443)
 - Markus Ihmsen, Nadir Akinci, Markus Becker, Matthias Teschner.
-  "A Parallel SPH Implementation on Multi-Core CPUs".
+  "A threaded_nhs_update SPH Implementation on Multi-Core CPUs".
```
efaulhaber, line 88: .

test/validation/validation.jl
```diff
@@ -35,25 +36,34 @@
...
+        @test isapprox(error_edac_P1, 0, atol=1e-9)
+        @test isapprox(error_edac_P2, 0, atol=1e-11)
+
+        if VERSION >= v"1.10"
+            @test isapprox(error_wcsph_P1, 0, atol=eps())
+            @test isapprox(error_wcsph_P2, 0, atol=eps())
+        else
+            # 1.9 causes a large difference in the solution
+            @test isapprox(error_wcsph_P1, 0, atol=10)
+            @test isapprox(error_wcsph_P2, 0, atol=1e-3)
```
efaulhaber, line 54: What would be the corresponding `rtol`? It's probably more reasonable.

validation/dam_break_2d/validation_dam_break_2d.jl
```diff
@@ -105,7 +105,7 @@ trixi_include(@__MODULE__, joinpath(examples_dir(), "fluid", "dam_break_2d.jl"),
               boundary_layers=4, state_equation=nothing,
               solution_prefix="validation_" * method * "_" * formatted_string,
               extra_callback=postprocessing_cb, tspan=tspan,
-              fluid_system=fluid_system_edac, tank=tank_edac)
+              fluid_system=fluid_system_edac, tank=tank_edac, threaded_grid_search=false)
```
efaulhaber, line 108:
```suggestion
              fluid_system=fluid_system_edac, tank=tank_edac,
              threaded_nhs_update=false) # To get the same results with different thread numbers
```
Same below.

efaulhaber (changes requested):

efaulhaber (approved):

efaulhaber (approved):

Issue #133, opened 2023-05-08
Bikeshedding: Error messages and comments


efaulhaber:
I'm sorry for the major bikeshedding, but let's settle this once and for all to avoid future bikeshedding in PRs.

How do we want error messages to look like?
1. Should the first word be capitalized or not?
2. Punctuation: When the message is just one sentence or a sentence fragment, should we end it with a period or exclamation mark, or no punctuation?

In Julia, sentence fragments starting with lowercase and no punctuation seem to be preferred:
![grafik](https://user-images.githubusercontent.com/44124897/236798532-82f9d54f-6787-4715-a034-91094190aeb4.png)
```
julia> Dict{1,2 ,3 }
ERROR: too many parameters for type

julia> if 4
       end
ERROR: TypeError: non-boolean (Int64) used in boolean context

julia> sqrt(-1)
ERROR: DomainError with -1.0:
sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
Stacktrace:
 [1-2] ⋮ internal
     @ Base.Math, Unknown
   [3] sqrt(x::Int64)
     @ Base.Math ./math.jl:1372
Use `err` to retrieve the full stack trace.

julia> f()
ERROR: UndefVarError: f not defined

julia> Int(1e100)
ERROR: InexactError: Int64(1.0e100)
Stacktrace:
   [1] Int64(x::Float64)
     @ Base ./float.jl:788

julia> x = []; x[1]
ERROR: BoundsError: attempt to access 0-element Vector{Any} at index [1]
Stacktrace:
   [1] getindex(A::Vector{Any}, i1::Int64)
     @ Base ./array.jl:924
```
But not consistent outside of base:
```
julia> using sfdljoisdf
ERROR: ArgumentError: Package sfdljoisdf not found in current path.
- Run `import Pkg; Pkg.add("sfdljoisdf")` to install the sfdljoisdf package.
```

Same for comments, while we're at it:
1. Should comments start with a lowercase letter?
2. Should sentence fragments end with a period?
3. Should full sentences end with a period? (Of course, when the comment consists of multiple sentences, we end all of them with a period.)

My personal opinion (when I say no opinion, I still want us to pick one and use it consistently):
- No opinion about starting error messages with lower or upper case.
- No opinion about ending without punctuation or with a period, but I am strongly against exclamation marks (too aggressive).
- I prefer starting comments with an uppercase letter.
[...]

PR #434, opened 2024-03-05
The Grand ReReReReReopening the final straw part 9: Intra-Particle-Force Surface Tension model

svchb:


svchb:
Add to news.md and add Docs.

src/setups/rectangular_tank.jl
```diff
@@ -149,27 +149,33 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real}
...
+            else
+                fluid = RectangularShape(particle_spacing, n_particles_per_dim,
+                                         zeros(NDIMS);
+                                         density=fluid_density, velocity, pressure,
+                                         acceleration, state_equation, mass=fluid_mass)
+            end
+            # Move the tank corner in the negative coordinate directions to the desired position
+            fluid.coordinates .+= min_coordinates
+        end
+
```
efaulhaber, line 178: What is happening here? If the fluid size is zero, return the boundary twice? That's awkward.

efaulhaber (changes requested):
First part of my review. Gotta go now. I'll have a look at the example files and `surface_tension.jl` later.

Also, can you please check that all comments follow #133?

docs/src/index.md
```diff
@@ -5,6 +5,7 @@ TrixiParticles.jl is a numerical simulation framework designed for particle-base
 ## Features
 - Incompressible Navier-Stokes
   - Methods: Weakly Compressible Smoothed Particle Hydrodynamics (WCSPH), Entropically Damped Artificial Compressibility (EDAC)
+  - Models: Surface Tension
```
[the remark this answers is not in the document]
svchb: Maybe we should list these under different categories since Density Diffusion or TVF don't model physical effects?

examples/fluid/dam_break_2d_surface_tension.jl
```diff
@@ -0,0 +1,38 @@
...
+
+# Set the surface tension to a value that is accurate in your case.
+# Note: This usually requires calibration to be physically accurate!
+surface_tension = SurfaceTensionAkinci(surface_tension_coefficient=0.25)
+
+# `density_diffusion` is deactivated since the interaction with the surface tension model can
+# cause stability problems.
+# `adhesion_coefficient` needs to be set to a value so that the fluid doesn't separate
+# from the boundary
+# Note: The viscosity will be increased by the surface tension model hence we can reduce the artificial viscosity value.
```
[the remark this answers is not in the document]
svchb: Actually it is complicated one would actually need to do a parameter study to find the correct combination of viscosity and surface tension at a given resolution and kernel values to fit a physical measurement.

examples/fluid/dam_break_2d_surface_tension.jl
```diff
@@ -0,0 +1,38 @@
...
+# Note: The surface tension model leds to an increase in compressibility of the fluid
+#       which needs to be rectified by an increase of the `sound_speed`.
+# Note: The Wendland Kernels don't work very well here since the SurfaceTensionAkinci
+#       model is optimized for smoothing_length=4r which is a too small smoothing_length for the Wendland Kernel functions.
```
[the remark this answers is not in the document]
svchb: Because the compact_support has a factor of 2?

src/schemes/boundary/system.jl
```diff
@@ -10,18 +10,25 @@ The interaction between fluid and boundary particles is specified by the boundar
 
 # Keyword Arguments
 - `movement`: For moving boundaries, a [`BoundaryMovement`](@ref) can be passed.
+- `adhesion_coefficient`: Coefficient specifying the adhesion of a fluid to the surface.
+   Note: currently it is assumed that all fluids have the same adhesion coefficient.
```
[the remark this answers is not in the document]
svchb: The problem is that you might want to set different adhesion coefficients for different surfaces but you've got only one fluid?

src/setups/rectangular_tank.jl
svchb: I use the tank without a fluid inside.

src/schemes/boundary/system.jl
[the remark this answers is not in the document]
svchb: The note is regarding two fluids for example water and oil which have different contact angles.

src/schemes/fluid/weakly_compressible_sph/rhs.jl
```diff
@@ -25,9 +28,9 @@ function interact!(dv, v_particle_system, u_particle_system,
         rho_mean = 0.5 * (rho_a + rho_b)
 
         # Determine correction values
-        viscosity_correction, pressure_correction = free_surface_correction(correction,
-                                                                            particle_system,
-                                                                            rho_mean)
+        viscosity_correction, pressure_correction, surface_tension_correction = free_surface_correction(correction,
+                                                                                                        particle_system,
+                                                                                                        rho_mean)
```
LasNikas, line 33:
```suggestion
        corrections = free_surface_correction(correction, particle_system, rho_mean)
```
Might be more readable.
And then refer to the correcions:
```julia

        corrections[1] = correction_viscosity
        corrections[2] = correction_pressure
        corrections[3] = correction_surface_tension 
```

src/schemes/fluid/weakly_compressible_sph/rhs.jl
```diff
@@ -153,3 +165,82 @@ end
...
+                                      surface_tension_a::CohesionForceAkinci,
+                                      surface_tension_b::CohesionForceAkinci)
+    (; smoothing_length) = particle_system
+
+    # no cohesion with oneself
+    if distance < sqrt(eps())
+        return zeros(SVector{ndims(particle_system), eltype(particle_system)})
+    end
+
+    m_b = neighbor_system.mass[neighbor]
```
LasNikas, line 181: use `hydrodynamic_mass` for consistency

src/schemes/fluid/surface_tension.jl
```diff
@@ -0,0 +1,193 @@
...
+    @threaded for particle in each_moving_particle(system)
+        particle_coords = current_coords(u_system, system, particle)
+
+        for neighbor in eachneighbor(particle_coords, neighborhood_search)
+            neighbor_coords = current_coords(u_system, system, neighbor)
+
+            pos_diff = particle_coords - neighbor_coords
+            distance = norm(pos_diff)
+            # correctness strongly depends on this being a symmetric distribution of particles
+            if sqrt(eps()) < distance <= compact_support(smoothing_kernel, smoothing_length)
```
LasNikas, line 169:
```suggestion
            distance2 = dot(pos_diff)
            # correctness strongly depends on this being a symmetric distribution of particles
            if eps() < distance2 <= compact_support(smoothing_kernel, smoothing_length)^2
            distance = sqrt(distance2)
```

maybe a small performance gain?

examples/fluid/dam_break_2d_surface_tension.jl
```diff
@@ -0,0 +1,38 @@
...
+              correction=AkinciFreeSurfaceCorrection(fluid_density),
+              density_diffusion=nothing, adhesion_coefficient=0.5, alpha=0.001,
+              sound_speed=100, tspan=(0.0, 2.0))
+
+# sol = solve(ode, RDPK3SpFSAL35(),
+#             abstol=1e-5, # Default abstol is 1e-6 (may need to be tuned to prevent boundary penetration)
+#             reltol=1e-4, # Default reltol is 1e-3 (may need to be tuned to prevent boundary penetration)
+#             dtmax=1e-2, # Limit stepsize to prevent crashing
+#             dt=1e-5,
+#             save_everystep=false, callback=callbacks);
```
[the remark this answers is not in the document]
svchb: Left from experimentation.

examples/fluid/dam_break_2d_surface_tension.jl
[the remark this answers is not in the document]
svchb: Particle spacing might be more than 2*r in some cases.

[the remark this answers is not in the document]
efaulhaber: So it's not optimized for a smoothing length of 4r, but a compact support of 4r?
> Particle spacing might be more than 2*r in some cases.

The initial particle spacing is per definition equal to 2r, isn't it?

src/setups/rectangular_tank.jl
efaulhaber: But that's a really awkward way to do that. Can't you just return an empty `InitialCondition` for the fluid in this case?

src/schemes/fluid/weakly_compressible_sph/system.jl
```diff
@@ -327,3 +341,39 @@ end
...
+    # reset surface normal
+    cache.surface_normal .= zero(eltype(cache.surface_normal))
+
+    @trixi_timeit timer() "compute surface normal" foreach_system(semi) do neighbor_system
+        u_neighbor_system = wrap_u(u_ode, neighbor_system, semi)
+        v_neighbor_system = wrap_v(v_ode, neighbor_system, semi)
+        nhs = get_neighborhood_search(system, semi)
+
+        calc_normal_akinci!(surface_tension, u, v_neighbor_system, u_neighbor_system, nhs,
+                            system, neighbor_system)
```
efaulhaber, line 362: This is not correct anymore.

src/schemes/fluid/weakly_compressible_sph/rhs.jl
efaulhaber: The current version is more Julian.

examples/fluid/dam_break_2d_surface_tension.jl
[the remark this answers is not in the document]
efaulhaber: Then I suggest:
```
# Note: The Wendland Kernels don't work very well here since the `SurfaceTensionAkinci`
#       model is optimized for a compact support of `2 * particle_spacing`, which would result
#       in a too small `smoothing_length` for the Wendland Kernel functions.
```

src/schemes/fluid/weakly_compressible_sph/system.jl
```diff
@@ -327,3 +341,39 @@ end
...
+        v_neighbor_system = wrap_v(v_ode, neighbor_system, semi)
+        nhs = get_neighborhood_search(system, semi)
+
+        calc_normal_akinci!(system, surface_tension, u, v_neighbor_system, u_neighbor_system, nhs,
+                            neighbor_system)
+    end
+    return system
+end
+
+@inline function get_normal(particle, particle_system::FluidSystem,
```
efaulhaber, line 367: We try to avoid `get_` functions since #68.
```suggestion
@inline function surface_normal(particle, particle_system::FluidSystem,
```

efaulhaber (changes requested):
Second part of my review. Example files and tests remain.

src/schemes/fluid/weakly_compressible_sph/system.jl
svchb: Yes this was of course #61

src/schemes/fluid/surface_tension.jl
```diff
@@ -0,0 +1,196 @@
...
+        particle_coords = current_coords(u_system, system, particle)
+
+        for neighbor in eachneighbor(particle_coords, neighborhood_search)
+            neighbor_coords = current_coords(u_system, system, neighbor)
+
+            pos_diff = particle_coords - neighbor_coords
+            distance2 = dot(pos_diff, pos_diff)
+            # Correctness strongly depends on this being a symmetric distribution of particles
+            if eps() < distance2 <= compact_support(smoothing_kernel, smoothing_length)^2
+                distance = sqrt(distance2)
```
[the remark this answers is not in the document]
svchb: Yes I replied in the discussion from a month ago.

src/schemes/fluid/surface_tension.jl
```diff
@@ -0,0 +1,196 @@
...
+end
+
+@fastpow @inline function cohesion_force_akinci(surface_tension::AkinciTypeSurfaceTension,
+                                                support_radius, mb, pos_diff, distance)
+    (; surface_tension_coefficient) = surface_tension
+
+    # Eq. 2
+    # We only reach this function when distance > eps
+    C = 0
+    if distance <= support_radius
```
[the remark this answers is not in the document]
svchb: This is how that was written in the paper. Anyway I tried to get around the limitations of the support radius by modifying the radius that is used for the surface tension model but that never really worked.

src/schemes/fluid/surface_tension.jl
```diff
@@ -0,0 +1,148 @@
...
+struct CohesionForceAkinci{ELTYPE} <: AkinciTypeSurfaceTension
+    surface_tension_coefficient::ELTYPE
+
+    function CohesionForceAkinci(; surface_tension_coefficient=1.0)
+        new{typeof(surface_tension_coefficient)}(surface_tension_coefficient)
+    end
+end
+
+@doc raw"""
+SurfaceTensionAkinci(surface_tension_coefficient=1.0)
```
efaulhaber, line 30:
```suggestion
    SurfaceTensionAkinci(surface_tension_coefficient=1.0)
```

docs/src/systems/weakly_compressible_sph.md
```diff
@@ -150,6 +150,44 @@ Pages = [joinpath("general", "corrections.jl")]
 
 
 ## [Surface Tension](@id surface_tension)
+
+### Akinci based intra-particle force surface tension and wall adhesion model
+The work by Akinci proposes three forces:
+- a cohesion force
+- a surface normal force
```
efaulhaber, line 157:
```suggestion
### Akinci-based intra-particle force surface tension and wall adhesion model
The work by Akinci proposes three forces:
- a cohesion force
- a surface normal force
```
And shouldn't "surface normal force" be "curvature minimization force"? I find it confusing that you list three forces here and explain a different force below.

docs/src/systems/weakly_compressible_sph.md
```diff
@@ -150,6 +150,44 @@ Pages = [joinpath("general", "corrections.jl")]
...
+
+#### Curvature Minimization Force
+To model the minimization of the surface curvature of the fluid a curvature force is used which is calculated
+```math
+F_{\text{surf_min}} = -\sigma (n_a - n_b)
+```
+
+#### Wall Adhesion Force
+The wall adhesion model proposed by Akinci et al. is based on a kernel function which is 0 from 0.0 to 0.5 support radiia with a maximum at 0.75.
+With the force calculated with a adhesion coefficient ``\beta``:
```
efaulhaber, line 186:
```suggestion
With the force calculated with an adhesion coefficient ``\beta`` as 
```

docs/src/systems/weakly_compressible_sph.md
```diff
@@ -150,6 +150,44 @@ Pages = [joinpath("general", "corrections.jl")]
...
+F_{\text{cohesion}} = -\sigma m_b C \frac{r}{\Vert r \Vert},
+```
+where:
+- ``\sigma`` represents the surface tension coefficient, adjusting the overall strength of the cohesion effect.
+- ``C`` is a scalar function of the distance between particles.
+
+#### Curvature Minimization Force
+To model the minimization of the surface curvature of the fluid a curvature force is used which is calculated
+```math
+F_{\text{surf_min}} = -\sigma (n_a - n_b)
```
efaulhaber, line 181:
```suggestion
F_{\text{surf\_min}} = -\sigma (n_a - n_b)
```
Doesn't render.

src/schemes/fluid/surface_tension.jl
```diff
@@ -10,31 +10,12 @@ and the merging or breaking of fluid bodies.
...
-
-```math
-F_{\text{cohesion}} = -\sigma m_b C \frac{\vec{r}}{|\vec{r}|}
-```
-where:
-- σ represents the surface_tension_coefficient, adjusting the overall strength of the cohesion effect.
-- C is a scalar function of the distance between particles.
+   enabling the tuning of the fluid's surface tension properties within the simulation.
 
 # Reference:
```
efaulhaber, line 15:
```suggestion
# References
```
Same below.

src/schemes/fluid/surface_tension.jl
```diff
@@ -93,49 +53,53 @@ struct SurfaceTensionAkinci{ELTYPE} <: AkinciTypeSurfaceTension
...
+# for enhanced computational speed. This is especially useful in scenarios where performance
+# is a higher priority than exact precision.
+@fastpow @inline function cohesion_force_akinci(surface_tension, support_radius, m_b,
+                                                pos_diff, distance)
     (; surface_tension_coefficient) = surface_tension
 
     # Eq. 2
-    # We only reach this function when distance > eps
+    # We only reach this function when distance > eps and `distance < support_radius`
     C = 0
```
efaulhaber, line 82:
```suggestion
    # We only reach this function when `sqrt(eps()) < distance < support_radius`
```
C is redundant now.

src/schemes/fluid/surface_tension.jl
```diff
@@ -93,49 +53,53 @@ struct SurfaceTensionAkinci{ELTYPE} <: AkinciTypeSurfaceTension
...
 end
 
-@inline function adhesion_force_akinci(surface_tension::AkinciTypeSurfaceTension,
-                                       support_radius, mb, pos_diff, distance,
-                                       adhesion_coefficient)
+@inline function adhesion_force_akinci(surface_tension, support_radius, m_b, pos_diff,
+                                       distance, adhesion_coefficient)
     # Eq. 7
-    # We only reach this function when distance > eps
+    # We only reach this function when `distance > eps`
```
efaulhaber, line 102:
```suggestion
    # We only reach this function when `sqrt(eps()) < distance < support_radius`
```
And the if condition below is not needed anymore.

src/schemes/fluid/surface_tension.jl
```diff
@@ -144,8 +108,9 @@ end
         end
     end
 
-    # Eq. 6 in acceleration form with mb being the boundary mass calculated as mb=rho_0 * volume (Akinci boundary condition treatment)
-    adhesion_force = -adhesion_coefficient * mb * A * pos_diff / distance
+    # Eq. 6 in acceleration form with `m_b`` being the boundary mass calculated as
+    # `m_b=rho_0 * volume`` (Akinci boundary condition treatment)
```
efaulhaber, line 112:
```suggestion
    # Eq. 6 in acceleration form with `m_b` being the boundary mass calculated as
    # `m_b = rho_0 * volume` (Akinci boundary condition treatment)
```

src/schemes/fluid/weakly_compressible_sph/system.jl
```diff
@@ -358,14 +358,15 @@ function compute_surface_normal!(system, surface_tension::SurfaceTensionAkinci,
...
+                            u_neighbor_system, nhs,
                             neighbor_system)
     end
     return system
 end
 
-@inline function get_normal(particle, particle_system::FluidSystem,
-                            ::SurfaceTensionAkinci)
+@inline function surface_normal(particle, particle_system::FluidSystem,
+                                ::SurfaceTensionAkinci)
```
efaulhaber, line 369:
```suggestion
@inline function surface_normal(::SurfaceTensionAkinci, particle_system::FluidSystem,
                                particle)
```
Consistency with other functions.

src/schemes/fluid/surface_tension.jl
```diff
@@ -93,49 +53,53 @@ struct SurfaceTensionAkinci{ELTYPE} <: AkinciTypeSurfaceTension
...
+#                                                                              distance)
+#     return cohesion_force_akinci(surface_tension, smoothing_length, m_b, pos_diff, distance)
+# end
+
+# function (surface_tension::SurfaceTensionAkinci)(support_radius, m_b, na, nb, pos_diff,
+#                                                  distance)
+#     (; surface_tension_coefficient) = surface_tension
+#     return cohesion_force_akinci(surface_tension, support_radius, m_b, pos_diff,
+#                                  distance) .- (surface_tension_coefficient * (na - nb))
+# end
```
efaulhaber, line 67: What about this comment?

src/schemes/fluid/weakly_compressible_sph/rhs.jl
```diff
@@ -153,3 +165,70 @@ end
 
     return p_a, p_a
 end
+
+@inline function calc_surface_tension(particle, neighbor, pos_diff, distance,
+                                      particle_system::FluidSystem,
+                                      neighbor_system::FluidSystem,
+                                      surface_tension_a::CohesionForceAkinci,
+                                      surface_tension_b::CohesionForceAkinci)
```
efaulhaber, line 173:
```suggestion
@inline function surface_tension_force(surface_tension_a::CohesionForceAkinci,
                                      surface_tension_b::CohesionForceAkinci,
                                      particle_system::FluidSystem,
                                      neighbor_system::FluidSystem,
                                      particle, neighbor, pos_diff, distance)
```
1. I think we only used `calc_` for functions that compute something and write it into some array, while this here returns a force.
2. I think this parameter order would be most consistent with the rest of TrixiP.
3. Do you think it would be easier for new developers to find these functions when they're inside `surface_tension.jl` instead of here?
4. Same for `calc_adhesion` below.

efaulhaber (changes requested):

src/schemes/fluid/surface_tension.jl
svchb: That is not correct since the value of this function is not defined for distance=support_radius. The comments are inaccurate its actually `<=`.

src/schemes/fluid/surface_tension.jl
```diff
@@ -146,3 +130,69 @@ function calc_normal_akinci!(system, neighbor_system, surface_tension, u_system,
...
 end
+
+@inline function surface_tension_force(surface_tension_a::CohesionForceAkinci,
+                                       surface_tension_b::CohesionForceAkinci,
+                                       particle_system::FluidSystem,
+                                       neighbor_system::FluidSystem, particle, neighbor,
+                                       pos_diff, distance)
+    (; smoothing_length) = particle_system
+    # No cohesion with oneself
+    distance < sqrt(eps()) && return zero(pos_diff)
```
efaulhaber, line 141:
```suggestion
    distance < sqrt(eps()) && return zero(pos_diff)
    
```

src/schemes/fluid/surface_tension.jl
```diff
@@ -146,3 +130,69 @@ function calc_normal_akinci!(system, neighbor_system, surface_tension, u_system,
...
+@inline function surface_tension_force(surface_tension_a::SurfaceTensionAkinci,
+                                       surface_tension_b::SurfaceTensionAkinci,
+                                       particle_system::FluidSystem,
+                                       neighbor_system::FluidSystem, particle, neighbor,
+                                       pos_diff, distance)
+    (; smoothing_length, smoothing_kernel) = particle_system
+    (; surface_tension_coefficient) = surface_tension_a
+
+    # No surface tension with oneself
+    distance < sqrt(eps()) && return zero(pos_diff)
```
efaulhaber, line 157:
```suggestion
    distance < sqrt(eps()) && return zero(pos_diff)
    
```
A little easier to read IMHO.

src/schemes/fluid/surface_tension.jl
svchb: yes but the part in the if becomes negative and so the result of the ^0.25 is a complex number.

efaulhaber: Why negative? It only becomes zero, no?

svchb: Yes the result should be 0 instead of the complex number

efaulhaber: The part inside the root becomes zero, and `0.0^0.25` is `0.0`. Or am I missing something?

svchb: Yes it should be but it doesn't. The simulations crashes because the value is negative if we remove the special handling.

docs/src/systems/weakly_compressible_sph.md
```diff
@@ -175,17 +175,35 @@ where:
...
+\end{cases}
+```
+
+#### Surface area minimization force
+To model the minimization of the surface area and curvature of the fluid a curvature force is used which is calculated
+```math
+F_{\text{curvature}} = -\sigma (n_a - n_b)
 ```
 
 #### Wall Adhesion Force
```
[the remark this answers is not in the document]
efaulhaber: Now it's `Wall adhesion Force`.

docs/src/systems/weakly_compressible_sph.md
```diff
@@ -147,3 +147,66 @@ Pages = [joinpath("schemes", "fluid", "weakly_compressible_sph", "density_diffus
...
+
+#### Wall Adhesion Force
+The wall adhesion model proposed by Akinci et al. is based on a kernel function which is 0 from 0.0 to 0.5 support radiia with a maximum at 0.75.
+With the force calculated with an adhesion coefficient ``\beta`` as 
+```math
+F_{\text{adhesion}} = -\beta m_b A \frac{r}{\Vert r \Vert},
+```
+with ``A`` being the adhesion kernel defined as
+```math
+\mathbf{A}(r)= \frac{0.007}{h^{3.25}}
```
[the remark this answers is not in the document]
efaulhaber: Still bold in the second equation.

docs/src/systems/weakly_compressible_sph.md
```diff
@@ -147,3 +147,66 @@ Pages = [joinpath("schemes", "fluid", "weakly_compressible_sph", "density_diffus
...
+```math
+F_{\text{cohesion}} = -\sigma m_b C \frac{r}{\Vert r \Vert},
+```
+where:
+- ``\sigma`` represents the surface tension coefficient, adjusting the overall strength of the cohesion effect.
+- ``C`` is a scalar function of the distance between particles.
+
+The Cohesion Kernel ``C`` is defined as 
+```math
+\mathbf{C(r)}=\frac{32}{\pi h^9}
```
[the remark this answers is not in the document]
efaulhaber: Still capitalized and bold.

test/examples/examples.jl
```diff
@@ -114,6 +114,63 @@
...
+            ]
+            @test sol.retcode == ReturnCode.Success
+            @test count_rhs_allocations(sol, semi) == 0
+        end
+
+        @trixi_testset "fluid/wetting_water_spheres_2d.jl" begin
+            @test_nowarn_mod trixi_include(@__MODULE__,
+                                           joinpath(examples_dir(), "fluid",
+                                                    "wetting_water_spheres_2d.jl"),
+                                           tspan=(0.0, 0.1))
```
efaulhaber, line 171: Since these example can easily be unstable with different resolutions, could you please include the exact file here without changing parameters, so that we can guarantee that the examples are working?
As I said earlier, the examples should be made cheaper anyway.

examples/fluid/wetting_water_spheres_2d.jl
```diff
@@ -0,0 +1,105 @@
+# In this example two circles of water drop to the floor demonstrating the difference
+# between the behavior with and without surface tension modelling.
```
efaulhaber, line 2: What is this example? This comment is the same as for falling spheres 3D. Are they just using different parameters? Why do we need these separate examples?

test/schemes/fluid/surface_tension.jl
```diff
@@ -0,0 +1,88 @@
+
+@testset verbose=true "surface tension" begin
+    @testset verbose=true "cohesion_force_akinci" begin
```
efaulhaber, line 3:
```suggestion
@testset verbose=true "Surface Tension" begin
    @testset verbose=true "`cohesion_force_akinci`" begin
```

test/schemes/fluid/surface_tension.jl
```diff
@@ -0,0 +1,88 @@
...
+        surface_tension = TrixiParticles.SurfaceTensionAkinci(surface_tension_coefficient=1.0)
```
efaulhaber, line 4:
```suggestion
        surface_tension = TrixiParticles.SurfaceTensionAkinci(surface_tension_coefficient=1.0)
```
```suggestion
        surface_tension = SurfaceTensionAkinci(surface_tension_coefficient=1.0)
```

test/schemes/fluid/surface_tension.jl
```diff
@@ -0,0 +1,88 @@
...
+        support_radius = 1.0
+        m_b = 1.0
+        pos_diff = [1.0, 1.0]
+
+        test_distance = 0.1
+        val = TrixiParticles.cohesion_force_akinci(surface_tension, support_radius, m_b,
+                                                   pos_diff, test_distance) * test_distance
+        @test isapprox(val[1], 0.1443038770421044, atol=6e-15)
+        @test isapprox(val[2], 0.1443038770421044, atol=6e-15)
```
efaulhaber, line 13: Where do these values come from? Could you please add a comment why these are the correct values? Same below.

test/schemes/fluid/surface_tension.jl
```diff
@@ -0,0 +1,88 @@
...
+
+        # maximum repulsion force
```
efaulhaber, line 15:
```suggestion
        # Maximum repulsion force
```
Please also check the other comments.

test/schemes/fluid/surface_tension.jl
```diff
@@ -0,0 +1,88 @@
...
+        test_distance = 0.01
+        max = TrixiParticles.cohesion_force_akinci(surface_tension, support_radius, m_b,
+                                                   pos_diff, test_distance) * test_distance
+        @test isapprox(max[1], 0.15913517632298307, atol=6e-15)
+        @test isapprox(max[2], 0.15913517632298307, atol=6e-15)
+
+        # near 0
+        test_distance = 0.2725
```
efaulhaber, line 23: Why is this supposed to be near zero? Please add a comment.

test/schemes/fluid/surface_tension.jl
```diff
@@ -0,0 +1,88 @@
...
+        # should be 0
+        test_distance = 1.0
+        zero = TrixiParticles.cohesion_force_akinci(surface_tension, support_radius, m_b,
+                                                    pos_diff, test_distance) * test_distance
+        @test isapprox(zero[1], 0.0, atol=6e-15)
+        @test isapprox(zero[2], 0.0, atol=6e-15)
+    end
+
+    @testset verbose=true "adhesion_force_akinci" begin
+        surface_tension = TrixiParticles.SurfaceTensionAkinci(surface_tension_coefficient=1.0)
```
efaulhaber, line 45:
```suggestion
    @testset verbose=true "adhesion_force_akinci" begin
        surface_tension = TrixiParticles.SurfaceTensionAkinci(surface_tension_coefficient=1.0)
```
```suggestion
    @testset verbose=true "`adhesion_force_akinci`" begin
        surface_tension = SurfaceTensionAkinci(surface_tension_coefficient=1.0)
```
This part also needs a lot more comments. Why these values? Why is it supposed to be zero? And so on.

efaulhaber (changes requested):

examples/fluid/sphere_surface_tension_2d.jl
```diff
@@ -1,28 +1,28 @@
 # In this example we can observe that the `SurfaceTensionAkinci` surface tension model correctly leads to a
-# surface minimization of the water square and approaches a circle.
+# surface minimization of the water square and approaches a sphere.
 using TrixiParticles
 using OrdinaryDiffEq
 
 fluid_density = 1000.0
 
 particle_spacing = 0.1
+square_size = 0.5
```
efaulhaber, line 9:
```suggestion
fluid_size = (0.5, 0.5)
```
More consistent with the other examples. Less confusing for users to work with `trixi_include`. Especially when using the 3D example as well.

examples/fluid/sphere_surface_tension_2d.jl
```diff
@@ -1,28 +1,28 @@
...
-# nu = 0.025
-
-fluid = RectangularShape(particle_spacing, (5, 5), (0.0, 0.0),
+# smoothing_kernel = WendlandC2Kernel{2}()
+# nu = 0.01
+
+smoothing_length = 1.0 * particle_spacing
+smoothing_kernel = SchoenbergCubicSplineKernel{2}()
+nu = 0.025
+no_square_particles = round(Int, square_size / particle_spacing)
```
efaulhaber, line 23:
```suggestion

n_particles_per_dimension = round.(Int, fluid_size ./ particle_spacing)
```
Same variable name as used internally.

examples/fluid/sphere_surface_tension_2d.jl
```diff
@@ -46,7 +46,10 @@ tspan = (0.0, 5.0)
 ode = semidiscretize(semi, tspan)
 
 info_callback = InfoCallback(interval=100)
-saving_callback = SolutionSavingCallback(dt=0.02)
+
+# For overwriting via trixi_include
+saving_interval = 0.02
+saving_callback = SolutionSavingCallback(dt=saving_interval)
```
efaulhaber, line 52: Why? One can just set `dt=0.01`. The `dt` below for the initial time step will be overwritten too, but it's irrelevant.

examples/fluid/sphere_surface_tension_3d.jl
```diff
@@ -0,0 +1,30 @@
...
+# surface minimization of the water cube and approaches a sphere.
+using TrixiParticles
+using OrdinaryDiffEq
+
+fluid_density = 1000.0
+
+particle_spacing = 0.1
+cube_size = 0.9
+
+sound_speed = 20
```
efaulhaber, line 11:
```suggestion
sound_speed = 20.0
```
Do we need this here? It's the same as in 2D.

examples/fluid/sphere_surface_tension_3d.jl
```diff
@@ -0,0 +1,30 @@
...
+
+# For all surface tension simulations, we need a compact support of `2 * particle_spacing`
+smoothing_length = 1.0 * particle_spacing
+
+no_cube_particles = round(Int, cube_size / particle_spacing)
+fluid = RectangularShape(particle_spacing,
+                         (no_cube_particles, no_cube_particles, no_cube_particles),
+                         (0.0, 0.0, 0.0), density=fluid_density)
```
efaulhaber, line 19: Not necessary when using `fluid_size` instead.

examples/fluid/wetting_water_spheres_2d.jl
```diff
@@ -1,13 +1,13 @@
-# In this example two circles of water drop to the floor demonstrating the difference
-# between the behavior with and without surface tension modelling.
+# In this example we try to approach the static shape of a water droplet on a horizontal plane.
+# The shape of a static droplet can be calculated from the Young-Laplace equation.
```
efaulhaber, line 2: Can you `trixi_include` the falling spheres example? They look very similar to me.

svchb:
353 commits now :sob:

efaulhaber (changes requested):

examples/fluid/sphere_surface_tension_3d.jl
svchb: needed in the calculation of alpha

examples/fluid/sphere_surface_tension_2d.jl
svchb: yes but this is in no way logical that dt changes the output.

efaulhaber: Well, none of this is documented anywhere, so one has to look into the example file anyway. And it's `dt` for all other example files.

examples/fluid/falling_water_spheres_2d.jl
```diff
@@ -0,0 +1,94 @@
+# In this example two circles of water drop to the floor demonstrating the difference
```
efaulhaber, line 1: 1. This example is unstable when I run it on multiple threads. It just continues with a time step < eps.
2. I don't see a difference between surface tension and no surface tension:
grafik

examples/fluid/sphere_surface_tension_2d.jl
```diff
@@ -0,0 +1,61 @@
...
+                         zeros(length(fluid_size)), density=fluid_density)
+
+alpha = 8 * nu / (smoothing_length * sound_speed)
+source_terms = SourceTermDamping(; damping_coefficient=0.5)
+fluid_system = WeaklyCompressibleSPHSystem(fluid, SummationDensity(),
+                                           state_equation, smoothing_kernel,
+                                           smoothing_length,
+                                           viscosity=ArtificialViscosityMonaghan(alpha=alpha,
+                                                                                 beta=0.0),
+                                           acceleration=(0.0, 0.0),
```
efaulhaber, line 37:
```suggestion
```
Then you also don't need to set this in the 3D case.

examples/fluid/sphere_surface_tension_3d.jl
efaulhaber: Not resolved. Still an `Int`.

examples/fluid/static_sphere_shape.jl
```diff
@@ -0,0 +1,92 @@
+# In this example we try to approach the static shape of a water droplet on a horizontal plane.
+# The shape of a static droplet can be calculated from the Young-Laplace equation.
+using TrixiParticles
```
efaulhaber, line 3: 1. I don't like the name. It's not a static example, so that's misleading. It's just as static as the `sphere_surface_tension`. Maybe `sphere_surface_tension_wall_2d.jl` or so? Because it's essentially the same as sphere surface tension, just on the ground, right?
2. Can't you just `trixi_include` the falling spheres example here? This would also make it clearer what is different in this example. Like:
```julia
using TrixiParticles
using OrdinaryDiffEq

fluid_particle_spacing = 0.0025
tspan = (0.0, 0.3)
tank_size = (1.0, 0.1)
sound_speed = 120.0
nu = 0.00025
sphere_radius = 0.05
sphere1_center = (0.5, sphere_radius)

trixi_include(@__MODULE__,
              joinpath(examples_dir(), "fluid", "falling_water_spheres_2d.jl"),
              sphere=nothing, ...,)
```

efaulhaber (changes requested):

examples/fluid/falling_water_spheres_2d.jl
svchb: 2. There is a difference. For more effect we need to increase the resolution and time the simulation runs.

examples/fluid/sphere_surface_tension_wall.jl
```diff
@@ -43,9 +43,9 @@ fluid_smoothing_kernel = SchoenbergCubicSplineKernel{2}()
 fluid_density_calculator = ContinuityDensity()
 
 # for perfect wetting
```
efaulhaber, line 45:
```suggestion
# For perfect wetting
```

examples/fluid/sphere_surface_tension_wall.jl
```diff
@@ -43,9 +43,9 @@ fluid_smoothing_kernel = SchoenbergCubicSplineKernel{2}()
...
-nu = 0.00025
+# nu = 0.0005
 # for no wetting
```
efaulhaber, line 47:
```suggestion
# For no wetting
```

examples/fluid/sphere_surface_tension_wall.jl
```diff
@@ -71,7 +71,7 @@ boundary_model = BoundaryModelDummyParticles(tank.boundary.density, tank.boundar
 
 # adhesion_coefficient = 1.0 and surface_tension_coefficient=0.01 for perfect wetting
 # adhesion_coefficient = 0.001 and surface_tension_coefficient=2.0 for no wetting
```
efaulhaber, line 73:
```suggestion
# `adhesion_coefficient = 1.0` and `surface_tension_coefficient = 0.01` for perfect wetting
# `adhesion_coefficient = 0.001` and `surface_tension_coefficient = 2.0` for no wetting
```

efaulhaber (changes requested):

examples/fluid/falling_water_spheres_2d.jl
efaulhaber: On multiple threads, it freezes, or when I pass `dtmin=eps()`, I get
```
#timesteps:   2100 │ Δt: 5.8212e-05 │ sim. time: 1.6747e-01 (55.823%)  │ run time: 1.1561e+01 s
#timesteps:   2150 │ Δt: 3.6845e-05 │ sim. time: 1.7009e-01 (56.697%)  │ run time: 1.1850e+01 s
#timesteps:   2200 │ Δt: 3.0012e-12 │ sim. time: 1.7114e-01 (57.045%)  │ run time: 1.2212e+01 s
#timesteps:   2250 │ Δt: 3.8658e-16 │ sim. time: 1.7114e-01 (57.045%)  │ run time: 1.2550e+01 s
┌ Warning: dt(2.220446049250313e-16) <= dtmin(2.220446049250313e-16) at t=0.17113580356914368, and step error estimate = 0.9780159000688392. Aborting. There is either an error in your model specification or the true solution is unstable.
└ @ SciMLBase ~/.julia/packages/SciMLBase/NjslX/src/integrator_interface.jl:619
```
Does the same happen for you? Or is it perfectly stable on your machine?

svchb: I don't run into any issues on my computer. 
Julia 1.10.2
OrdinaryDiffEq v6.76.0

PR #435, opened 2024-03-05
Add more tests

svchb:


.github/workflows/ci.yml
```diff
@@ -25,29 +25,27 @@ on:
...
 # Cancel redundant CI tests automatically
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
-  build:
-    name: Run Tests
-    runs-on: ubuntu-latest
+  # reduced to one version that runs the codecov upload since it fails alot
```
efaulhaber, line 34: Does it really fail a lot? Or did it only fail yesterday when the repository got disabled in Codecov after going public? Until I activated it again.

efaulhaber:
The problem is that unit tests are now only run on the latest Julia version. We want to run them on all platforms and Julia versions, but only once with coverage. Also, we need to test macOS-ARM.
I tried implementing this in #444. Let's see if I got it to work now.

PR #437, opened 2024-03-05
Reopen #382: Extrude shapes/geometry

LasNikas:
see #382


src/general/interpolation.jl
```diff
@@ -229,45 +229,14 @@ results = interpolate_plane_3d([0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0]
...
                               cut_off_bnd=true, clip_negative_pressure=false)
-    # Verify that points are in 3D space
-    if length(point1) != 3 || length(point2) != 3 || length(point3) != 3
-        throw(ArgumentError("all points must be 3D coordinates"))
-    end
+    coords, resolution_ = sample_plane((point1, point2, point3), resolution)
 
-    if ndims(ref_system) != 3
-        throw(ArgumentError("`interpolate_plane_3d` requires a 3D simulation"))
+    if !isapprox(resolution, resolution_, rtol=5e-2)
```
svchb, line 234: The interpolation functions don't have this alignment issue with particles since they are just points within a plane.

src/general/interpolation.jl
```diff
@@ -229,45 +229,14 @@ results = interpolate_plane_3d([0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [0.0, 1.0, 0.0]
 function interpolate_plane_3d(point1, point2, point3, resolution, semi, ref_system, sol;
                               smoothing_length=ref_system.smoothing_length,
                               cut_off_bnd=true, clip_negative_pressure=false)
-    # Verify that points are in 3D space
-    if length(point1) != 3 || length(point2) != 3 || length(point3) != 3
-        throw(ArgumentError("all points must be 3D coordinates"))
-    end
+    coords, resolution_ = sample_plane((point1, point2, point3), resolution)
```
svchb, line 232: This should keep the original resolution

src/setups/extrude_geometry.jl
```diff
@@ -0,0 +1,212 @@
...
+                    mass=nothing, density=nothing, pressure=0.0)
+Extrude either a line, a plane or a shape along a specific direction.
+
+# Arguments
+- `geometry`:           Either points defining a 3D plane (2D line)
+                        or particle coordinates defining a specific shape.
+
+# Keywords
+- `particle_spacing`:   Spacing between the particles.
+- `direction`:          Vector defining the extrusion direction.
```
svchb, line 13:
```suggestion
- `direction`:          A vector that specifies the direction in which the extrusion occurs.
```

src/setups/extrude_geometry.jl
```diff
@@ -0,0 +1,212 @@
...
+- `n_extrude=1`         Number of `geometry` layers in extrude direction.
```
svchb, line 14:
```suggestion
- `n_extrude=1`       Specifies the number of layers of geometry particles created in the direction of extrusion. The default value is 1.
```

src/setups/extrude_geometry.jl
```diff
@@ -0,0 +1,212 @@
...
+- `velocity`:           Either a function mapping each particle's coordinates to its velocity,
+                        or, for a constant fluid velocity, a vector holding this velocity.
+                        Velocity is constant zero by default.
+- `mass`:               Either `nothing` (default) to automatically compute particle mass from particle
```
svchb, line 18:
```suggestion
- `mass`:               Determines the mass of each particle and can be:
  - nothing (default), where the particle mass is automatically calculated using the particle density and spacing,
  - A function that gives a mass value based on each particle's coordinates,
  - A constant value that assigns the same mass to all particles.
```

src/setups/extrude_geometry.jl
```diff
@@ -0,0 +1,212 @@
...
+                        density and spacing, or a function mapping each particle's coordinates to its mass,
+                        or a scalar for a constant mass over all particles.
+- `density`:            Either a function mapping each particle's coordinates to its density,
```
svchb, line 21:
```suggestion
- `density`:            This parameter can be:

-A function that determines the density of each particle based on its coordinates,
-A constant value for uniform density across all particles.
Note: Setting this value is required if a state equation is not used and cannot be combined with the state_equation parameter.
```

src/setups/extrude_geometry.jl
```diff
@@ -0,0 +1,212 @@
...
+    dir1 = 0.5 * particle_spacing * direction
+    dir2 = 0.5 * particle_spacing * edge1
+    dir3 = 0.5 * particle_spacing * edge2
+
+    plane_point1 .+= dir1 + dir2 + dir3
+    plane_point2 .+= dir1 - dir2 + dir3
+    plane_point3 .+= dir1 + dir2 - dir3
+
+    return (plane_point1, plane_point2, plane_point3)
+end
```
svchb, line 212:
```suggestion
function adjust_plane_points(plane_points::NTuple{N, <:Any}, direction, particle_spacing) where N
    new_points = ()

    dir_shift = 0.5 * particle_spacing * direction

    for (i, point) in enumerate(plane_points)
        dir = dir_shift

        for (j, other_point) in enumerate(plane_points)
            if i != j
                dir += 0.5 * particle_spacing * normalize(other_point - point)
            end
        end

        new_point = copy(point) .+= dir
        new_points = (new_points..., new_point)
    end

    return new_points
end

function consider_particle_placement(plane_points::NTuple{2}, direction, particle_spacing, tlsph)
    (tlsph) && (return plane_points)
    return adjust_plane_points(plane_points, direction, particle_spacing)
end

function consider_particle_placement(plane_points::NTuple{3}, direction, particle_spacing, tlsph)
    (tlsph) && (return plane_points)
    return adjust_plane_points(plane_points, direction, particle_spacing)
end
```
Something like this (untested)

test/setups/extrude_geometry.jl
```diff
@@ -0,0 +1,136 @@
...
+                           0.05 0.15 0.24999999999999997 0.35 0.44999999999999996 0.55 0.65 0.7499999999999999 0.8499999999999999 0.95 0.05 0.15 0.24999999999999997 0.35 0.44999999999999996 0.55 0.65 0.7499999999999999 0.8499999999999999 0.95 0.05 0.15 0.24999999999999997 0.35 0.44999999999999996 0.55 0.65 0.7499999999999999 0.8499999999999999 0.95 0.05 0.15 0.24999999999999997 0.35 0.44999999999999996 0.55 0.65 0.7499999999999999 0.8499999999999999 0.95 0.05 0.15 0.24999999999999997 0.35 0.44999999999999996 0.55 0.65 0.7499999999999999 0.8499999999999999 0.95]
+        shape = ExtrudeGeometry((point1, point2); direction=[-1.0, 0.0],
+                                particle_spacing=0.1, n_extrude=5, density=1.0)
+
+        @test shape.coordinates ≈ expected_coords
+    end
+end
+
+# 3D
+@testset verbose=true "Extrude Geometry 3D" begin
```
svchb, line 78: All of these lists are very long. Can you load them from csv?

svchb (changes requested):

src/general/interpolation.jl
LasNikas: if the `resolution` is not a multiple of the edge length, the resolution changes.
The number of points is determined along each edge with `ceil(Int, edge_length/resolution)`.

svchb: Ah I see this doesn't change anything. Than I don't think this warning is necessary.

src/setups/extrude_geometry.jl
svchb: @efaulhaber Do you like the proposed list form more than the current text?

src/setups/extrude_geometry.jl
efaulhaber: grafik
grafik
I like it!
But let's keep the old version for now and change that in a new PR. Then we can also make the default values consistent (mentioning in text vs in the keys `mass=nothing`).

src/setups/extrude_geometry.jl
```diff
@@ -0,0 +1,212 @@
...
+shape = SphereShape(0.1, 0.5, (0.2, 0.4), 1000.0, n_layers=3,
+                    sphere_type=RoundSphere(end_angle=pi))
+
+direction = [0.0, 0.0, 1.0]
+
+shape = ExtrudeGeometry(shape; direction, particle_spacing=0.1, n_extrude=4, density=1000.0)
+```
+
+!!! warning
+    `particle_spacing` between extrusion layers may differ from shapes `particle_spacing`.
```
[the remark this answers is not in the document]
efaulhaber: And can the `density` of an `InitialCondition` just be extruded as well? This would be useful for creating a 3D parallelogram tank, and it would make the API easier by not having to pass `density` all the time.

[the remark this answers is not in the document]
efaulhaber: Speaking of API: After our discussion 2 days ago, it would be good to add a warning
```suggestion
!!! warning "Experimental Implementation"
    This is an experimental feature and may change in any future releases.
```
so that we can release breaking changes to it without releasing a new major release.
We should do that with all new features, and remove the warning once we've used it for a while and we're happy with the API.

src/setups/extrude_geometry.jl
```diff
@@ -0,0 +1,212 @@
...
+    face_coords, particle_spacing_ = sample_plane(geometry, particle_spacing)
+
+    if !isapprox(particle_spacing, particle_spacing_, rtol=5e-2)
+        @info "The desired size is not a multiple of the particle spacing $particle_spacing." *
+              "\nNew particle spacing is set to $particle_spacing_."
+    end
+
+    coords = (face_coords .+ i * particle_spacing_ * direction_ for i in 0:(n_extrude - 1))
+
+    coordinates = reshape(stack(coords), (NDIMS, size(face_coords, 2) * n_extrude))
```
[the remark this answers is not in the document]
LasNikas: It's slightly faster with `stack`:
```julia
infil> f(coords, NDIMS, face_coords, n_extrude) = reshape(stack(coords), (NDIMS, size(face_coords, 2) * n_extrude))
f (generic function with 2 methods)

infil> f2(coords) = hcat(coords...)
f2 (generic function with 1 method)

infil> @benchmark f2($coords)
BenchmarkTools.Trial: 10000 samples with 9 evaluations.
 Range (min … max):  2.167 μs … 902.048 μs  ┊ GC (min … max):  0.00% … 96.86%
 Time  (median):     2.569 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   3.200 μs ±  13.700 μs  ┊ GC (mean ± σ):  12.62% ±  3.73%

    ▁▂▅▇█▅▂                                                    
  ▂▅████████▅▄▄▄▅▄▄▃▃▃▃▂▃▃▂▂▂▂▂▂▂▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  2.17 μs         Histogram: frequency by time        5.32 μs <

 Memory estimate: 10.02 KiB, allocs estimate: 20.

infil> @benchmark f($coords, $NDIMS, $face_coords, $n_extrude)
BenchmarkTools.Trial: 10000 samples with 10 evaluations.
 Range (min … max):  1.445 μs … 177.663 μs  ┊ GC (min … max):  0.00% … 90.26%
 Time  (median):     1.679 μs               ┊ GC (median):     0.00%
 Time  (mean ± σ):   2.078 μs ±   6.000 μs  ┊ GC (mean ± σ):  11.59% ±  4.01%

    ▁▆█▇▆▁                                                     
  ▁▃██████▆▅▅▄▄▄▄▃▃▃▃▃▃▃▃▂▂▂▂▂▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁ ▂
  1.44 μs         Histogram: frequency by time         3.5 μs <

 Memory estimate: 9.78 KiB, allocs estimate: 13.

infil> 
```

src/setups/extrude_geometry.jl
```diff
@@ -0,0 +1,212 @@
...
+
+    particle_spacing_new = min(norm(edge1 / num_points_edge1),
+                               norm(edge2 / num_points_edge2))
+
+    return coords, particle_spacing_new
+end
+
+function consider_particle_placement(geometry::Union{AbstractMatrix, InitialCondition},
+                                     direction, particle_spacing, tlsph)
+    return geometry
```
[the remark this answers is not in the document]
LasNikas: I don't really know what you mean? 
But, in terms of open boundaries, I need the particles to start half a particle spacing away from the plane/line, like so:
image

src/setups/extrude_geometry.jl
```diff
@@ -0,0 +1,212 @@
+"""
+    ExtrudeGeometry(geometry; particle_spacing, direction, n_extrude=1,
+                    velocity=zeros(length(direction)),
+                    mass=nothing, density=nothing, pressure=0.0)
+Extrude either a line, a plane or a shape along a specific direction.
+
+# Arguments
+- `geometry`:           Either points defining a 3D plane (2D line)
+                        or particle coordinates defining a specific shape.
```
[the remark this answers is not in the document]
LasNikas: I was also thinking about that. "plane" can be misleading

src/setups/extrude_geometry.jl
efaulhaber: That was my point. There shouldn't be a default value for this.

src/setups/extrude_geometry.jl
```diff
@@ -0,0 +1,212 @@
...
+        for j in 0:num_points_edge2
+            point_on_plane = plane_points[1] + (i / num_points_edge1) * edge1 +
+                             (j / num_points_edge2) * edge2
+            coords[:, index] = point_on_plane
+            index += 1
+        end
+    end
+
+    particle_spacing_new = min(norm(edge1 / num_points_edge1),
+                               norm(edge2 / num_points_edge2))
```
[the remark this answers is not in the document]
efaulhaber: Yeah. Both is not great. Let's leave it like this for now and see if we run into problems later.

src/setups/extrude_geometry.jl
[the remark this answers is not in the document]
efaulhaber: Interesting. Then add a comment stating that this is faster than `hcat(coords...)`.

src/setups/extrude_geometry.jl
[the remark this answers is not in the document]
efaulhaber: Could you please open an issue for this? Then we can look into it in another PR.

test/setups/extrude_geometry.jl
LasNikas: Hm, we also hard code the coordinates for the `RectangularTank` tests. Is it worth loading the coordinates from a separate file? It took me a while to generate this, and I need to generate this again for a csv file. Can we leave it as is and if it gets complicated we can still add a separate csv file. @efaulhaber what's your opinion on this?

On the other hand, it's a good point to put coordinates for setup tests in a separate csv file...

svchb: Yes I also put long lists in the interpolation tests. But this is crap. Maybe we should agree to put these in csv files in the future?

src/setups/extrude_geometry.jl
svchb: @LasNikas Do you want to open an Issue for this?

LasNikas: already done in #479

test/setups/extrude_geometry.jl
efaulhaber: Let's discuss this in an issue.

efaulhaber: #490

src/setups/extrude_geometry.jl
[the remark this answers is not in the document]
efaulhaber:
```suggestion
                    mass=nothing, density=nothing, pressure=0.0)
Extrude either a line, a plane or a shape along a specific direction.

# Arguments
- `geometry`:           Either particle coordinates or an [`InitialCondition`](@ref) defining a 2D shape
                        to extrude to a 3D volume,
                        or 2 2D points defining a line to extrude to a plane in 2D,
                        or 3 3D points defining a plane to extrude to a volume.
```

src/setups/extrude_geometry.jl
```diff
@@ -31,45 +31,54 @@ Extrude either a line, a plane or a shape along a specific direction.
...
+# Extrude a 2D shape (here: a disc) to a 3D shape (here: a cylinder)
 shape = SphereShape(0.1, 0.5, (0.2, 0.4), 1000.0, n_layers=3,
                     sphere_type=RoundSphere(end_angle=pi))
 
 direction = [0.0, 0.0, 1.0]
 
-shape = ExtrudeGeometry(shape; direction, particle_spacing=0.1, n_extrude=4, density=1000.0)
+shape = extrude_geometry(shape; direction, particle_spacing=0.1, n_extrude=4, density=1000.0)
+
+# output
```
efaulhaber, line 60: Why not remove the filter above (idk why it's not working) and explicitly test for this output?
```suggestion
# output
┌ Info: The desired size is not a multiple of the particle spacing 0.1.
└ New particle spacing is set to 0.09387239731236392.
┌ Info: The desired size is not a multiple of the particle spacing 0.1.
└ New particle spacing is set to 0.09198039027185569.
```

src/setups/extrude_geometry.jl
```diff
@@ -1,5 +1,5 @@
-"""
-    ExtrudeGeometry(geometry; particle_spacing, direction, n_extrude=1,
+@doc raw"""
+    extrude_geometry(geometry; particle_spacing, direction, n_extrude=1,
                     velocity=zeros(length(direction)),
                     mass=nothing, density=nothing, pressure=0.0)
```
efaulhaber, line 4:
```suggestion
    extrude_geometry(geometry; particle_spacing, direction, n_extrude=1,
                     velocity=zeros(length(direction)),
                     mass=nothing, density=nothing, pressure=0.0)
```

src/setups/extrude_geometry.jl
LasNikas: no, I'll change it to `n_extrude::Integer`

src/setups/extrude_geometry.jl
```diff
@@ -31,45 +31,54 @@ Extrude either a line, a plane or a shape along a specific direction.
...
+                          mass=nothing, density=nothing, pressure=0.0)
     direction_ = normalize(direction)
     NDIMS = length(direction_)
 
-    geometry = consider_particle_placement(geometry, direction_, particle_spacing, tlsph)
+    if !(geometry isa InitialCondition) && particle_spacing == -1
+        throw(ArgumentError("`particle_spacing` must be specified when not extruding an `InitialCondition`"))
+    elseif geometry isa InitialCondition
+        particle_spacing = geometry.particle_spacing
+    end
```
efaulhaber, line 77:
```suggestion
    if geometry isa InitialCondition && geometry.particle_spacing > 0
        if particle_spacing > 0 && particle_spacing != geometry.particle_spacing
            throw(ArgumentError("`particle_spacing` must be -1 when using an `InitialCondition`"))
        end
        particle_spacing = geometry.particle_spacing
    end
    
    if particle_spacing <= 0
        throw(ArgumentError("`particle_spacing` must be specified when not extruding an `InitialCondition`"))
    end
```
The error message is not 100% correct, as we also print "`particle_spacing` must be specified when not extruding an `InitialCondition`" when an `InitialCondition` with `InitialCondition.particle_spacing == -1` is used, but I think explaining that in an error message is too complicated and confusing.

src/setups/extrude_geometry.jl
```diff
@@ -0,0 +1,244 @@
...
+                    velocity=zeros(length(direction)),
+                    mass=nothing, density=nothing, pressure=0.0)
+Extrude either a line, a plane or a shape along a specific direction.
+
+# Arguments
+- `geometry`:           Either points defining a 3D plane (2D line)
+                        or particle coordinates defining a specific shape.
+
+# Keywords
+- `particle_spacing`:   Spacing between the particles.
```
efaulhaber, line 12:
```suggestion
- `particle_spacing`:   Spacing between the particles. Can be omitted when `geometry` is an `InitialCondition` (unless `geometry.particle_spacing == -1`).
```

efaulhaber (changes requested):

svchb (approved):

src/general/interpolation.jl
```diff
@@ -262,6 +262,17 @@ function interpolate_plane_3d(point1, point2, point3, resolution, semi, ref_syst
                               sol::ODESolution;
                               smoothing_length=ref_system.smoothing_length,
                               cut_off_bnd=true, clip_negative_pressure=false)
+    v_ode = sol.u[end].x[1]
+    u_ode = sol.u[end].x[2]
```
efaulhaber, line 266:
```suggestion
    v_ode, u_ode = sol.u[end].x
```

efaulhaber (changes requested):

svchb (approved):

efaulhaber (approved):

PR #439, opened 2024-03-05
Reopen #351: Complex shapes

LasNikas:


src/preprocessing/point_in_poly/winding_number_horman.jl
```diff
@@ -0,0 +1,77 @@
...
+    v_unit = SVector(0.0, 1.0)
+
+    function quadrant(dot_v, dot_h)
+        (dot_v >= 0.0 && dot_h > 0.0) && return 0
+        (dot_v > 0.0 && dot_h <= 0.0) && return 1
+        (dot_v <= 0.0 && dot_h < 0.0) && return 2
+        (dot_v < 0.0 && dot_h >= 0.0) && return 3
+    end
+
+    Threads.@threads for query_point in axes(points, 2)
```
efaulhaber, line 29:
```suggestion
    @threaded points for query_point in axes(points, 2)
```
You can also use an array when no system is available. This will run `Polyester.@batch` when `points` is an `Array` and GPU code when it's a GPU array type.

LasNikas: Ah thanks!

src/preprocessing/shapes/io.jl
```diff
@@ -0,0 +1,135 @@
...
+    end
+
+    write(io, UInt32(length(faces))) # write triangle count
+    for i in faces
+        n = SVector{3, Float32}(normals[i])
+        triangle = points[i]
+        foreach(j -> write(io, n[j]), 1:3)
+        for point in triangle
+            foreach(p -> write(io, p), SVector{3, Float32}(point))
+        end
```
efaulhaber, line 115:
```suggestion

        for j in 1:3
            write(io, n[j])
        end

        for point in triangle, p in point
            write(io, Float32(p))
        end
```

src/preprocessing/shapes/io.jl
```diff
@@ -0,0 +1,135 @@
...
+        write(io, 0x0000) # write 16bit empty bit
```
efaulhaber, line 116: Please add a comment why.

src/preprocessing/shapes/shapes.jl
```diff
@@ -0,0 +1,13 @@
+abstract type Shapes{NDIMS} end
```
efaulhaber, line 1: This is a bit confusing. `ComplexShape` is no `Shapes`? And `RectangularShape` isn't either? I think you should find a different name for this abstract type.

src/preprocessing/shapes/io.jl
```diff
@@ -0,0 +1,135 @@
...
+# FileIO.jl docs:
+# https://juliaio.github.io/FileIO.jl/stable/implementing/#All-at-once-I/O:-implementing-load-and-save
+function load(fn::File{format}; element_types...) where {format}
+    open(fn) do s
+        skipmagic(s) # skip over the magic bytes
+        load(s; element_types...)
+    end
+end
+
+function load(fs::Stream{format"STL_BINARY"}; ELTYPE=Float64)
```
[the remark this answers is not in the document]
LasNikas: It makes sure that it is a binary .stl file and not a ascii .stl

src/preprocessing/shapes/io.jl
```diff
@@ -0,0 +1,135 @@
...
+    # Binary STL
+    # https://en.wikipedia.org/wiki/STL_%28file_format%29#Binary_STL
+    io = stream(fs)
+    read(io, 80) # throw out header
```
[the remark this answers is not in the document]
LasNikas: It's in the link above?
image

src/preprocessing/shapes/io.jl
```diff
@@ -0,0 +1,135 @@
...
+    for i in faces
+        n = SVector{3, Float32}(normals[i])
+        triangle = points[i]
+        foreach(j -> write(io, n[j]), 1:3)
+        for point in triangle
+            foreach(p -> write(io, p), SVector{3, Float32}(point))
+        end
+        write(io, 0x0000) # write 16bit empty bit
+    end
+end
```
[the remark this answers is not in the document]
LasNikas: Hm Idk why we should write .stl files. It was super helpful for debugging. What do you mean with throw in `test`?

src/preprocessing/shapes/shapes.jl
LasNikas: how about `AbstractGeometry`?

test/preprocessing/data/circle.asc
```diff
@@ -0,0 +1,65 @@
+# ASCII
```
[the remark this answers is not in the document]
LasNikas: There is no difference. But I couldn't load the files in CI IIRC

src/preprocessing/point_in_poly/winding_number_horman.jl
```diff
@@ -0,0 +1,77 @@
+"""
+    WindingNumberHorman()
+
+Algorithm for inside-outside segmentation of a complex shape proposed by Horman et al. (2001).
```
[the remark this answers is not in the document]
LasNikas: It's in the markdown file

src/preprocessing/shapes/io.jl
```diff
@@ -0,0 +1,135 @@
...
+function load(fs::Stream{format"STL_BINARY"}; ELTYPE=Float64)
+    # Binary STL
+    # https://en.wikipedia.org/wiki/STL_%28file_format%29#Binary_STL
+    io = stream(fs)
+    read(io, 80) # throw out header
+    n_faces = read(io, UInt32)
+
+    face_vertices = fill(ntuple(_ -> fill(zero(ELTYPE), SVector{3}), 3), n_faces)
+    vertices = fill(fill(zero(ELTYPE), SVector{3}), 3n_faces)
+    normals = fill(fill(zero(ELTYPE), SVector{3}), n_faces)
```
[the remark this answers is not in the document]
LasNikas: That is the data structure I'm using in `TriangleMesh` the query of a vertex is faster then

src/preprocessing/shapes/io.jl
```diff
@@ -0,0 +1,135 @@
...
+
+    i = 0
+    while !eof(io)
+        normals[i + 1] = SVector{3, ELTYPE}(read(io, Float32), read(io, Float32),
+                                            read(io, Float32))
```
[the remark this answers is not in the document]
LasNikas: Idk but isn't it cleaner that way?

src/preprocessing/shapes/shapes.jl
svchb: I would prefer to rename several things in this pr to geometry since this seems to me a more common name for what this pr is doing. While the concept of a shape is more limited to our own nomenclature.

test/preprocessing/data/circle.asc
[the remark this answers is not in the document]
svchb: didn't we have this question in slack and it was only a path thing?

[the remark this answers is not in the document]
LasNikas: Yes, that's why I copied the files to the test folder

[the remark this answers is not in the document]
svchb: Hmm I thought there was a solution

src/preprocessing/shapes/io.jl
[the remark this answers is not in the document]
efaulhaber: Just put it in the `test` directory instead of `src`. Then you have it available for debugging, and you can even write a test if writing and reading again yields the same result as before.

src/setups/complex_shape.jl
```diff
@@ -0,0 +1,108 @@
...
+
+
+!!! warning "Experimental Implementation"
+    This is an experimental feature and may change in any future releases.
+"""
+function ComplexShape(shape::Shapes; particle_spacing, density, pressure=0.0, mass=nothing,
+                      velocity=zeros(ndims(shape)),
+                      point_in_shape_algorithm=WindingNumberJacobson(; shape,
+                                                                     hierarchical_winding=false,
+                                                                     winding_number_factor=sqrt(eps())),
```
efaulhaber, line 47: This factor is never used?

src/preprocessing/shapes/polygon_shape.jl
```diff
@@ -0,0 +1,103 @@
+struct Polygon{NDIMS, ELTYPE} <: Shapes{NDIMS}
```
efaulhaber, line 1:
```suggestion
# This is the data format returned by `load(file)` when used with `.asc` files
struct Polygon{NDIMS, ELTYPE} <: Shapes{NDIMS}
```

src/preprocessing/shapes/triangle_mesh.jl
```diff
@@ -0,0 +1,137 @@
+struct TriangleMesh{NDIMS, ELTYPE} <: Shapes{NDIMS}
```
efaulhaber, line 1:
```suggestion
# This is the data format returned by `load(file)` when used with `.stl` files
struct TriangleMesh{NDIMS, ELTYPE} <: Shapes{NDIMS}
```

efaulhaber (changes requested):

src/setups/complex_shape.jl
LasNikas: Yes, it is. In the `point_in_poly` function to relax the winding number condition

efaulhaber: Ah, GitHub formatting made it look like it was a kwarg to this function here.

src/preprocessing/shapes/triangle_mesh.jl
```diff
@@ -0,0 +1,138 @@
...
+            v3 = face_vertices[i][3]
+
+            # TODO: This part is about 90% of the runtime.
+            # Vertex IDs are only needed for the hierarchical winding.
+            vertex_id1 = findfirst(x -> v1 == x, vertices)
+            vertex_id2 = findfirst(x -> v2 == x, vertices)
+            vertex_id3 = findfirst(x -> v3 == x, vertices)
+
+            face_vertices_ids[i] = (vertex_id1, vertex_id2, vertex_id3)
+        end
```
efaulhaber, line 33: You build the `vertices` array yourself.
```julia
        face_vertices[i + 1] = (v1, v2, v3)

        vertices[3 * i + 1] = v1
        vertices[3 * i + 2] = v2
        vertices[3 * i + 3] = v3
```
And then you have to find the indices again? This seems very unnecessary to me.
Why don't you make `face_vertices` an `NTuple{3, Int}`, which points to the vertex IDs?
Then you can also get rid of duplicate vertices. I guess each vertex appears once for each adjacent triangle.

efaulhaber (changes requested):

src/setups/complex_shape.jl
```diff
@@ -45,55 +45,51 @@ function ComplexShape(shape::Shapes; particle_spacing, density, pressure=0.0, ma
...
     NDIMS = ndims(shape)
 
-    if !(seed isa AbstractVector)
-        throw(ArgumentError("`seed` must be of type `AbstractVector`"))
-    elseif !(length(seed) == NDIMS)
-        throw(ArgumentError("`seed` must be of length $NDIMS for a $(NDIMS)D problem"))
-    end
-
-    min_corner = seed
+    min_corner = shape.min_corner .- grid_offset
```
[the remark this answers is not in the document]
LasNikas: That's enough. One only might need this when edges or faces are exactly axis aligned. This most likely happens for simple geometries like the `hexagon.asc` or the `bar.stl` example geometry

src/setups/complex_shape.jl
```diff
@@ -45,55 +45,51 @@ function ComplexShape(shape::Shapes; particle_spacing, density, pressure=0.0, ma
                       point_in_shape_algorithm=WindingNumberJacobson(; shape,
                                                                      hierarchical_winding=false,
                                                                      winding_number_factor=sqrt(eps())),
-                      store_winding_number=false, seed=shape.min_corner,
-                      max_nparticles=Int(1e7), pad_initial_particle_grid=2particle_spacing)
+                      store_winding_number=false, grid_offset::AbstractFloat=0.0,
```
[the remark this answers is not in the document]
LasNikas: why pretty bad? The sampling then starts somewhere in the shape. Maybe someone would like to have this as a feature? :upside_down_face:

src/preprocessing/shapes/triangle_mesh.jl
LasNikas: >  I guess each vertex appears once for each adjacent triangle.

No, that is the problem. The stl-file only stores the coordinates of the triangle vertices. That is, we do not know which triangles are adjacent or which triangles do share the same vertex. That means, each vertex ID is unique but there are vertices with the same coordinates.

src/preprocessing/shapes/shapes.jl
LasNikas: > how about AbstractGeometry?
?

src/setups/complex_shape.jl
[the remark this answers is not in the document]
efaulhaber: If you really think someone would want to cut off the corner of the geometry, then at least add a comment that this will happen for negative offsets.

src/preprocessing/shapes/triangle_mesh.jl
efaulhaber: No that's what I meant. You each vertex appears again for each adjacent triangle. But in the code I quoted above, you can check if this vertex already exists, and skip it in this case. But if you don't care if you have duplicate vertices, then don't do this.

efaulhaber (changes requested):

src/preprocessing/shapes/polygon_shape.jl
```diff
@@ -0,0 +1,101 @@
...
+
+            push!(edge_vertices, (v1, v2))
+            push!(edge_normals, edge_normal)
+        end
+
+        vertex_normals = Vector{Vector{SVector{NDIMS, ELTYPE}}}()
+
+        # Calculate vertex pseudo-normals
+        # A edge is defined with two vertices of which we calculate the pseudo-normals
+        # by averaging the normals of the neighbor edges.
```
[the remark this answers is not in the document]
LasNikas: Yes, it will be useful for the signed distances. 
We calculate the distances for a single edge and don't look for adjacent edges. That means, we do not iterate over all edges.

svchb (changes requested):
Missing a NEWS.md entry and a README.md adjustment.

src/preprocessing/shapes/polygon_shape.jl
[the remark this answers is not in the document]
efaulhaber: Then please add this in a comment.

src/preprocessing/shapes/triangle_mesh.jl
efaulhaber: My idea doesn't work, as discussed, because the vertices need to be unique.

src/preprocessing/shapes/triangle_mesh.jl
```diff
@@ -0,0 +1,146 @@
...
+        @threaded face_vertices for i in 1:n_faces
+            v1 = face_vertices[i][1]
+            v2 = face_vertices[i][2]
+            v3 = face_vertices[i][3]
+
+            # TODO: This part is about 90% of the runtime.
+            # Vertex IDs are only needed for the hierarchical winding.
+            vertex_id1 = findfirst(x -> v1 == x, vertices)
+            vertex_id2 = findfirst(x -> v2 == x, vertices)
+            vertex_id3 = findfirst(x -> v3 == x, vertices)
```
efaulhaber, line 36:
```suggestion
            # We removed duplicate vertices with an absolute tolerance of 1e-14.
            # `searchsortedfirst` finds the first value in `vertices` greater than or equivalent to the first argument.
            # We might have removed `v1` because there was `v1 - 1e-14` in `vertices`, so we can find
            # this value by subtracting `1e-14`.
            vertex_id1 = searchsortedfirst(v1 - 1e-14, vertices)
            vertex_id2 = searchsortedfirst(v2 - 1e-14, vertices)
            vertex_id3 = searchsortedfirst(v3 - 1e-14, vertices)
```

efaulhaber (changes requested):

src/preprocessing/shapes/io.jl
```diff
@@ -0,0 +1,100 @@
...
+        vertices[3 * i + 3] = v3
+
+        # After the 48 bytes of the normal and the vertices follows a 2-byte unsigned integer
+        # that is the "attribute byte count" – in the standard format, this should be zero
+        # because most software does not understand anything else.
+        skip(io, 2) # Skip attribute byte count
+        i += 1
+    end
+
+    union!(vertices)
```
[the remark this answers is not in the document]
efaulhaber: Quick benchmark:
```julia
julia> x = rand(SVector{3, Float64}, 100_000);

julia> @btime unique($x);
  5.604 ms (44 allocations: 12.39 MiB)

julia> @btime unique!($(copy(x)));
  6.012 ms (33 allocations: 8.33 MiB)

julia> @btime union($x);
  6.933 ms (44 allocations: 12.39 MiB)

julia> @btime union!($(copy(x)));
  5.410 ms (7 allocations: 6.25 MiB)

julia> @btime unique2($x);
  536.392 μs (5 allocations: 2.30 MiB)

julia> sort(unique(x)) == sort(union(x)) == unique2(x)
true
```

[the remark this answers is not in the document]
efaulhaber: Another benchmark with a non-unique vector:
```julia
julia> length(unique(x))
100000

julia> for i in 2:length(x)
           # Create duplicate values
           if rand(Bool)
               x[i] = x[i - 1]
           end
       end

julia> length(unique(x))
50099

julia> sort(unique(x)) == sort(union(x)) == unique2(x)
true

julia> @btime unique($x);
  4.729 ms (43 allocations: 10.71 MiB)

julia> @btime unique!($(copy(x)));
  3.711 ms (33 allocations: 8.33 MiB)

julia> @btime union($x);
  5.296 ms (43 allocations: 10.71 MiB)

julia> @btime union!($(copy(x)));
  2.640 ms (7 allocations: 3.13 MiB)

julia> @btime unique2($x);
  912.960 μs (5 allocations: 1.16 MiB)
```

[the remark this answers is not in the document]
efaulhaber: Bigger problem:
```julia
julia> x = rand(SVector{3, Float64}, 10_000_000);

julia> for i in 2:length(x)
           # Create duplicate values
           if rand(Bool)
               x[i] = x[i - 1]
           end
       end

julia> @btime unique!($(copy(x)));
  913.825 ms (58 allocations: 395.84 MiB)

julia> @btime union!($(copy(x)));
  889.142 ms (7 allocations: 200.00 MiB)

julia> @btime unique2($x);
  130.388 ms (6 allocations: 115.65 MiB)

julia> sort(unique(x)) == sort(union(x)) == unique2(x)
true

julia> length(unique(x))
5000910
```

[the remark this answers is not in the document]
efaulhaber: Small problem, just for fun:
```julia
julia> x = rand(SVector{3, Float64}, 10_000);

julia> for i in 2:length(x)
           # Create duplicate values
           if rand(Bool)
               x[i] = x[i - 1]
           end
       end

julia> @btime unique!($(copy(x)));
  285.236 μs (23 allocations: 534.55 KiB)

julia> @btime union!($(copy(x)));
  202.334 μs (7 allocations: 200.84 KiB)

julia> @btime unique2($x);
  80.052 μs (5 allocations: 116.75 KiB)
```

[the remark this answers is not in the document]
efaulhaber: Serial performance is also looking good:
```julia
julia> x = rand(SVector{3, Float64}, 1_000_000);

julia> for i in 2:length(x)
           # Create duplicate values
           if rand(Bool)
               x[i] = x[i - 1]
           end
       end

julia> @btime unique!($(copy(x)));
  49.137 ms (43 allocations: 45.84 MiB)

julia> @btime union!($(copy(x)));
  54.683 ms (7 allocations: 25.00 MiB)

julia> @btime unique2($x);
  25.445 ms (5 allocations: 11.57 MiB)
```
I get a ~2.5x speedup on 128 threads.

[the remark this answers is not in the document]
efaulhaber: Update: This is not robust for vectors. This should work, though:
```julia
        function unique2(vertices_)
           # Sort by the first entry of the vectors
           compare_first_element = (x, y) -> x[1] < y[1]
           vertices_sorted = sort!(vertices_, lt=compare_first_element)
           keep = trues(length(vertices_sorted))
           
           PointNeighbors.@threaded vertices_sorted for i in eachindex(vertices_sorted)
               # We only sorted by the first entry, so we have to check all previous vertices
               # until the first entry is too far away.
               j = i - 1
               while j >= 1 && isapprox(vertices_sorted[j][1], vertices_sorted[i][1], atol=1e-14)
                   if isapprox(vertices_sorted[i], vertices_sorted[j], atol = 1e-14)
                       keep[i] = false
                   end
                   j -= 1
               end
           end
           vertices = vertices_sorted[keep]
       end
```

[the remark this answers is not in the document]
LasNikas: Thanks a lot! I benchmarked it with a 12k faces geometry 
image

### your version:
```julia
julia> @benchmark load_shape($filename)
BenchmarkTools.Trial: 12 samples with 1 evaluation.
 Range (min … max):  353.833 ms … 650.794 ms  ┊ GC (min … max): 1.43% … 3.31%
 Time  (median):     429.000 ms               ┊ GC (median):    1.95%
 Time  (mean ± σ):   439.863 ms ±  84.863 ms  ┊ GC (mean ± σ):  1.87% ± 1.08%

  ███ ██      █     █   ██    █ █                             █  
  ███▁██▁▁▁▁▁▁█▁▁▁▁▁█▁▁▁██▁▁▁▁█▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  354 ms           Histogram: frequency by time          651 ms <

 Memory estimate: 44.48 MiB, allocs estimate: 966312.
```

### old (`vertex_id1 = findfirst(x -> v1 == x, vertices)`:
```julia
julia> @benchmark load_shape($filename)
BenchmarkTools.Trial: 11 samples with 1 evaluation.
 Range (min … max):  401.359 ms … 642.248 ms  ┊ GC (min … max): 1.26% … 1.34%
 Time  (median):     472.198 ms               ┊ GC (median):    1.37%
 Time  (mean ± σ):   477.176 ms ±  73.588 ms  ┊ GC (mean ± σ):  1.48% ± 1.10%

  █ ▁▁        ▁    ▁▁         ▁  ▁ ▁                          ▁  
  █▁██▁▁▁▁▁▁▁▁█▁▁▁▁██▁▁▁▁▁▁▁▁▁█▁▁█▁█▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁█ ▁
  401 ms           Histogram: frequency by time          642 ms <

 Memory estimate: 45.94 MiB, allocs estimate: 1012889.
```

Your version is slightly faster and also it doesn't matter when two vertices differ by a very small number. so I think we can go with yours?

src/preprocessing/shapes/triangle_mesh.jl
```diff
@@ -0,0 +1,182 @@
...
+    a = triangle_points[1]
+    b = triangle_points[2]
+    c = triangle_points[3]
+    sab = dot(b - a, b - a)
+    sbc = dot(b - c, b - c)
+    sac = dot(c - a, c - a)
+
+    alpha_ = (sab + sac - sbc) / (2 * sqrt(sab * sac))
+    beta_ = (sbc + sab - sac) / (2 * sqrt(sbc * sab))
+    gamma_ = (sac + sbc - sab) / (2 * sqrt(sac * sbc))
```
efaulhaber, line 150:
```suggestion
    # Applying the law of cosines
    # https://en.wikipedia.org/wiki/Law_of_cosines
    cos_alpha = (sab + sac - sbc) / (2 * sqrt(sab * sac))
    cos_beta = (sbc + sab - sac) / (2 * sqrt(sbc * sab))
    cos_gamma = (sac + sbc - sab) / (2 * sqrt(sac * sbc))
```

src/preprocessing/shapes/triangle_mesh.jl
```diff
@@ -0,0 +1,182 @@
...
+
+    alpha_ = isfinite(alpha_) ? clamp(alpha_, -1, 1) : zero(eltype(sab))
```
efaulhaber, line 152:
```suggestion
    # If one side has length zero, this assures that the two adjacent angles
    # become pi/2, while the third angle becomes zero.
    alpha_ = isfinite(alpha_) ? clamp(alpha_, -1, 1) : zero(eltype(sab))
```

efaulhaber (changes requested):

NEWS.md
```diff
@@ -9,6 +9,7 @@ We aim at 3 to 4 month between major release versions and about 2 weeks between
 ### Highlights
 
 ### Added
+Particle sampling for complex shapes from `.stl` and `.asc` files.
 
 ### Removed
```
svchb, line 15: This is not the correct format for the current dev cycle i.e.

## Version 0.2.1
### Added
Particle sampling for complex shapes from `.stl` and `.asc` files.

README.md
```diff
@@ -21,6 +21,7 @@ Its features include:
 - Solid-body mechanics
   - Methods:  Total Lagrangian SPH (TLSPH), Discrete Element Method (DEM)
 - Fluid-Structure Interaction
+- Preprocess complex shapes
```
svchb, line 24: No one is going to search for that. Add support for STL geometries.

README.md
```diff
@@ -21,6 +21,7 @@ Its features include:
...
 - Output formats:
   - VTK
```
svchb, line 27: An impressive picture of a complex geometry would also be nice.

svchb (changes requested):

src/preprocessing/shapes/io.jl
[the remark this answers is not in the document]
LasNikas: update:
It's a bit faster with
```julia
vertex_id1 = searchsortedfirst(vertices, v1 .- 1e-14)
@inbounds for vertex_id in eachindex(vertices)[vertex_id1:end]
    if isapprox(vertices[vertex_id], v1)
        vertex_id1 = vertex_id
        break
    end
end
```

```julia
julia> @benchmark load_shape($filename)
BenchmarkTools.Trial: 15 samples with 1 evaluation.
 Range (min … max):  322.934 ms … 361.311 ms  ┊ GC (min … max): 0.00% … 1.14%
 Time  (median):     336.595 ms               ┊ GC (median):    1.20%
 Time  (mean ± σ):   336.979 ms ±   7.912 ms  ┊ GC (mean ± σ):  1.10% ± 0.50%

                      ▃█▃▃                                       
  ▇▁▁▁▁▁▁▁▁▁▇▁▁▁▁▁▁▁▇▇████▁▁▁▁▇▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▁▇ ▁
  323 ms           Histogram: frequency by time          361 ms <

 Memory estimate: 44.48 MiB, allocs estimate: 966311.
```

src/preprocessing/point_in_poly/winding_number_jacobson.jl
```diff
@@ -0,0 +1,136 @@
...
+
+"""
+    WindingNumberJacobson(; shape=nothing, winding_number_factor=sqrt(eps()),
+                          hierarchical_winding=false)
+Algorithm for inside-outside segmentation of a complex shape proposed by Jacobson et al. (2013).
+
+# Keywords
+- `shape`: Complex shape returned by [`load_shape`](@ref) and is only required when using
+           `hierarchical_winding=true`.
+- `hierarchical_winding`: If set to `true`, an optimised hierarchical approach will be used
```
efaulhaber, line 37:
```suggestion
- `hierarchical_winding`: If set to `true`, an optimized hierarchical approach will be used
```

src/preprocessing/point_in_poly/winding_number_jacobson.jl
```diff
@@ -0,0 +1,136 @@
...
+                          which gives a significant speedup.
+                          It is only supported for 3D shapes yet.
+- `winding_number_factor`: For leaky shapes, a factor of `0.4` will give a better inside-outside segmentation.
+"""
+struct WindingNumberJacobson{ELTYPE}
+    winding_number_factor :: ELTYPE
+    winding               :: Union{NaiveWinding, HierarchicalWinding}
+
+    function WindingNumberJacobson(; shape=nothing, winding_number_factor=sqrt(eps()),
```
efaulhaber, line 46:
```suggestion
                          It is only supported for 3D geometries yet.
- `winding_number_factor`: For leaky geometries, a factor of `0.4` will give a better inside-outside segmentation.
"""
struct WindingNumberJacobson{ELTYPE}
    winding_number_factor :: ELTYPE
    winding               :: Union{NaiveWinding, HierarchicalWinding}

    function WindingNumberJacobson(; geometry=nothing, winding_number_factor=sqrt(eps()),
```
As discussed earlier, please rename shape to geometry everywhere, except for the `ComplexShape`.

src/preprocessing/point_in_poly/winding_number_horman.jl
```diff
@@ -0,0 +1,77 @@
...
+It is only supported for 2D shapes.
+[`WindingNumberHorman`](@ref) might handle edge cases a bit better, since the winding number is an integer value.
+Also, it is faster than [`WindingNumberJacobson`](@ref) for 2D shapes with about more than 100 edges.
```
efaulhaber, line 7:
```suggestion
Also, it is faster than [`WindingNumberJacobson`](@ref) for 2D geometries with more than about 100 edges.
```

src/preprocessing/point_in_poly/winding_number_jacobson.jl
```diff
@@ -0,0 +1,136 @@
...
+- `shape`: Complex shape returned by [`load_shape`](@ref) and is only required when using
+           `hierarchical_winding=true`.
+- `hierarchical_winding`: If set to `true`, an optimised hierarchical approach will be used
+                          which gives a significant speedup.
+                          It is only supported for 3D shapes yet.
+- `winding_number_factor`: For leaky shapes, a factor of `0.4` will give a better inside-outside segmentation.
+"""
+struct WindingNumberJacobson{ELTYPE}
+    winding_number_factor :: ELTYPE
+    winding               :: Union{NaiveWinding, HierarchicalWinding}
```
efaulhaber, line 44: Not type-stable!

src/preprocessing/point_in_poly/winding_number_jacobson.jl
```diff
@@ -0,0 +1,136 @@
...
+
+    winding_numbers = Float64[]
+    store_winding_number && (winding_numbers = resize!(winding_numbers, length(inpoly)))
+
+    @threaded points for query_point in axes(points, 2)
+        p = point_position(points, mesh, query_point)
+
+        winding_number = winding(mesh, p)
+
+        winding_number /= 4pi
```
efaulhaber, line 72:
```suggestion
        winding_number = winding(mesh, p) / 4pi
```

src/preprocessing/point_in_poly/winding_number_jacobson.jl
```diff
@@ -0,0 +1,136 @@
...
+        if !(-winding_number_factor < winding_number < winding_number_factor)
+            inpoly[query_point] = true
+        end
+    end
+
+    return inpoly, winding_numbers
+end
+
+# The following functions distinguish between actual triangles and reconstructed triangles
+# in the hierarchical winding approach
```
efaulhaber, line 119:
```suggestion
# in the hierarchical winding approach.
```

src/preprocessing/point_in_poly/winding_number_jacobson.jl
```diff
@@ -0,0 +1,136 @@
...
+    (; face_vertices) = mesh
+
+    return naive_winding(mesh, face_vertices, query_point)
+end
+
+@inline function naive_winding(mesh, faces, query_point)
+    winding_number = sum(faces, init=0.0) do face
+
+        # A. Van Oosterom 1983,
+        # The Solid Angle of a Plane Triangle (doi: 10.1109/TBME.1983.325207)
```
efaulhaber, line 13:
```suggestion
        # Eq. 6 of Jacobsen et al. Based on A. Van Oosterom (1983),
        # "The Solid Angle of a Plane Triangle" (doi: 10.1109/TBME.1983.325207)
```

src/preprocessing/point_in_poly/winding_number_jacobson.jl
```diff
@@ -0,0 +1,136 @@
...
+        a = face_vertex(mesh, face, 1) - query_point
+        b = face_vertex(mesh, face, 2) - query_point
+        c = face_vertex(mesh, face, 3) - query_point
+        a_ = norm(a)
+        b_ = norm(b)
+        c_ = norm(c)
+
+        divisor = a_ * b_ * c_ + dot(a, b) * c_ + dot(b, c) * a_ + dot(c, a) * b_
+
+        return 2atan(det([a b c]), divisor)
```
efaulhaber, line 23:
```suggestion
        return 2 * atan(det([a b c]) / divisor)
```
1. I find it easier to read with `2 * `
2. They just use a regular arctan in the paper, which is almost 2x faster.

src/preprocessing/point_in_poly/winding_number_jacobson.jl
```diff
@@ -0,0 +1,136 @@
...
+    store_winding_number && (winding_numbers = resize!(winding_numbers, length(inpoly)))
+
+    @threaded points for query_point in axes(points, 2)
+        p = point_position(points, mesh, query_point)
+
+        winding_number = sum(edge_vertices, init=0.0) do edge
+            a = edge[1] - p
+            b = edge[2] - p
+
+            return atan(det([a b]), (dot(a, b)))
```
efaulhaber, line 102:
```suggestion
            return atan(det([a b]) / (dot(a, b)))
```
Can this also just be a regular arctan?

src/preprocessing/point_in_poly/hierarchical_winding.jl
```diff
@@ -0,0 +1,238 @@
+# This bounding box is used for the hierarchical evaluation of the winding number.
```
efaulhaber, line 1:
```suggestion
# This bounding box is used for the hierarchical evaluation of the `WindingNumberJacobsen`.
```
Make it clear to everyone that this is only used for the Jacobsen winding number.

efaulhaber (changes requested):
Please also mark all docstrings with our typical "experimental feature" box, so that we can change the API in a minor release.

src/preprocessing/point_in_poly/winding_number_jacobson.jl
LasNikas: we need all quadrants (full range of `-pi < theta <= pi`) for a correct evaluation of the winding number.

src/preprocessing/point_in_poly/winding_number_jacobson.jl
LasNikas: see comment above

src/preprocessing/point_in_poly/winding_number_jacobson.jl
LasNikas: https://en.wikipedia.org/wiki/Atan2

LasNikas: image

efaulhaber (changes requested):

src/preprocessing/point_in_poly/winding_number_jacobson.jl
```diff
@@ -0,0 +1,138 @@
+struct NaiveWinding end
+
+@inline function (winding::NaiveWinding)(mesh, query_point)
+    (; face_vertices) = mesh
+
+    return naive_winding(mesh, face_vertices, query_point)
+end
+
+@inline function naive_winding(mesh, faces, query_point)
+    winding_number = sum(faces, init=0.0) do face
```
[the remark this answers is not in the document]
LasNikas: Yes, for the hierarchical winding (for empty bounding boxes)

src/preprocessing/point_in_poly/hierarchical_winding.jl
```diff
@@ -0,0 +1,239 @@
...
+    if length(bounding_box.closing_faces) >= length(faces)
+        bounding_box.is_leaf[] = true
+
+        return bounding_box
+    end
+
+    # Bisect the box splitting its longest side
+    box_edges = max_corner - min_corner
+
+    split_direction = findfirst(x -> maximum(box_edges) == x, box_edges)
```
efaulhaber, line 79:
```suggestion
    split_direction = argmax(box_edges)
```

src/preprocessing/point_in_poly/hierarchical_winding.jl
```diff
@@ -0,0 +1,239 @@
...
+        v3 = face_vertices_ids[face][3]
+
+        # Flip order of vertices
+        push!(bounding_box.closing_faces, (v3, v2, v1))
+    end
+
+    return bounding_box
+end
+
+function in_bounding_box(mesh, faces, min_corner, max_corner)
```
efaulhaber, line 187:
```suggestion
function faces_in_box(mesh, faces, min_corner, max_corner)
```

src/preprocessing/point_in_poly/hierarchical_winding.jl
```diff
@@ -0,0 +1,239 @@
...
+            end
+            if edge_vertices_ids[edge3] == (v3, v1)
+                count_directed_edge[edge3] += 1
+            else
+                count_directed_edge[edge3] -= 1
+            end
+
+        elseif !in_bounding_box(vertices[v1], min_corner, max_corner) ||
+               !in_bounding_box(vertices[v2], min_corner, max_corner) ||
+               !in_bounding_box(vertices[v3], min_corner, max_corner)
```
efaulhaber, line 142:
```suggestion
        else
```
This is literally the negation of the if condition.

src/preprocessing/point_in_poly/hierarchical_winding.jl
```diff
@@ -0,0 +1,239 @@
...
+            # Face is intersecting the boundaries
+
+            push!(intersecting_faces, face)
+        end
+    end
+
+    closing_edges = findall(!iszero, count_directed_edge)
```
efaulhaber, line 149: Aren't these called exterior edges?

efaulhaber (changes requested):

src/preprocessing/point_in_poly/hierarchical_winding.jl
```diff
@@ -0,0 +1,239 @@
...
+        if v1 == closing_vertex || v2 == closing_vertex
+            continue
+        end
+
+        if count_directed_edge[edge] < 0
+            @inbounds for _ in 1:abs(count_directed_edge[edge])
+                push!(bounding_box.closing_faces, (v1, v2, closing_vertex))
+            end
+        elseif count_directed_edge[edge] > 0
+            @inbounds for _ in 1:abs(count_directed_edge[edge])
```
efaulhaber, line 169:
```suggestion
            for _ in 1:abs(count_directed_edge[edge])
                push!(bounding_box.closing_faces, (v1, v2, closing_vertex))
            end
        elseif count_directed_edge[edge] > 0
            for _ in 1:abs(count_directed_edge[edge])
```

src/preprocessing/point_in_poly/hierarchical_winding.jl
```diff
@@ -0,0 +1,239 @@
...
+function in_bounding_box(mesh, faces, min_corner, max_corner)
+    faces_in_bbox = Int[]
+
+    for face in faces
+        if in_bounding_box(barycenter(mesh, face), min_corner, max_corner)
+            push!(faces_in_bbox, face)
+        end
+    end
+
+    return faces_in_bbox
```
efaulhaber, line 196:
```suggestion
    return filter(face -> in_bounding_box(barycenter(mesh, face), min_corner, max_corner), faces)
```

efaulhaber (changes requested):

test/preprocessing/geometries/geometries.jl
```diff
@@ -0,0 +1,189 @@
...
+                rot([0.0, 1.0], rot_angle),
+                rot([1.0, 0.0], rot_angle),
+                rot([0.0, -1.0], rot_angle),
+            ]
+
+            vertex_normals = [
+                rot([[-√2 / 2, -√2 / 2], [-√2 / 2, √2 / 2]], rot_angle), # edge 1
+                rot([[-√2 / 2, √2 / 2], [√2 / 2, √2 / 2]], rot_angle),   # edge 2
+                rot([[√2 / 2, √2 / 2], [√2 / 2, -√2 / 2]], rot_angle),   # edge 3
+                rot([[√2 / 2, -√2 / 2], [-√2 / 2, -√2 / 2]], rot_angle), # edge 4
```
efaulhaber, line 27: Please only use ASCII characters in code.

src/preprocessing/point_in_poly/hierarchical_winding.jl
```diff
@@ -0,0 +1,239 @@
+# This bounding box is used for the hierarchical evaluation of the `WindingNumberJacobsen`.
+# It is implementing a binary tree and thus stores the left and right child and also the
+# faces and closing faces which are inside the bounding box.
+struct BoundingBoxTree{MC}
+    faces         :: Vector{Int}
+    min_corner    :: MC
+    max_corner    :: MC
+    is_leaf       :: Ref{Bool}
+    closing_faces :: Vector{NTuple{3, Int}}
+    children      :: Vector{BoundingBoxTree}
```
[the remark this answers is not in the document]
LasNikas: I made it a `Vector{BoundingBoxTree}`, because we then can still use a immutable struct.

src/preprocessing/point_in_poly/hierarchical_winding.jl
LasNikas: Hm I overload the function and use the method also to check if the barycenter is in the box.

LasNikas: How about `is_in_box` ?

efaulhaber: Sounds good to me.

docs/src/preprocessing.md
```diff
@@ -0,0 +1,16 @@
...
+```
+```@autodocs
+Modules = [TrixiParticles]
+Pages = map(file -> joinpath("preprocessing", file), readdir(joinpath("..", "src", "preprocessing", "geometries")))
+```
+
+### [References](@id references_complex_shape)
+- Alec Jacobson, Ladislav Kavan, and Olga Sorkine-Hornung "Robust inside-outside segmentation using generalized winding numbers".
+  In: ACM Transactions on Graphics, 32.4 (2013), pages 1--12.
+  [doi: 10.1145/2461912.2461916](https://doi.org/10.1145/2461912.2461916)
```
efaulhaber, line 13: Please add the URL of the updated version.

src/TrixiParticles.jl
```diff
@@ -6,8 +6,10 @@ using Adapt: Adapt
 using CSV: CSV
 using Dates
 using DataFrames: DataFrame
+using DelimitedFiles
```
efaulhaber, line 9: Please only import names that you actually use.

src/TrixiParticles.jl
```diff
@@ -6,8 +6,10 @@ using Adapt: Adapt
...
 using DiffEqCallbacks: PeriodicCallback, PeriodicCallbackAffect, PresetTimeCallback
 using FastPow: @fastpow
+using FileIO: FileIO, File, query, @format_str, Stream, stream
```
efaulhaber, line 12: 1. I would not import `query`, and instead use it as `FileIO.query`, as not to confuse users where `query` is defined.
2. `@format_str` is not used, is it?
3. Since `File`, `Stream` and `stream` are only used once or twice, you might want to think about calling these explicitly as `FileIO.` as well to make things clearer.

src/preprocessing/point_in_poly/hierarchical_winding.jl
```diff
@@ -0,0 +1,217 @@
...
+                directed_edges[edge2] -= 1
+            end
+            if edge_vertices_ids[edge3] == (v3, v1)
+                directed_edges[edge3] += 1
+            else
+                directed_edges[edge3] -= 1
+            end
+
+        else # Face is intersecting the boundaries
+            push!(intersecting_faces, face)
```
efaulhaber, line 131:
```suggestion
            # Remove intersecting faces in the calculation of the exterior edges.
            # This way, all exterior edges are inside the bounding box, and so will be the closing surface.
            # The intersecting faces are later added with opposite orientation,
            # so that the closing is actually a closing of the exterior plus intersecting faces.
            push!(intersecting_faces, face)
```

src/preprocessing/point_in_poly/hierarchical_winding.jl
```diff
@@ -0,0 +1,217 @@
...
+                push!(closing_faces, (v1, v2, closing_vertex))
+            end
+        elseif directed_edges[edge] > 0
+            for _ in 1:abs(directed_edges[edge])
+                push!(closing_faces, (v2, v1, closing_vertex))
+            end
+        end
+    end
+
+    for face in intersecting_faces
```
efaulhaber, line 161:
```suggestion
    # See comment above as to why intersecting faces are treated seperately
    for face in intersecting_faces
```

src/preprocessing/point_in_poly/winding_number_jacobson.jl
```diff
@@ -0,0 +1,124 @@
...
+
+"""
+    WindingNumberJacobson(; geometry=nothing, winding_number_factor=sqrt(eps()),
+                          hierarchical_winding=false)
+Algorithm for inside-outside segmentation of a complex geometry proposed by Jacobson et al. (2013).
+
+# Keywords
+- `geometry`: Complex geometry returned by [`load_geometry`](@ref) and is only required when using
+              `hierarchical_winding=true`.
+- `hierarchical_winding`: If set to `true`, an optimized hierarchical approach will be used
```
efaulhaber, line 54:
```suggestion
- `hierarchical_winding`: If set to `true`, an optimized hierarchical approach will be used,
```

src/preprocessing/point_in_poly/winding_number_jacobson.jl
```diff
@@ -0,0 +1,124 @@
...
+
+    return mesh.vertices[v_id]
+end
+
+# `face` is the index of the face
+@inline function face_vertex(mesh, face::Int, index)
+    (; face_vertices) = mesh
+
+    return face_vertices[face][index]
+end
```
efaulhaber, line 124: Can you please clarify why and when this is needed?

src/setups/complex_shape.jl
```diff
@@ -0,0 +1,103 @@
+"""
+    ComplexShape(geometry::Union{TriangleMesh, Polygon}; particle_spacing, density,
+                 pressure=0.0, mass=nothing, velocity=zeros(ndims(geometry)),
+                 point_in_geometry_algorithm=WindingNumberJacobson(; geometry,
+                                                                hierarchical_winding=false,
+                                                                winding_number_factor=sqrt(eps())),
+                 grid_offset::Real=0.0, max_nparticles=10^7,
+                 pad_initial_particle_grid=2particle_spacing)
+
+Complex geometry filled with particles. Returns an [`InitialCondition`](@ref).
```
efaulhaber, line 10:
```suggestion
Sample a complex geometry with particles. Returns an [`InitialCondition`](@ref).
```

src/setups/complex_shape.jl
```diff
@@ -0,0 +1,103 @@
...
+Note that an initial particle grid is generated inside the bounding box of the geometry.
+An `point_in_geometry_algorithm` checks if particles are inside the geometry or not.
```
efaulhaber, line 12:
```suggestion
A `point_in_geometry_algorithm` checks if particles are inside the geometry or not.
```

src/setups/complex_shape.jl
```diff
@@ -0,0 +1,103 @@
...
+For more information about the method see [`WindingNumberJacobson`](@ref) or [`WindingNumberHorman`](@ref).
+
+# Arguments
+- `geometry`: Complex geometry returned by [`load_geometry`](@ref).
```
efaulhaber, line 16:
```suggestion
- `geometry`: Geometry returned by [`load_geometry`](@ref).
```

src/setups/complex_shape.jl
```diff
@@ -0,0 +1,103 @@
...
+- `velocity`:           Either a function mapping each particle's coordinates to its velocity,
+                        or, for a constant fluid velocity, a vector holding this velocity.
+                        Velocity is constant zero by default.
+- `mass`:               Either `nothing` (default) to automatically compute particle mass from particle
+                        density and spacing, or a function mapping each particle's coordinates to its mass,
+                        or a scalar for a constant mass over all particles.
+- `pressure`:           Scalar to set the pressure of all particles to this value.
+                        This is only used by the [`EntropicallyDampedSPHSystem`](@ref) and
+                        will be overwritten when using an initial pressure function in the system.
+                        Cannot be used together with hydrostatic pressure gradient.
```
efaulhaber, line 31:
```suggestion
```

examples/preprocessing/complex_shape_2d.jl
```diff
@@ -0,0 +1,15 @@
+using TrixiParticles
+
+particle_spacing = 0.05
+
+file = "hexagon"
+filename = joinpath("examples", "preprocessing", file * ".asc")
```
efaulhaber, line 6:
```suggestion
filename = "hexagon"
file = joinpath("examples", "preprocessing", file * ".asc")
```
I would switch those names.

examples/preprocessing/complex_shape_2d.jl
```diff
@@ -0,0 +1,15 @@
...
+
+geometry = load_geometry(filename)
+
+trixi2vtk(geometry)
+
+# Returns `InitialCondition`.
```
efaulhaber, line 12:
```suggestion
# Returns `InitialCondition`
```

examples/preprocessing/complex_shape_3d.jl
```diff
@@ -0,0 +1,22 @@
+using TrixiParticles
+
+particle_spacing = 0.05
+
+file = "sphere"
+filename = joinpath("examples", "preprocessing", file * ".stl")
+
+# The following triangle mesh is corrupt.
+# For more robustness, use `winding_number_factor=0.4`.
+# filename = joinpath("examples", "preprocessing", "drive_gear.stl")
```
efaulhaber, line 10:
```suggestion
filename = "sphere"

# The following triangle mesh is corrupt.
# For more robustness, use `winding_number_factor=0.4`.
# filename = "drive_gear"

file = joinpath("examples", "preprocessing", file * ".stl")
```

examples/preprocessing/drive_gear.stl
efaulhaber, line 1: What is a "drive gear"?

test/preprocessing/data/drive_gear.stl
efaulhaber, line 1: Duplicate files? This file and all of the following also appear in `src`.

efaulhaber (changes requested):
Only tests remaining.

src/preprocessing/point_in_poly/hierarchical_winding.jl
```diff
@@ -5,29 +5,61 @@ struct BoundingBoxTree{MC}
...
-    function BoundingBoxTree(faces, min_corner, max_corner)
+    function BoundingBoxTree(geometry, faces, directed_edges, min_corner, max_corner)
         closing_faces = Vector{NTuple{3, Int}}()
-        children = Vector{BoundingBoxTree}()
 
-        return new{typeof(min_corner)}(faces, min_corner, max_corner, Ref(false),
-                                       closing_faces, children)
+        if length(faces) < 100
+            return new{typeof(min_corner)}(faces, min_corner, max_corner, true,
+                                           closing_faces)
```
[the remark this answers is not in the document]
LasNikas: No, if it's a leaf, then we don't use the closure

image

examples/preprocessing/complex_shape_3d.jl
LasNikas: This comment is obsolete. I'll remove it

test/preprocessing/data/drive_gear.stl
LasNikas: as I mentioned above, it's for the CI

src/preprocessing/point_in_poly/hierarchical_winding.jl
```diff
@@ -170,6 +183,67 @@ function determine_closure!(closing_faces, min_corner, max_corner, mesh::Triangl
...
+            # Remove intersecting edges in the calculation of the exterior vertices.
+            # This way, all exterior vertices are inside the bounding box,
+            # and so will be the closing surface.
+            # The intersecting edges are later added with opposite orientation,
+            # so that the closing is actually a closing of the exterior plus intersecting edges.
+            push!(intersecting_edges, edge)
+        end
+    end
+
+    exterior_vertices = sort(findall(!iszero, vertex_count))
```
efaulhaber, line 217: Isn't this always sorted? It's just the indices.

src/preprocessing/point_in_poly/hierarchical_winding.jl
```diff
@@ -0,0 +1,291 @@
+# This bounding box is used for the hierarchical evaluation of the `WindingNumberJacobsen`.
+# It is implementing a binary tree and thus stores the left and right child and also the
+# faces and closing faces which are inside the bounding box.
+struct BoundingBoxTree{MC, NDIMS}
+    faces         :: Vector{Int}
+    min_corner    :: MC
+    max_corner    :: MC
+    is_leaf       :: Bool
+    closing_faces :: Vector{NTuple{NDIMS, Int}}
```
efaulhaber, line 9: I find it confusing that `faces` is a collection of face indices of the `mesh`, while `closing_faces` is a collection of tuples of vertex indices of the `mesh`.
Can you make both collections of tuples, or would that hurt performance?

efaulhaber (changes requested):

NEWS.md
```diff
@@ -4,11 +4,12 @@ TrixiParticles.jl follows the interpretation of [semantic versioning (semver)](h
 used in the Julia ecosystem. Notable changes will be documented in this file for human readability.
 We aim at 3 to 4 month between major release versions and about 2 weeks between minor versions.
 
-## Version 0.3.x
```
svchb, line 7: ## Version 0.3.x

### Highlights

### Added
Particle sampling for complex geometries from `.stl` and `.asc` files.

### Removed

### Deprecated

## Version 0.2.1

### Highlights
Particle sampling for geometries from `.stl` and `.asc` files.

README.md
```diff
@@ -21,6 +21,7 @@ Its features include:
 - Solid-body mechanics
   - Methods:  Total Lagrangian SPH (TLSPH), Discrete Element Method (DEM)
 - Fluid-Structure Interaction
+- Preprocess complex shapes: Support for STL geometries
```
svchb, line 24:
```suggestion
- Preprocess complex shapes
  - Support for STL files
```

svchb (changes requested):

svchb (approved):

NEWS.md
```diff
@@ -9,17 +9,22 @@ We aim at 3 to 4 month between major release versions and about 2 weeks between
 ### Highlights
 
 ### Added
+Particle sampling for complex geometries from `.stl` and `.asc` files.
```
svchb, line 12:
```suggestion

```

svchb (changes requested):

test/preprocessing/geometries/geometries.jl
```diff
@@ -0,0 +1,186 @@
+@testset verbose=true "Shapes" begin
+    @testset verbose=true "Rectangular Analytical" begin
+        rot(vec, α) = SVector{2}([cos(α) -sin(α); sin(α) cos(α)] * vec)
```
efaulhaber, line 3: Please use ASCII characters in code.

test/preprocessing/geometries/geometries.jl
```diff
@@ -0,0 +1,186 @@
...
+    @testset verbose=true "Real World Data" begin
+        data_dir = pkgdir(TrixiParticles, "examples", "preprocessing", "data")
+
+        @testset verbose=true "2D" begin
+            files = ["hexagon", "circle", "inverted_open_curve"]
+            n_edges = [6, 63, 240]
+
+            @testset "Test File `$(files[i])`" for i in eachindex(files)
+                # Checked in ParaView with `trixi2vtk(geometry)`
+                data = TrixiParticles.CSV.read(joinpath(validation_dir(), "preprocessing",
```
efaulhaber, line 66: The validation directory is for validation of simulation results against literature experiments. Can you please move these files to `test/`?

test/preprocessing/geometries/geometries.jl
```diff
@@ -0,0 +1,186 @@
...
+            if rand(Bool)
+                x[i] = x[i - 1]
+            end
+        end
+
+        @test TrixiParticles.unique_sorted(copy(x)) == sort(unique(x))
+    end
+end
+
+# ==========================================================================================
```
efaulhaber, line 143: I don't think it's a good idea to leave code commented out that might be used later. Once we start doing this, it usually escalates and results in a lot of unused code. How about leaving it in a stash or in a draft PR until you need it?

test/setups/complex_shape.jl
```diff
@@ -0,0 +1,125 @@
+@testset verbose=true "Complex Shapes" begin
```
efaulhaber, line 1:
```suggestion
@testset verbose=true "ComplexShape" begin
```

test/setups/complex_shape.jl
```diff
@@ -0,0 +1,125 @@
...
+    data_dir = pkgdir(TrixiParticles, "examples", "preprocessing", "data")
+
+    @testset verbose=true "Complex Shapes 2D" begin
```
efaulhaber, line 4:
```suggestion
    @testset verbose=true "2D" begin
```

test/setups/complex_shape.jl
```diff
@@ -0,0 +1,125 @@
...
+        @testset verbose=true "Rectangular Shifted" begin
```
efaulhaber, line 5:
```suggestion
        @testset verbose=true "Shifted Rectangle" begin
```

test/setups/complex_shape.jl
```diff
@@ -0,0 +1,125 @@
...
+
+            test_name(algorithm, shift, particle_spacing) = "Algorithm: $(TrixiParticles.type2string(algorithm))" *
+                                                            ", Shift: $shift" *
+                                                            ", Particle Spacing: $particle_spacing"
+            @testset verbose=true "$(test_name(point_in_geometry_algorithm, shift,
+            particle_spacing))" for point_in_geometry_algorithm in algorithms,
+                                    shift in shifts,
+                                    particle_spacing in particle_spacings
+
+                points_rectangular = stack([[0.0, 0.0], [1.0, 0.0],
```
efaulhaber, line 18:
```suggestion
                points_rectangle = stack([[0.0, 0.0], [1.0, 0.0],
```

test/setups/complex_shape.jl
```diff
@@ -0,0 +1,125 @@
...
+                    end
+
+                    data = TrixiParticles.CSV.read(joinpath(validation_dir(),
+                                                            "preprocessing",
+                                                            "coordinates_" *
+                                                            algorithm_names[i] * "_" *
+                                                            files[j] * ".csv"),
+                                                   TrixiParticles.DataFrame)
+
+                    coords = vcat((data.var"Points:0")', (data.var"Points:1")')
```
efaulhaber, line 61:
```suggestion
                    # Access the field called `Points:0` of `data`. Since `data.Points:0` is not a valid
                    # identifier name, we need to use `var"name"`.
                    # See https://docs.julialang.org/en/v1/base/base/#var%22name%22
                    coords = vcat((data.var"Points:0")', (data.var"Points:1")')
```

efaulhaber (changes requested):

src/setups/complex_shape.jl
```diff
@@ -0,0 +1,105 @@
...
+    n_particles_per_dimension = Tuple(ceil.(Int,
+                                            (max_corner .- min_corner .+ 2padding) ./
+                                            particle_spacing))
+
+    n_particles = prod(n_particles_per_dimension)
+
+    if n_particles > max_nparticles
+        throw(ArgumentError("Number of particles of the initial grid ($n_particles) exceeds " *
+                            "`max_nparticles` = $max_nparticles"))
+    end
```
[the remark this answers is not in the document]
LasNikas: If one accidentally choose a very low resolution (or rather the scale of the geometry is way too large for `particle_spacing`), it will take forever.

test/preprocessing/geometries/geometries.jl
```diff
@@ -0,0 +1,186 @@
...
+                @test all(isapprox.(geometry_clockwise.edge_vertices[edge],
+                                    edge_vertices[edge], atol=1e-14))
+            end
+            for vertex in eachindex(vertex_normals)
+                @test all(isapprox.(geometry_clockwise.vertex_normals[vertex],
+                                    vertex_normals[vertex], atol=1e-14))
+            end
+            @test all(isapprox(geometry_clockwise.edge_normals, edge_normals, atol=1e-14))
+
+            geometry_cclockwise = TrixiParticles.Polygon(Matrix(points_rectangular_counter_clockwise))
```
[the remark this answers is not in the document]
LasNikas: It's counter clockwise, therefore the double "c"

src/setups/complex_shape.jl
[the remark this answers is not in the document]
efaulhaber: I see. It might be useful to add a short sentence in the docs. Like "this is only used to avoid accidentally choosing a `particle_spacing` that is too small for the scale of the geometry".

test/setups/complex_shape.jl
```diff
@@ -1,8 +1,9 @@
-@testset verbose=true "Complex Shapes" begin
+@testset verbose=true "ComplexShapes" begin
```
efaulhaber, line 1:
```suggestion
@testset verbose=true "ComplexShape" begin
```
I meant to use the struct name.

test/setups/complex_shape.jl
```diff
@@ -1,8 +1,9 @@
...
     data_dir = pkgdir(TrixiParticles, "examples", "preprocessing", "data")
+    validation_dir = pkgdir(TrixiParticles, "test", "preprocessing", "data")
 
-    @testset verbose=true "Complex Shapes 2D" begin
-        @testset verbose=true "Rectangular Shifted" begin
+    @testset verbose=true "2D" begin
+        @testset verbose=true "Shifted Rectangular" begin
```
efaulhaber, line 6:
```suggestion
        @testset verbose=true "Shifted Rectangle" begin
```

efaulhaber (changes requested):

efaulhaber (approved):

svchb (approved):

PR #441, opened 2024-03-05
Reopen #363: Hydrostatic water column (FSI)

LasNikas:


LasNikas:
@efaulhaber and @svchb do we need this example or should I close this PR?

svchb:
I am fine with adding this example if the results are correct.

efaulhaber:
What is this again? Was it the hydrostatic water column resting on an elastic beam?
IIRC, this was a toy example but pretty useless for validation, as the "analytical solution" used simplifications, so that the error is always quite large and it doesn't validate anything, really.

I'm against merging more examples that we don't really need. I prefer having useful validation examples from the literature only. Trixi.jl has too many examples, which makes it really hard for beginners.

svchb:
Niklas and I as far as I know are not against having as many examples as possible. But of course we would need to restructure the examples folder structure + docs page at some point.

LasNikas:
It's a validation example from Sun et al. (2019), as well as from O'Connor and Rogers (2021).

efaulhaber:
> It's a validation example from Sun et al. (2019), as well as from O'Connor and Rogers (2021).

Please add this as a comment and a description of what that simulation does.

svchb:
Superseeded by #724

PR #440, opened 2024-03-05
Reopen #390: Update callback

LasNikas:
see #390

PR #442, opened 2024-03-05
Add open boundaries

LasNikas:
based on #437 and #440


svchb:
Add to news.md and update examples doc page.

src/general/semidiscretization.jl
```diff
@@ -150,6 +150,17 @@ end
...
+@inline function compact_support(system::OpenBoundarySPHSystem, neighbor)
+    # Use the compact support of the fluid
+    return compact_support(neighbor, system)
+end
+
+@inline function compact_support(system::OpenBoundarySPHSystem,
+                                 neighbor::OpenBoundarySPHSystem)
+    # This NHS is never used
+    return 0.0
+end
```
svchb, line 162: Move below to the BoundaryModelMonaghanKajtar version which returns 0.0.

src/general/semidiscretization.jl
```diff
@@ -623,6 +634,16 @@ function nhs_coords(system::FluidSystem,
     return current_coordinates(u, neighbor)
 end
 
+function nhs_coords(system::FluidSystem,
+                    neighbor::OpenBoundarySPHSystem, u)
+    return current_coordinates(u, neighbor)
+end
```
svchb, line 640: Combine with TLSPH version below.

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,613 @@
...
+```
+"""
+struct OpenBoundarySPHSystem{BZ, NDIMS, ELTYPE <: Real, S, RV, RP, RD, B} <: System{NDIMS}
+    initial_condition        :: InitialCondition{ELTYPE}
+    mass                     :: Array{ELTYPE, 1} # [particle]
+    density                  :: Array{ELTYPE, 1} # [particle]
+    volume                   :: Array{ELTYPE, 1} # [particle]
+    pressure                 :: Array{ELTYPE, 1} # [particle]
+    characteristics          :: Array{ELTYPE, 2} # [characteristics, particle]
+    previous_characteristics :: Array{ELTYPE, 2} # [characteristics, particle]
```
svchb, line 89: Apply #486

svchb (changes requested):
Just looked at the first few files for now.

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,622 @@
...
+        if !(0 <= dot(particle_position, span_dim) <= dot(span_dim, span_dim))
+
+            # Particle is not in boundary zone.
+            return false
+        end
+    end
+
+    # Particle is in boundary zone.
+    return true
+end
```
LasNikas, line 296: FYI: I can make this faster by adding a flag that the boundary zone is axis-aligned and only check the coordinates. However, Idk if this is necessary?

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,622 @@
...
+
+    # Use all other systems for the characteristics
+    foreach_system(semi) do neighbor_system
+        evaluate_characteristics!(system, neighbor_system, v, u, v_ode, u_ode, semi, t)
+    end
+
+    # Only some of the in-/outlet particles are in the influence of the fluid particles.
+    # Thus, we find the characteristics for the particle which are outside the influence
+    # using the average of the values of the previous time step.
+    # Negi (2020) https://doi.org/10.1016/j.cma.2020.113119
```
LasNikas, line 348: eq. 27 in https://doi.org/10.1016/j.cma.2020.113119

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,622 @@
...
+                   typeof(buffer)}(initial_condition, mass, density, volume, pressure,
+                                   characteristics, previous_characteristics, sound_speed,
+                                   boundary_zone, flow_direction_, zone_origin,
+                                   spanning_set_, reference_velocity_, reference_pressure_,
+                                   reference_density_, buffer, false)
+    end
+end
+
+timer_name(::OpenBoundarySPHSystem) = "open_boundary"
+vtkname(system::OpenBoundarySPHSystem) = "open_boundary"
```
svchb, line 202: move to file_system.jl

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,622 @@
...
+                   typeof(reference_density_),
+                   typeof(buffer)}(initial_condition, mass, density, volume, pressure,
+                                   characteristics, previous_characteristics, sound_speed,
+                                   boundary_zone, flow_direction_, zone_origin,
+                                   spanning_set_, reference_velocity_, reference_pressure_,
+                                   reference_density_, buffer, false)
+    end
+end
+
+timer_name(::OpenBoundarySPHSystem) = "open_boundary"
```
svchb, line 201: Move to general.jl

svchb (changes requested):

src/general/buffer.jl
```diff
@@ -0,0 +1,90 @@
...
+@inline function deactivate_particle!(system, particle, u)
+    (; active_particle) = system.buffer
+
+    active_particle[particle] = false
+
+    # Set particle far away from simulation domain
+    for dim in 1:ndims(system)
+        # Inf or NaN causes instability outcome.
+        u[dim, particle] = inv(eps())
+    end
```
efaulhaber, line 87:
```suggestion
@inline function deactivate_particle!(system, particle)
    system.buffer.active_particle[particle] = false
```
According to our discussion if large coordinates are necessary.

src/general/buffer.jl
```diff
@@ -0,0 +1,90 @@
...
+end
+
+allocate_buffer(initial_condition, buffer) = initial_condition
+
+function allocate_buffer(initial_condition, buffer::SystemBuffer)
+    (; buffer_size) = buffer
+
+    # Initialize particles far away from simulation domain
+    coordinates = inv(eps()) * ones(ndims(initial_condition), buffer_size)
```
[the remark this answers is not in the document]
LasNikas: `write_u0` writes inactive and active particles into `u`. So,  `Inf` or `NaN` causes instability outcome.

The `active_particles` function is to write only active particles to vtk to overcome visualization problems.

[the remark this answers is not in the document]
LasNikas: We could use `active_particles` for the nhs but we then need to remove inactive particles from the hashtables (neighboring lists).

For now, we can stick with large particle coordinates which is okay for "normal" simulations. But, I added a `TODO` in `update_nhs!` to keep that in mind.

src/general/buffer.jl
LasNikas: see comment above

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -548,34 +548,23 @@ end
...
-    set_particle_pressure(particle_new, v_new, system_new, pressure)
-
-    # Exchange position and velocity
-    for dim in 1:ndims(system_new)
-        u_new[dim, particle_new] = u_old[dim, particle_old]
-        v_new[dim, particle_new] = v_old[dim, particle_old]
+    # Exchange integrated quantities
+    for i in axes(u_new, 1)
+        u_new[i, particle_new] = u_old[i, particle_old]
+        v_new[i, particle_new] = v_old[i, particle_old]
```
efaulhaber, line 565: You need two loops. One over `axes(u_new, 1)` for `u` and one over `axes(v_new, 1)` for `v` to include density/pressure.

LasNikas: Oh right thanks!

src/schemes/fluid/entropically_damped_sph/system.jl
```diff
@@ -62,7 +63,10 @@ struct EntropicallyDampedSPHSystem{NDIMS, ELTYPE <: Real, IC, M, DC, K, V,
                                          alpha=0.5, viscosity=nothing,
                                          acceleration=ntuple(_ -> 0.0,
                                                              ndims(smoothing_kernel)),
-                                         source_terms=nothing)
+                                         source_terms=nothing, buffer=nothing)
+        (buffer ≠ nothing) && (buffer = SystemBuffer(nparticles(initial_condition), buffer))
```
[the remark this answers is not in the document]
efaulhaber: Wait, apparently I misunderstood something here.
I thought the buffer is just to mark particles in the system as active and inactive. Therefore, shouldn't the buffer size always be exactly the same as the system size?

src/schemes/fluid/viscosity.jl
```diff
@@ -11,6 +11,18 @@ function dv_viscosity(particle_system, neighbor_system,
                         sound_speed, m_a, m_b, rho_mean)
 end
 
+function dv_viscosity(particle_system, neighbor_system::OpenBoundarySPHSystem,
+                      v_particle_system, v_neighbor_system,
+                      particle, neighbor, pos_diff, distance,
+                      sound_speed, m_a, m_b, rho_mean)
+    viscosity = viscosity_model(particle_system)
```
efaulhaber, line 18:
```suggestion
    # No viscosity in the open boundary system. Use viscosity of the fluid system.
    viscosity = viscosity_model(particle_system)
```

src/schemes/schemes.jl
```diff
@@ -1,3 +1,7 @@
+# Include open boundary system first,
+# since this system should not depend on specific systems.
+include("boundary/open_boundary/system.jl")
```
efaulhaber, line 3: That's not a reason to force some include order. Do any of the other systems depend on the open boundary system? If not, this should be irrelevant.

src/visualization/write2vtk.jl
```diff
@@ -181,11 +181,12 @@ function write2vtk!(vtk, v, u, t, system; write_meta_data=true)
...
 function write2vtk!(vtk, v, u, t, system::FluidSystem; write_meta_data=true)
-    vtk["velocity"] = view(v, 1:ndims(system), :)
+    vtk["velocity"] = [current_velocity(v, system, particle)
+                       for particle in each_moving_particle(system)]
     vtk["density"] = [particle_density(v, system, particle)
-                      for particle in eachparticle(system)]
+                      for particle in each_moving_particle(system)]
     vtk["pressure"] = [particle_pressure(v, system, particle)
-                       for particle in eachparticle(system)]
+                       for particle in each_moving_particle(system)]
```
efaulhaber, line 189: The coordinates are those of `active_particles`, so this here should use the exact same iterator.
For now, we don't have any active fixed particles for the fluid, so it's equivalent, but this might change.

Same below.

efaulhaber (changes requested):
Docs, examples, tests and the actual open boundary `system.jl` are left for review.

src/schemes/fluid/entropically_damped_sph/system.jl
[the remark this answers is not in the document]
LasNikas: > Therefore, shouldn't the buffer size always be exactly the same as the system size?

Yes, but initially the `InitialCondition` defines the size of the system. By passing a buffer size, one can increase the `InitialCondition` by recreating/reallocating them and thus increase the system size with inactive particles. The problem is, that I can't just `resize!` the `InitialCOndition`s since arrays are used.

src/schemes/boundary/system.jl
```diff
@@ -27,8 +28,7 @@ struct BoundarySPHSystem{BM, NDIMS, IC, CO, M, IM, CA} <: BoundarySystem{NDIMS}
             typeof(initial_condition), typeof(coordinates),
             typeof(movement), typeof(ismoving), typeof(cache)}(initial_condition,
                                                                coordinates, boundary_model,
-                                                               movement,
-                                                               ismoving, cache)
+                                                               movement, ismoving, cache)
```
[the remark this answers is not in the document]
LasNikas: I can add the `buffer` in the constructor for consistency, but it is not necessary if the field `buffer` is declared as `Nothing`.

I would omit it in the constructor since it is just a dummy field for these systems.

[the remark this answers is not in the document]
LasNikas: Thinking more about it, it might be better to add an extra `nothing`, since this only works, when `buffer` is the last field.

src/schemes/boundary/open_boundary/system.jl
LasNikas: Ah now I remember why I didn't do it that way. 

`v_new` doesn't have the same size as `v_old` because density and pressure of the open boundary particles is extrapolated from the fluid domain and not evolved over time...

src/schemes/boundary/system.jl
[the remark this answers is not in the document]
efaulhaber: I didn't know this was possible:
```julia
julia> struct B
           a::Float64
           b::Nothing

           function B(a)
               new(a)
           end
       end

julia> B(1.0)
B(1.0, nothing)
```
Is this documented somewhere?
I guess it's best to explicitly add `nothing` and not rely on this feature.

src/schemes/schemes.jl
efaulhaber: Can't you just include `boundary/boundary.jl` before `fluid/fluid.jl`?

LasNikas:
```julia
include("boundary/open_boundary/system.jl")
include("fluid/fluid.jl")
include("boundary/boundary.jl")
```
is possible. 
But `include("boundary/boundary.jl")` before fluid is same problem  - viscosity dependencies

src/general/density_calculators.jl
```diff
@@ -35,6 +35,16 @@ end
     return v[end, particle]
 end
 
+# *Note* that these functions are intended to internally set the density for buffer particles
+# and density correction. It cannot be used to set up an initial condition,
+# as the particle density depends on the particle positions.
+
+@inline set_particle_density(particle, v, ::SummationDensity, system, density) = system
```
efaulhaber, line 42:
```suggestion
# WARNING!
# These functions are intended to be used internally to set the density
# of newly activated particles in a callback.
# DO NOT use outside a callback. OrdinaryDiffEq does not allow changing `v` and `u`
# outside of callbacks.
@inline set_particle_density!(v, system, ::SummationDensity, particle, density) = v
```
1. Made this a more aggressive warning.
2. This is only allowed to be used inside callbacks.
3. Should have a `!` appended.
4. Changed order of arguments to be consistent with other functions. (And it should now return `v`, as I made it the first argument.)

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -548,23 +548,34 @@ end
...
-@inline function activate_particle!(system_new, particle_old, v_new, u_new, v_old, u_old)
+@inline function activate_particle!(system_new, system_old, particle_old,
+                                    v_new, u_new, v_old, u_old)
     particle_new = activate_next_particle(system_new)
 
-    # Exchange integrated quantities
-    for i in axes(u_new, 1)
-        u_new[i, particle_new] = u_old[i, particle_old]
-        v_new[i, particle_new] = v_old[i, particle_old]
+    # Exchange densities
```
efaulhaber, line 563:
```suggestion
    # Transfer densities
```
This is only in one direction, so it's a transfer and not an exchange.

Shall we rename the function too? I was very confused that `activate_particle!` needs two systems. It's not just activating a particle but transferring it from one system to another. So maybe `transfer_particle!`? I'm open for suggestions.

src/schemes/schemes.jl
```diff
@@ -1,9 +1,6 @@
-# Include open boundary system first,
-# since this system should not depend on specific systems.
-include("boundary/open_boundary/system.jl")
-
 # Include all schemes without rhs first. The rhs depends on the systems to define
 # interactions between the different system types.
```
efaulhaber, line 2:
```suggestion
# interactions between the different system types.
# Viscosity requires the open boundary system.
```
Or is it only the viscosity? Could you just include the viscosity later where the rhs is included? After `Include rhs for all schemes`.

efaulhaber (changes requested):

src/schemes/boundary/open_boundary/system.jl
LasNikas: Very good point! `transfer_quantities!` ?

efaulhaber: You're actually activating and deactivating particles, so I'd prefer `transfer_particle!` to `transfer_quantities!`

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -0,0 +1,275 @@
...
+end
+
+function spanning_vectors(plane_points::NTuple{3}, zone_width)
+    # Vectors spanning the plane
+    edge1 = plane_points[2] - plane_points[1]
+    edge2 = plane_points[3] - plane_points[1]
+
+    if !isapprox(dot(edge1, edge2), 0.0, atol=1e-7)
+        throw(ArgumentError("the provided points do not span a rectangular plane"))
+    end
```
efaulhaber, line 244:
```suggestion
    if !isapprox(dot(edge1, edge2), 0.0, atol=1e-7)
        throw(ArgumentError("the vectors AB and AC for the provided points A, B, C must be orthogonal"))
    end
```

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -0,0 +1,275 @@
...
+                   typeof(flow_direction_)}(ic, spanning_set_, zone_origin, zone_width,
+                                            flow_direction_)
+    end
+end
+
+@inline Base.ndims(::Union{InFlow{NDIMS}, OutFlow{NDIMS}}) where {NDIMS} = NDIMS
+
+# function calculate_spanning_vectors(plane::Shapes, zone_width)
+#     # TODO: Handle differently
+# end
```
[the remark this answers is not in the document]
LasNikas: No, I want to pass a `.stl` file which contains a in-/outflow plane like the red plane in the image:
image

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -0,0 +1,275 @@
...
+
+    # Convert to tuple
+    return spanning_vectors(tuple(plane_points...), zone_width)
+end
+
+function spanning_vectors(plane_points::NTuple{2}, zone_width)
+    plane_size = plane_points[2] - plane_points[1]
+
+    # Calculate normal vector of plane
+    b = Vector(normalize([-plane_size[2]; plane_size[1]]) * zone_width)
```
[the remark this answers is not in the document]
LasNikas: ah nvm...

src/schemes/boundary/open_boundary/boundary_zones.jl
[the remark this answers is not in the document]
efaulhaber: It would still make more sense IMO to remove this and create an issue.

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -1,21 +1,47 @@
...
            initial_condition=nothing, extrude_geometry=nothing,
-           open_boundary_layers::Integer=0)
-
-Inflow boundary zone for [`OpenBoundarySPHSystem`](@ref)
+           open_boundary_layers::Integer)
+
+Inflow boundary zone for [`OpenBoundarySPHSystem`](@ref).
+
+The specified plane (line in 2D or rectangle in 3D) will be extruded in upstream
+direction to create a box for the boundary zone.
```
efaulhaber, line 9:
```suggestion
direction (the direction opposite to `flow_direction`) to create a box for the boundary zone.
```

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -104,24 +129,50 @@ struct InFlow{NDIMS, IC, S, ZO, ZW, FD}
...
 end
 
-"""
+@doc raw"""
     OutFlow(; plane, flow_direction, density, particle_spacing,
             initial_condition=nothing, extrude_geometry=nothing,
-            open_boundary_layers::Integer=0)
+            open_boundary_layers::Integer)
 
 Outflow boundary zone for [`OpenBoundarySPHSystem`](@ref)
```
efaulhaber, line 137:
```suggestion
Outflow boundary zone for [`OpenBoundarySPHSystem`](@ref).
```

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -1,12 +1,12 @@
...
 
 """
-    OpenBoundarySPHSystem(plane_points, boundary_zone::Union{InFlow, OutFlow},
-                          sound_speed;
-                          sample_geometry=plane_points, particle_spacing,
-                          flow_direction, open_boundary_layers::Integer=0, density,
-                          buffer=nothing, reference_velocity=zero(flow_direction),
-                          reference_pressure=0.0, reference_density=density)
+    OpenBoundarySPHSystem(boundary_zone::Union{InFlow, OutFlow}, sound_speed;
+                          buffer=nothing,
```
efaulhaber, line 5:
```suggestion
                          buffer_size=0,
```

efaulhaber (changes requested):
Can you please add
```
```
to buffer, boundary zones and open boundary system?
Then we can change the API in the future without having to release a new major version.

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,508 @@
...
+
+    v_neighbor_system = wrap_v(v_ode, neighbor_system, semi)
+    u_neighbor_system = wrap_u(u_ode, neighbor_system, semi)
+
+    nhs = get_neighborhood_search(system, neighbor_system, semi)
+
+    system_coords = current_coordinates(u, system)
+    neighbor_coords = current_coordinates(u_neighbor_system, neighbor_system)
+
+    # Loop over all fluid neighbors within the kernel cutoff.
```
efaulhaber, line 285:
```suggestion
    # Loop over all fluid neighbors within the kernel cutoff
```

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,508 @@
...
+
+        # Determine current and prescribed quantities
+        rho = particle_density(v_neighbor_system, neighbor_system, neighbor)
+        rho_ref = reference_density(neighbor_position, t)
+
+        p = particle_pressure(v_neighbor_system, neighbor_system, neighbor)
+        p_ref = reference_pressure(neighbor_position, t)
+
+        v_neighbor = current_velocity(v_neighbor_system, neighbor_system, neighbor)
+        v_neighbor_ref = reference_velocity(neighbor_position, t)
```
efaulhaber, line 298:
```suggestion
        rho_b = particle_density(v_neighbor_system, neighbor_system, neighbor)
        rho_ref = reference_density(neighbor_position, t)

        p_b = particle_pressure(v_neighbor_system, neighbor_system, neighbor)
        p_ref = reference_pressure(neighbor_position, t)

        v_b = current_velocity(v_neighbor_system, neighbor_system, neighbor)
        v_ref = reference_velocity(neighbor_position, t)
```

src/schemes/boundary/open_boundary/system.jl
efaulhaber: Then please add this to the comment.

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,508 @@
...
+
+@inline function prescribe_conditions!(characteristics, particle, ::InFlow)
+    # Allow only J3 to propagate upstream to the boundary
+    characteristics[1, particle] = zero(eltype(characteristics))
+    characteristics[2, particle] = zero(eltype(characteristics))
+
+    return characteristics
+end
+
+@inline function update_quantities!(system::OpenBoundarySPHSystem, v, u, t)
```
efaulhaber, line 334: What exactly is happening here and why? Can you please add a comment?

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,508 @@
...
+    for particle in each_moving_particle(system)
+        foreach_system(semi) do fluid_system
+            check_fluid_domain!(system, fluid_system, particle, v, u, v_ode, u_ode, semi)
+        end
+    end
+end
+
+function check_fluid_domain!(system, neighbor_system, particle, v, u, v_ode, u_ode, semi)
+    return system
+end
```
efaulhaber, line 374: Why the two different function names?

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,508 @@
...
+    (; boundary_zone) = system
+
+    particle_coords = current_coords(u, system, particle)
+
+    u_fluid = wrap_u(u_ode, fluid_system, semi)
+    v_fluid = wrap_v(v_ode, fluid_system, semi)
+
+    neighborhood_search = get_neighborhood_search(system, fluid_system, semi)
+
+    # Check if the particle position is outside the boundary zone.
```
efaulhaber, line 387:
```suggestion
    # Check if the particle position is outside the boundary zone
```

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,508 @@
...
+
+    return system
+end
+
+function check_domain!(system, v, u, v_ode, u_ode, semi)
+    # TODO: Is a thread supported version possible?
+    for particle in each_moving_particle(system)
+        foreach_system(semi) do fluid_system
+            check_fluid_domain!(system, fluid_system, particle, v, u, v_ode, u_ode, semi)
+        end
```
efaulhaber, line 368: What if you have multiple fluid systems? Will a particle leaving the inflow zone be spawned in each of the fluid systems?

efaulhaber (changes requested):

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,508 @@
...
+These open boundaries use the characteristic variables to propagate the appropriate values
+to the outlet or inlet and has been proposed by Lastiwka et al (2009). For more information
+about the method see [Open Boundary System](@ref open_boundary).
+
+# Arguments
+- `boundary_zone`: Use [`InFlow`](@ref) for an inflow and [`OutFlow`](@ref) for an outflow boundary.
+- `sound_speed`: Speed of sound.
+
+# Keywords
+- `buffer_size`: Number of buffer particles.
```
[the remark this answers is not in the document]
LasNikas: Yes, it will also work if  `buffer_size=0`. When I have a stationary flow with no turbulence or relative velocities, all particles leaving the zone will be deactivated and activated at same time. I tested this and it works

src/schemes/boundary/open_boundary/system.jl
LasNikas: Good point! See [4aa2a89](https://github.com/trixi-framework/TrixiParticles.jl/pull/442/commits/4aa2a899eb2fa8b9dd1d2f48408363d922457c4a)

efaulhaber: Would it maybe make more sense to permanently link the open boundary system to one fluid system?

LasNikas: I originally did it that way. But it was really cumbersome and didn't entail any advantages. The current approach is more TrixiParticles style, since we only need one open boundary for multiple fluid systems.

efaulhaber: But how do you use one open boundary for multiple fluid systems? Say one is using density diffusion and the other isn't, so you can compare them. So you have two pipe flows, one with and one without DD. How do I connect the no-DD open boundary to the no-DD fluid system and vice versa?

efaulhaber: I want particles coming out of one inflow to go into one fluid system and particles coming out of the other inflow to go into another fluid system.

docs/Project.toml
```diff
@@ -2,8 +2,8 @@
 Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 TrixiBase = "9a0f1c46-06d5-4909-a5a3-ce25d3fa3284"
+TrixiParticles = "66699cd8-9c01-4e9d-a059-b96c86d16b3a"
```
efaulhaber, line 5: This has to be added manually be `]develop`, or we're making the docs of the released package instead.

docs/Project.toml
```diff
@@ -2,8 +2,8 @@
...
 
 [compat]
 Documenter = "1"
 OrdinaryDiffEq = "6"
-TrixiBase = "0.1"
```
efaulhaber, line 9: ?

src/schemes/boundary/open_boundary/system.jl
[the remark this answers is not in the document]
efaulhaber: Okay then. My comment still stands, though.

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,476 @@
...
+function update_open_boundary_eachstep!(system::OpenBoundarySPHSystem, v_ode, u_ode,
+                                        semi, t)
+    u = wrap_u(u_ode, system, semi)
+    v = wrap_v(v_ode, system, semi)
+
+    # Update density, pressure and velocity depending on the characteristic variables.
+    # See eq. 13-15 in Lastiwka (2009) https://doi.org/10.1002/fld.1971
+    @trixi_timeit timer() "update quantities" update_quantities!(system, v, u, t)
+
+    @trixi_timeit timer() "check domain" check_domain!(system, v, u, v_ode, u_ode, semi)
```
efaulhaber, line 184: I think the name "check domain" is not very descriptive. But I also don't have any better ideas...

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,476 @@
...
+
+    # Update buffer
```
efaulhaber, line 186:
```suggestion
    # Update buffers
```

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,476 @@
...
+
+    set_zero!(characteristics)
+    set_zero!(volume)
+
+    # Evaluate the characteristic variables with the fluid system
+    evaluate_characteristics!(system, system.fluid_system, v, u, v_ode, u_ode, semi, t)
+
+    # Only some of the in-/outlet particles are in the influence of the fluid particles.
+    # Thus, we find the characteristics for the particle which are outside the influence
+    # using the average of the values of the previous time step.
```
efaulhaber, line 214:
```suggestion
    # Thus, we compute the characteristics for the particles that are outside the influence
    # of fluid particles by using the average of the values of the previous time step.
```

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,476 @@
...
+    return characteristics
+end
+
+@inline function update_quantities!(system::OpenBoundarySPHSystem, v, u, t)
+    (; density, pressure, characteristics, flow_direction, sound_speed,
+    reference_velocity, reference_pressure, reference_density) = system
+
+    @threaded for particle in each_moving_particle(system)
+        particle_position = current_coords(u, system, particle)
+
```
efaulhaber, line 320:
```suggestion

# Update quantities of boundary particles by evaluating characteristics
```
Something like this?

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,476 @@
...
+    v_fluid = wrap_v(v_ode, fluid_system, semi)
+
+    neighborhood_search = get_neighborhood_search(system, fluid_system, semi)
+
+    for particle in each_moving_particle(system)
+        particle_coords = current_coords(u, system, particle)
+
+        # Check if boundary particle is outside the boundary zone
+        if !is_in_boundary_zone(boundary_zone, particle_coords)
+            transform_particle!(system, fluid_system, boundary_zone, particle,
```
efaulhaber, line 356: How about `convert_particle!`? This also avoids confusion with `transfer_particle!`.

efaulhaber (changes requested):

src/schemes/boundary/open_boundary/system.jl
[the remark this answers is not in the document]
LasNikas: No, it wouldn't work, because activating and deactivating particles is only possible with a `SystemBuffer`

src/schemes/boundary/open_boundary/system.jl
LasNikas: Hm, the characteristics are evaluated prior to `upate_quantities!`.

`# Update quantities as functions of the characteristic variables`
?

src/schemes/boundary/open_boundary/system.jl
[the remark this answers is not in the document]
efaulhaber: But please implement my suggestion above:
```suggestion
- `buffer_size=0`: Number of buffer particles.
```

src/schemes/boundary/open_boundary/system.jl
efaulhaber:
```
# Update quantities based on the characteristic variables
```
?

examples/fluid/pipe_flow_2d.jl
```diff
@@ -0,0 +1,120 @@
...
+
+boundary_system = BoundarySPHSystem(pipe.boundary, boundary_model)
+
+# ==========================================================================================
+# ==== Simulation
+semi = Semidiscretization(fluid_system,
+                          open_boundary_in,
+                          open_boundary_out,
+                          boundary_system,
+                          neighborhood_search=GridNeighborhoodSearch)
```
efaulhaber, line 107:
```suggestion
semi = Semidiscretization(fluid_system, open_boundary_in,
                          open_boundary_out, boundary_system)
```

efaulhaber (changes requested):
Only tests and docs to go

docs/src/systems/boundary.md
```diff
@@ -234,3 +234,86 @@ Pages = [joinpath("schemes", "boundary", "monaghan_kajtar", "monaghan_kajtar.jl"
 - Alireza Valizadeh, Joseph J. Monaghan. "A study of solid wall models for weakly compressible SPH."
   In: Journal of Computational Physics 300 (2015), pages 5–19.
   [doi: 10.1016/J.JCP.2015.07.033](https://doi.org/10.1016/J.JCP.2015.07.033)
+
+# [Open Boundary System](@id open_boundary)
+
+TODO: Make the entire doc nicer and clearer
```
efaulhaber, line 240: If you want to merge it like this, please remove this TODO and add an issue instead. Or do you want to revise it in this PR?

docs/src/systems/boundary.md
```diff
@@ -234,3 +234,86 @@ Pages = [joinpath("schemes", "boundary", "monaghan_kajtar", "monaghan_kajtar.jl"
 - Alireza Valizadeh, Joseph J. Monaghan. "A study of solid wall models for weakly compressible SPH."
   In: Journal of Computational Physics 300 (2015), pages 5–19.
   [doi: 10.1016/J.JCP.2015.07.033](https://doi.org/10.1016/J.JCP.2015.07.033)
+
+# [Open Boundary System](@id open_boundary)
```
efaulhaber, line 238:
```suggestion
# [Open Boundaries](@id open_boundary)
```

docs/src/systems/boundary.md
```diff
@@ -234,3 +234,86 @@ Pages = [joinpath("schemes", "boundary", "monaghan_kajtar", "monaghan_kajtar.jl"
...
+```
+```math
+J_3 = - \rho c_s (v - v_{\text{ref}}) + (p - p_{\text{ref}})
+```
+where the subscript "ref" denotes the reference flow near the boundaries, which can be prescribed.
+
+Specifying the reference variables is **not** equivalent to prescription of ``\rho``, ``v`` and ``p``
+directly, since the perturbation from the reference flow is allowed.
+
+Lastiwka et al (2009) applied the method of characteristic to SPH and determine the number of variables that should be
```
efaulhaber, line 265:
```suggestion
Lastiwka et al. (2009) applied the method of characteristic to SPH and determine the number of variables that should be
```

docs/src/systems/boundary.md
```diff
@@ -234,3 +234,86 @@ Pages = [joinpath("schemes", "boundary", "monaghan_kajtar", "monaghan_kajtar.jl"
...
+- **outflow** boundary:
+    - Prescribe *upstream*-running characteristic ``J_3``
+    - Transmit ``J_1`` and ``J_2`` from the fluid domain.
+
+Prescribing is done by simply setting the characteristics to zero. To transmit the characteristics from the fluid
+domain, or in other words, to carry the information of the fluid to the boundaries, Negi (2020) use a Shepard Interpolation
+```math
+f_i = \frac{\sum_j^N f_j W_{ij}}{\sum_j^N W_{ij}},
+```
+where the ``i``th particle is a boundary particle, ``f`` is either  ``J_1``, ``J_2`` or ``J_3`` and ``N`` is the set of
```
efaulhaber, line 282:
```suggestion
where the ``i``-th particle is a boundary particle, ``f`` is either  ``J_1``, ``J_2`` or ``J_3`` and ``N`` is the set of
```

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,477 @@
...
+
+# Arguments
+- `boundary_zone`: Use [`InFlow`](@ref) for an inflow and [`OutFlow`](@ref) for an outflow boundary.
+
+# Keywords
+- `sound_speed`: Speed of sound.
+- `fluid_system`: The corresponding fluid system
+- `buffer_size=0`: Number of buffer particles.
+- `reference_velocity`: Reference velocity is either a function mapping each particle's coordinates
+						and time to its velocity, an array where the ``i``-th column holds
```
efaulhaber, line 21:
```suggestion
- `reference_velocity`: Reference velocity is either a function mapping each particle's coordinates
			and time to its velocity, an array where the ``i``-th column holds
```
Same below.

docs/src/systems/boundary.md
```diff
@@ -234,3 +234,86 @@ Pages = [joinpath("schemes", "boundary", "monaghan_kajtar", "monaghan_kajtar.jl"
...
+  In: AIAA Journal , Vol. 28, No. 12 pages 2050--2058
+  [doi: 10.2514/3.10521](https://doi.org/10.2514/3.10521)
+- M. Lastiwka, M. Basa, N. J. Quinlan.
+  "Permeable and non-reflecting boundary conditions in SPH".
+  In: International Journal for Numerical Methods in Fluids 61, (2009), pages 709--724.
+  [doi: 10.1002/fld.1971](https://doi.org/10.1002/fld.1971)
+- P. Negi, P. Ramachandran, A. Haftu.
+  "An improved non-reflecting outlet boundary condition for weakly-compressible SPH".
+  In: Computer Methods in Applied Mechanics and Engineering 367, (2020), pages 113119.
+  [doi: 10.1016/j.cma.2020.113119](https://doi.org/10.1016/j.cma.2020.113119)
```
efaulhaber, line 319:
```suggestion
- M. B. Giles. "Nonreflecting boundary conditions for Euler equation calculations".
  In: AIAA Journal 28.12, pages 2050--2058.
  [doi: 10.2514/3.10521](https://doi.org/10.2514/3.10521)
- M. Lastiwka, M. Basa, N. J. Quinlan.
  "Permeable and non-reflecting boundary conditions in SPH".
  In: International Journal for Numerical Methods in Fluids 61 (2009), pages 709--724.
  [doi: 10.1002/fld.1971](https://doi.org/10.1002/fld.1971)
- P. Negi, P. Ramachandran, A. Haftu.
  "An improved non-reflecting outlet boundary condition for weakly-compressible SPH".
  In: Computer Methods in Applied Mechanics and Engineering 367 (2020), pages 113--119.
  [doi: 10.1016/j.cma.2020.113119](https://doi.org/10.1016/j.cma.2020.113119)
```

efaulhaber (changes requested):
Now only tests remaining.

examples/fluid/pipe_flow_2d.jl
```diff
@@ -0,0 +1,120 @@
...
+reynolds_number = 100
+const prescribed_velocity = 2.0
+
+particle_spacing = domain_length_factor * domain_size[1]
+
+boundary_size = (domain_size[1] + 2 * particle_spacing * open_boundary_layers,
+                 domain_size[2])
+
+fluid_density = 1000.0
+pressure = 1000.0
```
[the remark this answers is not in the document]
LasNikas: Do you mean the pressure?
For this particular example, it is necessary to have a background pressure. Otherwise the suction at the outflow is to big and the simulation becomes unstable. With edac and TVF we can use higher pressure so that we don't have this "suction phenomena"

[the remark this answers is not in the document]
efaulhaber: Then please add a comment.

test/systems/open_boundary_system.jl
```diff
@@ -0,0 +1,90 @@
+@testset verbose=true "OpenBoundarySPHSystem" begin
```
efaulhaber, line 1:
```suggestion
@testset verbose=true "`OpenBoundarySPHSystem`" begin
```

test/systems/open_boundary_system.jl
```diff
@@ -0,0 +1,90 @@
...
+                    "each particle's coordinates and time to its density, " *
+                    "a vector holding the density of each particle, or a scalar"
+
+        reference_density = [1.0, 1.0]
+        @test_throws ArgumentError(error_str) OpenBoundarySPHSystem(inflow; sound_speed=1.0,
+                                                                    buffer_size=0,
+                                                                    fluid_system=FluidSystemMock2(),
+                                                                    reference_density)
+    end
+    @testset "show" begin
```
efaulhaber, line 43:
```suggestion
    @testset "`show`" begin
```

test/general/buffer.jl
```diff
@@ -0,0 +1,67 @@
+@testset verbose=true "SystemBuffer" begin
```
efaulhaber, line 1:
```suggestion
@testset verbose=true "`SystemBuffer`" begin
```

test/general/buffer.jl
```diff
@@ -0,0 +1,67 @@
...
+                                          fluid_system=FluidSystemMock3())
+
+    n_particles = nparticles(system)
+
+    @testset "Iterators" begin
+        @test Base.OneTo(n_particles) == TrixiParticles.each_moving_particle(system)
+
+        @test Base.OneTo(n_particles) == TrixiParticles.each_moving_particle(system_buffer)
+
+        particle_ID = TrixiParticles.activate_next_particle(system_buffer)
```
efaulhaber, line 19:
```suggestion
        particle_id = TrixiParticles.activate_next_particle(system_buffer)
```

test/general/buffer.jl
```diff
@@ -0,0 +1,67 @@
...
+                    open_boundary_layers=2, density=1.0, flow_direction=[1.0, 0.0])
+    system = OpenBoundarySPHSystem(inflow; sound_speed=1.0, fluid_system=FluidSystemMock3(),
+                                   buffer_size=0)
+    system_buffer = OpenBoundarySPHSystem(inflow; sound_speed=1.0, buffer_size=5,
+                                          fluid_system=FluidSystemMock3())
+
+    n_particles = nparticles(system)
+
+    @testset "Iterators" begin
+        @test Base.OneTo(n_particles) == TrixiParticles.each_moving_particle(system)
```
efaulhaber, line 15:
```suggestion
        @test TrixiParticles.each_moving_particle(system) == 1:n_particles
```
You can just use `1:n` here and below.

test/general/buffer.jl
```diff
@@ -0,0 +1,67 @@
...
+        @test masses == ic_with_buffer.mass
+
+        densities = initial_condition.density[1] .* ones(nparticles(ic_with_buffer))
+        @test densities == ic_with_buffer.density
+
+        pressures = initial_condition.pressure[1] .* ones(nparticles(ic_with_buffer))
+        @test pressures == ic_with_buffer.pressure
+
+        @testset "Illegal Input" begin
+            ic = rectangular_patch(0.1, (3, 3))
```
efaulhaber, line 61:
```suggestion
            # The rectangular patch has a perturbed, non-constant density
            ic = rectangular_patch(0.1, (3, 3))
```

efaulhaber (changes requested):
Boundary zone and characteristic variables tests remain.

Very nice and thorough tests btw!

svchb:
Update NEWS.md and README.md. Also add pictures to the docs for the examples or add it to the list here #495.

docs/src/systems/boundary.md
```diff
@@ -234,3 +234,87 @@ Pages = [joinpath("schemes", "boundary", "monaghan_kajtar", "monaghan_kajtar.jl"
...
+```math
+ \rho - \rho_{\text{ref}} = \frac{1}{c_s^2} \left( -J_1 + \frac{1}{2} J_2 + \frac{1}{2} J_3 \right),
+```
+```math
+u - u_{\text{ref}}= \frac{1}{2\rho c_s} \left( J_2 - J_3 \right),
+```
+```math
+p - p_{\text{ref}} = \frac{1}{2} \left( J_2 + J_3 \right).
+```
+Thus, determined ``J_1``, ``J_2`` and ``J_3``, we can easily solve for the actual variables for each particle.
```
svchb, line 307:
```suggestion
with ``J_1``, ``J_2`` and ``J_3`` determined we can easily solve for the actual variables for each particle.
```

docs/src/systems/boundary.md
```diff
@@ -234,3 +234,87 @@ Pages = [joinpath("schemes", "boundary", "monaghan_kajtar", "monaghan_kajtar.jl"
...
+- **outflow** boundary:
+    - Prescribe *upstream*-running characteristic ``J_3``
+    - Transmit ``J_1`` and ``J_2`` from the fluid domain.
+
+Prescribing is done by simply setting the characteristics to zero. To transmit the characteristics from the fluid
+domain, or in other words, to carry the information of the fluid to the boundaries, Negi (2020) use a Shepard Interpolation
+```math
+f_i = \frac{\sum_j^N f_j W_{ij}}{\sum_j^N W_{ij}},
+```
+where the ``i``-th particle is a boundary particle, ``f`` is either  ``J_1``, ``J_2`` or ``J_3`` and ``N`` is the set of
```
svchb, line 293:
```suggestion
where the ``i``-th boundary particle, ``f`` is either  ``J_1``, ``J_2`` or ``J_3`` and ``N`` is the set of
```

docs/src/systems/boundary.md
```diff
@@ -234,3 +234,87 @@ Pages = [joinpath("schemes", "boundary", "monaghan_kajtar", "monaghan_kajtar.jl"
...
+- **inflow** boundary:
+    - Prescribe *downstream*-running characteristics ``J_1`` and ``J_2``
+    - Transmit ``J_3`` from the fluid domain (allow ``J_3`` to propagate upstream to the boundary).
+
+- **outflow** boundary:
+    - Prescribe *upstream*-running characteristic ``J_3``
+    - Transmit ``J_1`` and ``J_2`` from the fluid domain.
+
+Prescribing is done by simply setting the characteristics to zero. To transmit the characteristics from the fluid
+domain, or in other words, to carry the information of the fluid to the boundaries, Negi (2020) use a Shepard Interpolation
```
svchb, line 289:
```suggestion
domain, or in other words, to carry the information of the fluid to the boundaries, Negi (2020) uses a Shepard Interpolation
```

docs/src/systems/boundary.md
```diff
@@ -234,3 +234,87 @@ Pages = [joinpath("schemes", "boundary", "monaghan_kajtar", "monaghan_kajtar.jl"
...
+```
+```math
+J_3 = - \rho c_s (v - v_{\text{ref}}) + (p - p_{\text{ref}})
+```
+where the subscript "ref" denotes the reference flow near the boundaries, which can be prescribed.
+
+Specifying the reference variables is **not** equivalent to prescription of ``\rho``, ``v`` and ``p``
+directly, since the perturbation from the reference flow is allowed.
+
+Lastiwka et al. (2009) applied the method of characteristic to SPH and determine the number of variables that should be
```
svchb, line 276:
```suggestion
Lastiwka et al. (2009) applied the method of characteristic to SPH and determined the number of variables that should be
```

docs/src/systems/boundary.md
```diff
@@ -234,3 +234,87 @@ Pages = [joinpath("schemes", "boundary", "monaghan_kajtar", "monaghan_kajtar.jl"
...
+**prescribed** at the boundary and the number which should be **propagated** from the fluid domain to the boundary:
+
+Flow enters the domain through an
+- **inflow** boundary:
```
svchb, line 280:
```suggestion
For an **inflow** boundary:
```

docs/src/systems/boundary.md
```diff
@@ -234,3 +234,87 @@ Pages = [joinpath("schemes", "boundary", "monaghan_kajtar", "monaghan_kajtar.jl"
...
+    - Prescribe *downstream*-running characteristics ``J_1`` and ``J_2``
+    - Transmit ``J_3`` from the fluid domain (allow ``J_3`` to propagate upstream to the boundary).
+
+- **outflow** boundary:
```
svchb, line 284:
```suggestion
For an **outflow** boundary:
```

examples/fluid/pipe_flow_2d.jl
```diff
@@ -0,0 +1,122 @@
...
+# ==========================================================================================
+# ==== Resolution
+particle_spacing = 0.05
+
+# Make sure that the kernel support of fluid particles at a boundary is always fully sampled
+boundary_layers = 3
+
+# Make sure that the kernel support of fluid particles at an open boundary is always
+# fully sampled.
+# Due to the dynamics at the inlets and outlets, it is recommended to use
```
svchb, line 15:
```suggestion
# Note: Due to the dynamics at the inlets and outlets of open boundaries, it is recommended to use
```

examples/fluid/pipe_flow_2d.jl
```diff
@@ -0,0 +1,122 @@
...
+
+# ==========================================================================================
+# ==== Resolution
+particle_spacing = 0.05
+
+# Make sure that the kernel support of fluid particles at a boundary is always fully sampled
+boundary_layers = 3
+
+# Make sure that the kernel support of fluid particles at an open boundary is always
+# fully sampled.
```
svchb, line 14:
```suggestion
```

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -0,0 +1,336 @@
...
+Inflow boundary zone for [`OpenBoundarySPHSystem`](@ref).
+
+The specified plane (line in 2D or rectangle in 3D) will be extruded in upstream
+direction (the direction opposite to `flow_direction`) to create a box for the boundary zone.
+There are three ways to specify the actual shape of the inflow:
+1. Don't pass `initial_condition` or `extrude_geometry`. The boundary zone box will then
+   be filled with inflow particles.
+2. Pass a 1D shape in 2D or a 2D shape in 3D, which is then extruded in upstream direction
+   to create the inflow particles.
+   In 2D, the shape must be either an initial condition with 2D coordinates, which lies
```
svchb, line 15:
```suggestion
  - In 2D, the shape must be either an initial condition with 2D coordinates, which lies
```

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -0,0 +1,336 @@
...
+   on the line specified by `plane`, or an initial condition with 1D coordinates, which lies
+   on the line specified by `plane` when a y-coordinate of `0` is added.
+   In 3D, the shape must be either an initial condition with 3D coordinates, which lies
```
svchb, line 18:
```suggestion
   - In 3D, the shape must be either an initial condition with 3D coordinates, which lies
```

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -0,0 +1,336 @@
...
+   in the rectangle specified by `plane`, or an initial condition with 2D coordinates,
+   which lies in the rectangle specified by `plane` when a z-coordinate of `0` is added.
+   Note that particles outside the boundary zone box will be removed.
```
svchb, line 21:
```suggestion
   **Note** that particles outside the boundary zone box will be removed.
```

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -0,0 +1,336 @@
...
+3. Pass a 2D initial condition in 2D or a 3D initial condition in 3D, which will be used
```
svchb, line 22:
```suggestion
3. Pass an initial condition, which will be used
```

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -0,0 +1,336 @@
...
+Outflow boundary zone for [`OpenBoundarySPHSystem`](@ref).
+
+The specified plane (line in 2D or rectangle in 3D) will be extruded in downstream
+direction (the direction in `flow_direction`) to create a box for the boundary zone.
+There are three ways to specify the actual shape of the outflow:
+1. Don't pass `initial_condition` or `extrude_geometry`. The boundary zone box will then
+   be filled with outflow particles.
+2. Pass a 1D shape in 2D or a 2D shape in 3D, which is then extruded in downstream direction
+   to create the outflow particles.
+   In 2D, the shape must be either an initial condition with 2D coordinates, which lies
```
svchb, line 149:
```suggestion
   - In 2D, the shape must be either an initial condition with 2D coordinates, which lies
```

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -0,0 +1,336 @@
...
+   on the line specified by `plane`, or an initial condition with 1D coordinates, which lies
+   on the line specified by `plane` when a y-coordinate of `0` is added.
+   In 3D, the shape must be either an initial condition with 3D coordinates, which lies
```
svchb, line 152:
```suggestion
   - In 3D, the shape must be either an initial condition with 3D coordinates, which lies
```

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -0,0 +1,336 @@
...
+   in the rectangle specified by `plane`, or an initial condition with 2D coordinates,
+   which lies in the rectangle specified by `plane` when a z-coordinate of `0` is added.
+   Note that particles outside the boundary zone box will be removed.
```
svchb, line 155:
```suggestion
   **Note** that particles outside the boundary zone box will be removed.
```

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -0,0 +1,336 @@
...
+3. Pass a 2D initial condition in 2D or a 3D initial condition in 3D, which will be used
```
svchb, line 156:
```suggestion
3. Pass an initial condition, which will be used
```

svchb (changes requested):

src/schemes/boundary/open_boundary/boundary_zones.jl
LasNikas: Here, the information if it's 2D or 3D is important, IMO (see second point).

svchb: So 2. is passing both initial_condition and extrude_geometry?

LasNikas: 1. `initial_condition=nothing` and `extrude_geometry=nothing` (default)
2. specify  `extrude_geometry` (ignore kwarg `initial_condition`)
3. specify `initial_condition` (ignore kwarg `extrude_geometry`)

efaulhaber (changes requested):
`test/boundary_zone.jl` remaining.

test/schemes/boundary/open_boundary/boundary_zone.jl
```diff
@@ -0,0 +1,221 @@
...
+                outflow = OutFlow(; plane=(point_1, point_2), particle_spacing,
+                                  flow_direction=flow_directions[j], density=1.0,
+                                  open_boundary_layers)
+
+                boundary_zones = [
+                    inflow,
+                    outflow,
+                ]
+
+                @testset "$boundary_zone" for boundary_zone in boundary_zones
```
efaulhaber, line 33: My REPL just keeps printing numbers forever when including this file 💀 
```suggestion
                @testset "$(nameof(typeof(boundary_zone)))" for boundary_zone in boundary_zones
```

test/schemes/boundary/open_boundary/boundary_zone.jl
```diff
@@ -0,0 +1,221 @@
...
+                    zone_width = open_boundary_layers *
+                                 boundary_zone.initial_condition.particle_spacing
+                    sign_ = (boundary_zone isa InFlow) ? -1 : 1
+
+                    @test plane_points_1[i] == boundary_zone.zone_origin
+                    @test plane_points_2[i] - boundary_zone.zone_origin ==
+                          boundary_zone.spanning_set[2]
+                    @test sign_ * flow_directions[j] ≈
+                          normalize(boundary_zone.spanning_set[1])
+                    @test zone_width ≈ norm(boundary_zone.spanning_set[1])
```
efaulhaber, line 43:
```suggestion
                    @test isapprox(zone_width, norm(boundary_zone.spanning_set[1]), atol=1e-14)
```
Same above and for the other errors. Better set an explicit tolerance.

test/schemes/boundary/open_boundary/boundary_zone.jl
```diff
@@ -0,0 +1,221 @@
+@testset verbose=true "Boundary Zone" begin
+    @testset verbose=true "Boundary Zone 2D" begin
+        particle_spacing = 0.2
+        open_boundary_layers = 4
+
+        plane_points_1 = [[0.0, 0.0], [0.5, -0.5], [1.0, 0.5]]
+        plane_points_2 = [[0.0, 1.0], [0.2, 2.0], [2.3, 0.5]]
+
+        @testset "Points $(i)" for i in eachindex(plane_points_1)
```
efaulhaber, line 9:
```suggestion
        @testset verbose=true "Points $(i)" for i in eachindex(plane_points_1)
```
In our current configuration, we set everything to verbose. But we might want to discuss that in the future.

test/schemes/boundary/open_boundary/boundary_zone.jl
```diff
@@ -0,0 +1,221 @@
...
+                outflow = OutFlow(; plane=(point_1, point_2, point_3), particle_spacing,
+                                  flow_direction=flow_directions[j], density=1.0,
+                                  open_boundary_layers)
+
+                boundary_zones = [
+                    inflow,
+                    outflow,
+                ]
+
+                @testset "$boundary_zone" for boundary_zone in boundary_zones
```
efaulhaber, line 92: Same here

src/schemes/boundary/open_boundary/boundary_zones.jl
svchb: Ok than start the subpoints with exactly that information. Especially the text under 2 reads like you need to have both.

LasNikas: good point. thx

test/schemes/boundary/open_boundary/boundary_zone.jl
```diff
@@ -0,0 +1,221 @@
...
+                "On Point 3" => (point_3, true),
+                "Closely On Point 1" => (plane_points[1] .+ perturb_ * flow_direction,
+                                         false))
+
+            @testset "$k" for k in keys(query_points)
+                (particle_position, evaluation) = query_points[k]
+
+                @test evaluation ==
+                      TrixiParticles.is_in_boundary_zone(boundary_zone,
+                                                         particle_position .± eps())
```
efaulhaber, line 145: Why? How?

efaulhaber (changes requested):

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -0,0 +1,477 @@
...
+    characteristics          :: ARRAY2D # Array{ELTYPE, 2}: [characteristic, particle]
+    previous_characteristics :: ARRAY2D # Array{ELTYPE, 2}: [characteristic, particle]
+    sound_speed              :: ELTYPE
+    boundary_zone            :: BZ
+    flow_direction           :: SVector{NDIMS, ELTYPE}
+    reference_velocity       :: RV
+    reference_pressure       :: RP
+    reference_density        :: RD
+    buffer                   :: B
+    update_callback_used     :: Ref{Bool}
```
efaulhaber, line 53: This has to be reset to `false` at some point. Otherwise, one can run one proper simulation with a system and then reuse this system for simulations without `UpdateCallback`.
We might want to reset it in `semidiscretize`, and hope that no users solve the same `ODEProblem` again without an `UpdateCallback`.
To make this even safer, we could maybe also reset it in the initialization of the `InfoCallback`, arguing that this callback is almost always used.

LasNikas: can't I just reset to `false` in `restart_with!`?

efaulhaber: No. You can create a system and then run a simulation. Then you can take the same system and run another simulation from the REPL.

NEWS.md
```diff
@@ -14,11 +14,15 @@ We aim at 3 to 4 month between major release versions and about 2 weeks between
 
 ### Deprecated
 
+## Version 0.1.3
+
+### Added
+Open boundaries using the method of characteristics based on the work of Lastiwka et al., "Permeable and non-reflecting boundary conditions in SPH" (2009) was added for WCSPH and EDAC.
```
efaulhaber, line 20:
```suggestion
Open boundaries using the method of characteristics based on the work of Lastiwka et al., "Permeable and non-reflecting boundary conditions in SPH" (2009) were added for WCSPH and EDAC.

```

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -146,6 +146,12 @@ function Base.show(io::IO, ::MIME"text/plain", system::OpenBoundarySPHSystem)
     end
 end
 
+function initial_callback_flag!(system::OpenBoundarySPHSystem)
```
efaulhaber, line 149:
```suggestion
function reset_callback_flag!(system::OpenBoundarySPHSystem)
```

test/schemes/boundary/open_boundary/boundary_zone.jl
```diff
@@ -63,7 +64,7 @@
             [0.3113730847835541, 0.25057315826416016, -0.02374829351902008],
         ]
 
-        @testset "Points $(i)" for i in eachindex(plane_points_1)
+        @testset verbose=true "Points $(i)" for i in eachindex(plane_points_1)
```
efaulhaber, line 67:
```suggestion
        @testset verbose=true "Points $i" for i in eachindex(plane_points_1)
```

test/schemes/boundary/open_boundary/boundary_zone.jl
```diff
@@ -76,7 +77,7 @@
                 -normalize(cross(edge1, edge2)),
             ]
 
-            @testset "Flow Direction $(j)" for j in eachindex(flow_directions)
+            @testset verbose=true "Flow Direction $(j)" for j in eachindex(flow_directions)
```
efaulhaber, line 80:
```suggestion
            @testset verbose=true "Flow Direction $j" for j in eachindex(flow_directions)
```

efaulhaber (changes requested):

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -146,7 +146,7 @@ function Base.show(io::IO, ::MIME"text/plain", system::OpenBoundarySPHSystem)
     end
 end
 
-function initial_callback_flag!(system::OpenBoundarySPHSystem)
+function reset_callback_flag(system::OpenBoundarySPHSystem)
```
efaulhaber, line 149:
```suggestion
function reset_callback_flag!(system::OpenBoundarySPHSystem)
```

efaulhaber (changes requested):

efaulhaber (approved):

PR #444, opened 2024-03-06
Add Windows and macOS CI tests and only run coverage on Ubuntu with latest Julia version

efaulhaber:
This produces the following runs:
1. Ubuntu latest, Julia latest. This is our "main run" and it's running unit tests with coverage and uploading to Codecov.
2. Ubuntu latest, Julia 1.9.
3. Windows latest, Julia latest.
4. macOS-ARM, Julia latest.

Closes #435.


sloede:
FWIW, I'd additionally run
* macOS (x64), Julia latest
* Ubuntu latest, Julia v1.8

though with a reduced test set. If you are only able to run the full testset, then it's probably too much though

efaulhaber:
We don't have a reduced test set (yet).

LasNikas (approved):

svchb:
1.8 doesn't really work since there is this timestepping bug in OrdinaryDiffEq.jl that was fixed in a version, which has julia compat 1.9.

efaulhaber:
> 1.8 doesn't really work since there is this timestepping bug in OrdinaryDiffEq.jl that was fixed in a version, which has julia compat 1.9.

You're right. It doesn't make sense to support Julia 1.8. This bug made a lot of simulations abort with "instability detected" when using saving and postprocess callback.
IIRC, this was https://github.com/SciML/OrdinaryDiffEq.jl/issues/1616, fixed by https://github.com/SciML/OrdinaryDiffEq.jl/pull/2098.

PR #445, opened 2024-03-06
Warn user when choosing wrong wall viscosity model

LasNikas:
Allow `ArtificialViscosityMonaghan` as wall viscosity model but print out warning. 
(Otherwise it prints weird cache errors when accidentally choosing the wrong viscosity model

As we can see, it would work...
image


src/schemes/boundary/dummy_particles/dummy_particles.jl
```diff
@@ -151,6 +151,16 @@ function create_cache_model(viscosity::Nothing, n_particles, n_dims)
     return (;)
 end
 
+function create_cache_model(viscosity::ArtificialViscosityMonaghan, n_particles, n_dims)
+    ELTYPE = eltype(viscosity.alpha)
+
+    wall_velocity = zeros(ELTYPE, n_dims, n_particles)
+
+    @warn "`ArtificialViscosityMonaghan` for `BoundaryModelDummyParticles` has not been verified yet."
+
```
efaulhaber, line 160: Do we want to have a warning here? I'd say it's enough to have a warning in the docs.

src/schemes/boundary/system.jl
```diff
@@ -198,11 +198,11 @@ end
     return viscous_velocity(v, system.boundary_model.viscosity, system, particle)
 end
 
-@inline function viscous_velocity(v, ::ViscosityAdami, system, particle)
+@inline function viscous_velocity(v, viscosity, system, particle)
     return extract_svector(system.boundary_model.cache.wall_velocity, system, particle)
 end
 
-@inline function viscous_velocity(v, viscosity, system, particle)
+@inline function viscous_velocity(v, ::Nothing, system, particle)
```
svchb, line 205:
```suggestion
@inline function viscous_velocity(v, viscosity::Nothing, system, particle)
```

svchb (changes requested):

svchb (approved):

PR #449, opened 2024-03-08
Add plots and package installation instruction to docs

efaulhaber:
- Installation instructions via Pkg.
- Plots for default example.
- I made sure that the information on the readme is the same as in the installation and gettings started pages.


LasNikas (approved):
LGTM! Thanks a lot!

PR #453, opened 2024-03-10
Heisenbug

efaulhaber:
This is a reproducibility branch for the heisenbug that we've been investigating in the last days.

1. Run `include("validation/dam_break_2d/validation_dam_break_2d.jl")`. Ignore the error.
2. Copy `validation_result_dam_break_wcsph_0015.json` from the out folder to `validation/dam_break_2d`. This is the reference that we're comparing against now.
3. Repeating 1. should now run without errors and yield an error of 0.0. Running that on a single thread or multiple threads should be equivalent because both parallel NHS update and threading in `interact!` is disabled.
4. Changing `parallel=false` to `parallel=true` in the fluid `interact!` (weakly compressible `rhs.jl`, line 24) and running it on multiple threads yields a non-zero error for me.
5. When I comment out the two lines in `semidiscretization.jl` containing `PtrArray` and use the commented out `unsafe_wrap` lines there instead, I get the same results with and without threading (error zero, so the same as with `PtrArray` without threading).


efaulhaber:
Update: @svchb can't reproduce, and I can't reproduce either on an AMD CPU. I conclude that this is an macOS-ARM problem.

svchb:
@efaulhaber Move this to an issue than we can close this.

Issue #424, opened 2024-03-05
Improve custom quantities API


efaulhaber:
Currently, custom quantities for `trixi2vtk`, `SolutionSavingCallback` and `PostprocessCallback` must be functions of `(v, u, t, system)`.
The problem with that is that one needs to use internal, undocumented functions like `each_moving_particle`, `particle_pressure`, `current_coords` and so on to properly work with `v` and `u`.

I therefore changed the docs in #423 to mention that functions must have the arguments `(v, u, t, system)`, but I also mentioned that these required undocumented internal functions to work with. I also removed the examples defining such functions and instead used the pre-defined custom quantities like `kinetic_energy`.

I see the following options:
1. Leave the API as is, add more pre-defined quantities, and then add a docs page explaining internals like `each_moving_particle`, `particle_pressure`, `current_coords` and so on. We should probably have such a page anyway, as it's very useful for users who want to develop TrixiParticles.
2. Allow custom quantity functions to either have the arguments `(v, u, t, system)` (which is then non-public API) or `(coordinates, velocity, density, pressure)`, similar to how the source terms work, but passing global arrays instead of per-particle quantities. Note, however, that we don't want to allocate these global arrays, so we'll have to pass views, which can be modified by the user, causing unpredictable behavior. This option definitely requires less knowledge about TrixiParticle internals to be able to write simple custom quantity functions.

efaulhaber:
To be able to use our callbacks for more complicated things that require data from multiple systems, like interpolation, we also need an option to pass a function with the arguments `v, u, t, v_ode, u_ode, semi`. I think there is no doubt that this should be internal and not public API, since the whole wrapping stuff is also not public API.

So,
1. Leave the public API as is, add internal API that uses arguments `v, u, t, system, v_ode, u_ode, semi` (same as public API, but with `v_ode, u_ode, semi`), add more pre-defined quantities, and then add a docs page explaining internals like `each_moving_particle`, `particle_pressure`, `current_coords` and so on. We should probably have such a page anyway, as it's very useful for users who want to develop TrixiParticles. 
[...]

PR #462, opened 2024-03-13
Example on how to save interpolated planes using postprocessing callback

svchb:

PR #466, opened 2024-03-15
Add wrapper for internal use of interpolation functions

LasNikas:
We should then resume discussion in #424 and if we want to have `v_ode` and `u_ode` in `custom_quantities`.
Also, it might make sense to have `dv_ode` with `get_du` in `custom_quantities`


svchb:
Conflict with #462

LasNikas:
Ah didn't notice. But why should we create a named tuple which is then unpacked again in the method?

svchb:
Yes I wanted to mention that this conflicts. I also thought about this but I wanted to keep the changes minimal in #462.

svchb (approved):

src/general/interpolation.jl
```diff
@@ -51,14 +51,33 @@ results = interpolate_plane_2d([0.0, 0.0], [1.0, 1.0], 0.2, semi, ref_system, so
...
 """
-function interpolate_plane_2d(min_corner, max_corner, resolution, semi, ref_system, sol;
+function interpolate_plane_2d(min_corner, max_corner, resolution, semi, ref_system,
+                              sol::ODESolution;
                               smoothing_length=ref_system.smoothing_length,
                               cut_off_bnd=true, clip_negative_pressure=false)
     # Filter out particles without neighbors
     filter_no_neighbors = true
+    v_ode = sol.u[end].x[1]
+    u_ode = sol.u[end].x[2]
```
efaulhaber, line 61:
```suggestion
    v_ode, u_ode = sol.u[end].x
```

efaulhaber (approved):

src/general/interpolation.jl
efaulhaber, line 61:
```suggestion
    v_ode, u_ode = sol.u[end].x
```

efaulhaber (changes requested):

svchb (approved):

svchb (approved):

PR #472, opened 2024-03-19
TrixiParticles on GPUs with CUDA.jl

efaulhaber:
I spent the last two days building a first prototype for TrixiParticles.jl with CUDA.jl.

The good news:
- It works! I can run the oscillating drop simulation successfully on an RTX 3090.

The bad news:
- It doesn't work with boundary systems yet.
- It's about 30x slower than on the CPU.


efaulhaber:
Superseded by #474.

PR #474, opened 2024-03-20
[Proof of Concept] TrixiParticles.jl on GPUs with KernelAbstractions.jl

efaulhaber:
Like #472, but now with KernelAbstractions.jl. By using the correct array types, we should thus be able to run the exact same code on AMD, Intel or even Apple GPUs.
This PR required minimal changes from #472. See https://github.com/efaulhaber/TrixiParticles.jl/pull/1 for a diff.

> It only works for problem sizes smaller than the GPU (#particles < #threads of the GPU). This is because I just set the number of threads to the number of particles instead of using blocks.
This is now solved here. As opposed to CUDA.jl, KernelAbstractions.jl is choosing thread number and block size automatically. Maybe CUDA.jl can do the same, idk.

The big advantage now is that we can use problem sizes larger than the GPU, and there it's actually fast!
Here is a comparison between the M2 Pro CPU in my laptop (the Thredripper workstation is currently busy) and an RTX 3090:
gpu1


efaulhaber:
Superseded by PRs in #484.

Issue #484, opened 2024-04-02
Road to TrixiParticles.jk on GPUs


efaulhaber:
- [ ] Make array types in structs generic. Change types like `Array{ELTYPE, 2}` to type parameters (Proof of Concept in #474).
- [ ] Recursively convert (adapt) `Semidiscretization` object to GPU types with Adapt.jl (PoC in #474).
- [ ] Minimally invasive working GPU implementation with KernelAbstractions.jl and CUDA backend without neighborhood search (PoC in #474).
- [ ] As above, but with AMD backend.
- [ ] First memory-hungry GPU-compatible grid neighborhood search, allocating the full grid as one big array (PoC in #483).
- [ ] Memory-efficient GPU-compatible grid neighborhood search.
- [ ] Benchmark, profile and optimize interactions on GPUs.

efaulhaber:
Here are some remaining To-Dos from #723:
- [ ] The `SchoenbergQuarticSplineKernel` does not work on GPUs (https://github.com/JuliaGPU/CUDA.jl/issues/2681 and https://github.com/JuliaGPU/Metal.jl/issues/550).
- [ ] Some weird Heisenbug breaking the 3D dam break on Metal (https://github.com/JuliaGPU/Metal.jl/issues/549). But only on my machine, apparently.
- [ ] The open boundary system is not GPU-compatible (https://github.com/trixi-framework/PointNeighbors.jl/issues/20).
- [ ] Monaghan-Kajtar BC don't work with `Float32` (https://github.com/JuliaGPU/Metal.jl/issues/552).
- [ ] Intel GPUs just report a segfault in the first example (https://buildkite.com/julialang/trixiparticles/builds/121#01956711-2f9f-45c6-a3ce-b591179eac07).

PR #493, opened 2024-04-03
Minimally invasive GPU implementation with KernelAbstractions.jl without neighborhood search

efaulhaber:
See #484.
Based on #532 and #533.

This PR is working only with KernelAbstractions.jl, without even using CUDA.jl in TrixiParticles or in an extension module.
All you have to do to run the `dam_break_2d.jl` example on an NVidia GPU is the following:
1. Change the `semidiscretize` call in the file to `ode = semidiscretize(semi, tspan, data_type=nothing)`. This doesn't do anything, as `data_type=nothing` is the default, but it allows us to overwrite it with `trixi_include`. Note that I didn't add this to the file because this feature is experimental and not (yet) documented.
2. Change the `Semidiscretization` call in the file to `semi = Semidiscretization(fluid_system, boundary_system, neighborhood_search=nothing)`, since neighborhood search is not yet supported.
3. Run
    ```julia
   julia> using TrixiParticles, CUDA

   julia> trixi_include("examples/fluid/dam_break_2d.jl", tspan=(0.0, 0.1), data_type=CuArray, saving_callback=nothing, saving_paper=nothing);
   ```
   - We're passing `data_type=CuArray` to run this on the GPU with CUDA.jl. We can also pass `data_type=nothing` (the default), which runs the usual code with Polyester.jl. And we can pass `data_type=Array` to run the GPU kernels compiled by KernelAbstractions.jl on the CPU. We could also pass `data_type=ROCArray` to run a simulation on an AMD GPU.
   - We need to disable the two saving callbacks, as I haven't made them GPU-compatible yet.


svchb:
Fix doc strings.

efaulhaber:
Superseded by #534.

PR #502, opened 2024-04-22
Replace MSE with MRE in validation

svchb:


validation/validation_util.jl
```diff
@@ -34,6 +48,40 @@ function interpolated_mse(reference_time, reference_values, simulation_time,
...
+    # Remove reference time points outside the simulation time
+    start = searchsortedfirst(reference_time, first(simulation_time))
+    end_ = searchsortedlast(reference_time, last(simulation_time))
+    common_time_range = reference_time[start:end_]
+
+    # Interpolate simulation data at the common time points
+    interpolated_values = [linear_interpolation(simulation_time, simulation_values, t)
+                           for t in common_time_range]
+
+    filtered_values = reference_values[start:end_]
```
efaulhaber, line 75: Can you please extract a function here to avoid duplicate code?
Like
```suggestion
function interpolated_errors(reference_time, reference_values, simulation_time, simulation_values)
    if last(simulation_time) > last(reference_time)
        @warn "simulation time range is larger than reference time range. " *
              "Only checking values within reference time range."
    end

    # Remove reference time points outside the simulation time
    start = searchsortedfirst(reference_time, first(simulation_time))
    end_ = searchsortedlast(reference_time, last(simulation_time))
    common_time_range = reference_time[start:end_]

    # Interpolate simulation data at the common time points
    interpolated_values = [linear_interpolation(simulation_time, simulation_values, t)
                           for t in common_time_range]

    filtered_values = reference_values[start:end_]
    
    return abs.(interpolated_values .- filtered_values)
end
```

efaulhaber (changes requested):

efaulhaber:
Short summary of our discussion:
MRE is causing problems when the reference value is close to zero. So we might end up with quite large MRE, just because we have a single pressure value close to zero in the reference.

The whole reason for the relative error discussion in the first place was that pressure errors are quite large due to pressures being in the range of 1e5.
Instead, we came to the conclusion that it makes sense to just write errors as `10eps() * 1e5`, since we already know that the variable we're comparing is in the range of 1e5. This eliminates any additional problems with relative errors.

PR #510, opened 2024-05-03
Mark Release V0.1.1

svchb:


sloede (approved):

sloede:
Usually in Trixi.jl we just push directly to `main` the version update in `Project.toml`, then do a release, and then push the next `-dev` version. No need to go through PRs for this IMHO, but of course, that's up to you :-)

PR #516, opened 2024-05-04
Move neighborhood search to a new standalone package

efaulhaber:


src/general/semidiscretization.jl
```diff
@@ -605,81 +608,98 @@ end
 
 # NHS updates
 # To prevent hard to spot errors there is not default version
-
```
LasNikas, line 608: can you change the comment above to e.g.
`To avoid hard-to-find bugs, there is no default version`

Also, since there is no default version, I'm wondering if it might make sense to comment above the updates and classify them - something like "normal update", "special update only when neighbor is moving" and so on...

src/TrixiParticles.jl
```diff
@@ -21,16 +20,16 @@ using SciMLBase: CallbackSet, DiscreteCallback, DynamicalODEProblem, u_modified!
 @reexport using StaticArrays: SVector
 using StaticArrays: @SMatrix, SMatrix, setindex
 using StrideArrays: PtrArray, StaticInt
-using ThreadingUtilities
 using TimerOutputs: TimerOutput, TimerOutputs, print_timer, reset_timer!
 using TrixiBase: trixi_include
+@reexport using PointNeighbors: TrivialNeighborhoodSearch, GridNeighborhoodSearch
+using PointNeighbors: PointNeighbors, for_particle_neighbor
```
LasNikas, line 26: Isn't this redundant `PointNeighbors: PointNeighbors`?

sloede: No. This allows you to use `PointNeighbors.xyz` inside TrixiP for some unexported symbol `xyz`.

LasNikas (approved):
LGTM!

svchb (approved):

PR #518, opened 2024-05-06
Change NHS API to update with both systems' coordinates

efaulhaber:
I'd prefer changing the API before moving the NHS to a separate package.
With this PR, I pass both system's coordinates to the NHS, together with the information which of the coordinates are changing and which are static. The NHS can then decide if and what to update.

For example, the grid NHS doesn't need to update for fluid-boundary NHS. A neighborhood search that explicitly computes the neighbor lists will have to update them because the fluid particles are moving.


efaulhaber:
Okay, this is too annoying. I'll move this directly into #516.

PR #526, opened 2024-05-15
Remove `pinv` in correction matrix inversion step

efaulhaber:
The `SMatrix` implementation of `pinv` is just falling back to the regular matrix definition, which is then causing allocations and is 400x slower than `inv` for an `SMatrix`.
We found that the matrix is singular if and only if a particle and all neighbors are collinear (in 2D) or lie on the same plane (in 3D). This only happens when either two (in 2D) or three (in 3D) particles are isolated, or on the edges of a very deformed domain. So only in edge cases where the fluid itself has a singularity.
In these cases, we now just disable the correction and use the uncorrected gradient instead. Also see the proof in the comments.

This PR fixes the failing tests in #434.


svchb (changes requested):
Its fine just update the reference files.

bennibolm (approved):

LasNikas (dismissed):
Nice catch! Also, thanks for the proof. That makes sense

LasNikas (approved):

svchb (approved):

PR #529, opened 2024-05-15
Particle packing

LasNikas:
based on #439 and #436


svchb:
Can you please fix the tests first?

svchb:
@LasNikas Please fix the tests first than I will review this PR.

src/preprocessing/particle_packing/rhs.jl
```diff
@@ -0,0 +1,35 @@
+function interact!(dv, v_particle_system, u_particle_system,
+                   v_neighbor_system, u_neighbor_system, neighborhood_search,
+                   system::ParticlePackingSystem, neighbor_system::ParticlePackingSystem)
+    system_coords = current_coordinates(u_particle_system, system)
+    neighbor_coords = current_coordinates(u_neighbor_system, neighbor_system)
+
+    # Loop over all pairs of particles and neighbors within the kernel cutoff.
+    foreach_point_neighbor(system, neighbor_system, system_coords, neighbor_coords,
+                           neighborhood_search) do particle, neighbor, pos_diff, distance
+        # Only consider particles with a distance > 0.
```
efaulhaber, line 10:
```suggestion
    # Loop over all pairs of particles and neighbors within the kernel cutoff
    foreach_point_neighbor(system, neighbor_system, system_coords, neighbor_coords,
                           neighborhood_search) do particle, neighbor, pos_diff, distance
        # Only consider particles with a distance > 0
```

src/preprocessing/particle_packing/signed_distance.jl
```diff
@@ -0,0 +1,273 @@
+"""
+    SignedDistanceField(geometry, particle_spacing;
+                        point_grid=nothing,
+                        max_signed_distance=4particle_spacing,
+                        use_for_boundary_packing=false)
+
+Generate particles along a surface of a complex shape holding the signed distances and normals
```
efaulhaber, line 7:
```suggestion
Generate particles along a surface of a complex geometry holding the signed distances and normals
```

src/preprocessing/particle_packing/signed_distance.jl
```diff
@@ -0,0 +1,273 @@
+"""
+    SignedDistanceField(geometry, particle_spacing;
+                        point_grid=nothing,
+                        max_signed_distance=4particle_spacing,
```
efaulhaber, line 4:
```suggestion
                        max_signed_distance=4 * particle_spacing,
```

docs/src/preprocessing/preprocessing.md
```diff
@@ -255,3 +255,17 @@ Pages = [joinpath("preprocessing", "point_in_poly", "winding_number_jacobson.jl"
...
+TODO
+
+```@autodocs
+Modules = [TrixiParticles]
+Pages = [joinpath("preprocessing", "particle_packing", "system.jl")]
+```
+
+```@autodocs
+Modules = [TrixiParticles]
+Pages = [joinpath("preprocessing", "particle_packing", "signed_distance.jl")]
```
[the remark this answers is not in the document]
LasNikas: How?

src/TrixiParticles.jl
```diff
@@ -52,6 +52,7 @@ include("general/semidiscretization.jl")
 include("general/gpu.jl")
 include("visualization/write2vtk.jl")
 include("visualization/recipes_plots.jl")
+include("preprocessing/particle_packing/particle_packing.jl")
```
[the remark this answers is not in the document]
LasNikas: On first glance, it depends on the `FluidSystem` and `InitialCondition`

docs/src/preprocessing/preprocessing.md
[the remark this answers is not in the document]
efaulhaber: See e.g. the setups:
````
```@autodocs
Modules = [TrixiParticles]
Pages = map(file -> joinpath("setups", file), readdir(joinpath("..", "src", "setups")))
```
````

File: src/TrixiParticles.jl (hunk 5 above)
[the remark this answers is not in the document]
Comment (efaulhaber) [reply src/TrixiParticles.jl:55]: But nothing depends on the preprocessing, right? Then all preprocessing can be included after `include("schemes/schemes.jl")`, before the special includes.

File: src/preprocessing/geometries/io.jl [hunk 6]
```diff
@@ -82,9 +82,17 @@ end
...
-    vertex_normals = stack([geometry.vertex_normals[edge][1] for edge in eachface(geometry)])
+    vertex_normals = Vector{SVector{2, eltype(geometry)}}()
+    vertices = Vector{SVector{2, eltype(geometry)}}()
+
+    for edge in eachface(geometry)
+        push!(vertex_normals, geometry.vertex_normals[edge][1])
+        push!(vertex_normals, geometry.vertex_normals[edge][2])
+        push!(vertices, geometry.vertices[geometry.edge_vertices_ids[edge][1]])
+        push!(vertices, geometry.vertices[geometry.edge_vertices_ids[edge][2]])
+    end
```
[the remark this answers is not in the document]
LasNikas: This version is more robust. consider a `deleteat!` on `geometry`.

src/preprocessing/particle_packing/nhs_faces.jl
```diff
@@ -0,0 +1,263 @@
...
+    ray_intersection(min_corner, max_corner, v1, ray_direction2) && return true
+
+    ray_direction3 = v3 - v1
+    ray_intersection(min_corner, max_corner, v1, ray_direction3) && return true
+
+    normal = face_normal(triangle, geometry)
+
+    # Check if triangle plane intersects cell (for very large triangles)
+    return triangle_plane_intersection(v1, normal, min_corner, max_corner, cell_size)
+end
```
svchb, line 141:
```suggestion
function cell_intersection(face, geometry, cell, neighborhood_search::FaceNeighborhoodSearch{NDIMS}) where {NDIMS}
    (; cell_size) = neighborhood_search

    vertice_list = face_vertices(face, geometry)

    # Check if one of the vertices is inside cell
    for v in vertice_list
        cell == PointNeighbors.cell_coords(v, neighborhood_search) && return true
    end

    # Check if line segments intersect cell
    min_corner = SVector(cell .* cell_size...)
    max_corner = min_corner + SVector(cell_size...)

    ray_direction1 = vertice_list[2] - vertice_list[1]
    ray_intersection(min_corner, max_corner, v1, ray_direction1) && return true
    
    if NDIMS == 3
      ray_direction2 = vertice_list[2] -vertice_list[3]
      ray_intersection(min_corner, max_corner, v1, ray_direction2) && return true
  
      ray_direction3 = vertice_list[3] - vertice_list[1]
      ray_intersection(min_corner, max_corner, v1, ray_direction3) && return true
      
       # For 3D,  Check if triangle plane intersects cell (for very large triangles)
       normal = face_normal(face, geometry)
       return triangle_plane_intersection(vs[1], normal, min_corner, max_corner, cell_size)
    end
    
    return false
end
```

src/preprocessing/particle_packing/nhs_faces.jl
```diff
@@ -0,0 +1,263 @@
...
+
+    # Check if line segment intersects cell
+    min_corner = SVector(cell .* cell_size...)
+    max_corner = min_corner + SVector(cell_size...)
+
+    ray_direction1 = v2 - v1
+    ray_intersection(min_corner, max_corner, v1, ray_direction1) && return true
+
+    ray_direction2 = v2 - v3
+    ray_intersection(min_corner, max_corner, v1, ray_direction2) && return true
```
svchb, line 132: Also shouldn't this be v3?
ray_intersection(min_corner, max_corner, v3, ray_direction2)

src/preprocessing/particle_packing/nhs_faces.jl
```diff
@@ -0,0 +1,263 @@
...
+    cell_vertex_8 = max_corner - dirz
+    pos_diff_8 = cell_vertex_8 - point
+
+    # Corners: bottom south-east and top north-east
+    dot8 = sign(dot(pos_diff_8, plane_normal))
+    !(dot7 == dot8) && return true
+
+    # All edge vertices are on one side of the plane
+    return false
+end
```
svchb, line 226:
```suggestion
function triangle_plane_intersection(point, plane_normal, min_corner, max_corner, cell_size)
    # Generate all eight corners of the cell (axis-aligned bounding box).
    # Each corner is calculated by adding a combination of cell_size components
    # to the min_corner, based on the indices i, j, k, which take values 0 or 1.
    # This creates all possible combinations, resulting in the eight corners of the cell.
    corners = [
        min_corner + SVector(i * cell_size[1], j * cell_size[2], k * cell_size[3])
        for i in (0, 1), j in (0, 1), k in (0, 1)
    ]

    # For each corner, compute the vector from a known point on the plane to the corner.
    # Then, calculate the dot product of this vector with the plane's normal vector.
    # The dot product indicates the relative position of the corner to the plane:
    # - Positive value: corner is on the same side as the normal vector points.
    # - Zero: corner lies exactly on the plane.
    # - Negative value: corner is on the opposite side from where the normal vector points.
    # The sign() function converts the dot product into -1, 0, or 1 accordingly.
    signs = [sign(dot(corner - point, plane_normal)) for corner in corners]

    # Determine if the plane intersects the cell by checking the signs.
    # - If all signs are the same (all corners on one side of the plane), there's no intersection.
    # - If signs vary (some corners on different sides), the plane intersects the cell.
    # If maximum(signs) > minimum(signs), it means there are both positive and negative signs.
    return maximum(signs) > minimum(signs)
end

```

src/preprocessing/particle_packing/nhs_faces.jl
```diff
@@ -0,0 +1,263 @@
...
+
+    cell1 = PointNeighbors.cell_coords(v1, neighborhood_search)
+    cell2 = PointNeighbors.cell_coords(v2, neighborhood_search)
+    cell3 = PointNeighbors.cell_coords(v3, neighborhood_search)
+
+    mins = min.(cell1, cell2, cell3)
+    maxs = max.(cell1, cell2, cell3)
+
+    return meshgrid(mins, maxs)
+end
```
svchb, line 254:
```suggestion
@inline function cell_grid(face, geometry, neighborhood_search::FaceNeighborhoodSearch{NDIMS}) where {NDIMS}
    vertice_list = face_vertices(face, geometry)
    
    # Compute the cell coordinates for each vertex
    cells = [PointNeighbors.cell_coords(v, neighborhood_search) for v in vertice_list]
    
    # Compute the element-wise minimum and maximum cell coordinates across all vertices
    mins = reduce((a, b) -> min.(a, b), cells)
    maxs = reduce((a, b) -> max.(a, b), cells)
    
    return meshgrid(mins, maxs)
end
```

src/preprocessing/particle_packing/system.jl
```diff
@@ -0,0 +1,277 @@
...
+                 The thickness of the boundary is specified by passing the [`SignedDistanceField`](@ref) of `boundary` with:
+                    - `use_for_boundary_packing=true`
+                    - `max_signed_distance=boundary_thickness`
+- `signed_distance_field`: To constrain particles onto the surface, the information about
+                          the signed distance from a particle to a face is required.
+                          The precalculated signed distances will be interpolated to each particle during the packing procedure.
+- `smoothing_kernel`: Smoothing kernel to be used for this system.
+                      See [Smoothing Kernels](@ref smoothing_kernel).
+- `smoothing_length`: Smoothing length to be used for this system.
+                      See [Smoothing Kernels](@ref smoothing_kernel).
```
svchb, line 33: Please align

src/preprocessing/particle_packing/system.jl
```diff
@@ -0,0 +1,277 @@
...
+    is_boundary           :: Bool
+    shift_condition       :: ELTYPE
+    neighborhood_search   :: N
+    signed_distances      :: Vector{ELTYPE} # Only for visualization
+    buffer                :: Nothing
+    update_callback_used  :: Ref{Bool}
+
+    function ParticlePackingSystem(shape::Union{InitialCondition, ComplexShape};
+                                   smoothing_kernel=SchoenbergCubicSplineKernel{ndims(shape)}(),
+                                   smoothing_length=1.2shape.particle_spacing,
```
svchb, line 53:
```suggestion
                                   smoothing_length=1.2*shape.particle_spacing,
```

svchb (changes requested):

src/preprocessing/particle_packing/nhs_faces.jl
LasNikas: Thanks for this. Your version is definitely more readable. I'm wondering if my version is slightly faster as we don't have to check each combination in best case.

However, I prefer your readable version. What's @efaulhaber opinion on this?

svchb: Please just test what the performance is like compared to the current version.

LasNikas: That's not that easy. I mean it's case dependent. In best case (first condition is true), it should be faster. But in worst case it should be same performance, no?

svchb: Yes but I would just compare it for a large example and check if the impact is actually measurable.

LasNikas: I think the performance difference is negligible, that's why I also prefer your version.

src/preprocessing/particle_packing/system.jl
LasNikas: it is aligned?

svchb: we have usually also aligned the ":"

LasNikas: like so? 
image

src/preprocessing/particle_packing/nhs_faces.jl
efaulhaber:
```julia
function triangle_plane_intersection(point_on_plane, plane_normal, cell_min_corner, cell_size)
    cell_center = cell_min_corner .+ cell_size ./ 2

    # The corner that is furthest in the direction of `plane_normal`.
    # When the plane is not intersecting the cell, this corner is the one closest
    # to the plane when the normal is pointing away from the cell or furthest from the plane
    # when the normal is pointing towards the cell.
    # Note that this could also be a face or edge midpoint when the plane is axis-aligned.
    normal_unit = sign.(plane_normal)
    corner1 = cell_center + normal_unit .* cell_size / 2
    corner2 = cell_center - normal_unit .* cell_size / 2

    # These two vectors are on the same side of the plane
    # if and only if the plane intersects the cell.
    plane_to_corner1 = corner1 - point_on_plane
    plane_to_corner2 = corner2 - point_on_plane

    # Return true if the two vectors are on different sides of the plane
    return dot(plane_normal, plane_to_corner1) * dot(plane_normal, plane_to_corner2) < 0
end
```

Tested for random samples:
```
Default:
  7.340 ns (0 allocations: 0 bytes)
Sven:
  63.733 ns (4 allocations: 416 bytes)
Erik:
  4.791 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  107.200 ns (4 allocations: 416 bytes)
Erik:
  4.833 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  114.166 ns (4 allocations: 416 bytes)
Erik:
  4.791 ns (0 allocations: 0 bytes)

Default:
  7.340 ns (0 allocations: 0 bytes)
Sven:
  75.069 ns (4 allocations: 416 bytes)
Erik:
  4.833 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  111.201 ns (4 allocations: 416 bytes)
Erik:
  4.833 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  72.371 ns (4 allocations: 416 bytes)
Erik:
  4.791 ns (0 allocations: 0 bytes)

Default:
  3.666 ns (0 allocations: 0 bytes)
Sven:
  115.297 ns (4 allocations: 416 bytes)
Erik:
  4.833 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  74.966 ns (4 allocations: 416 bytes)
Erik:
  4.833 ns (0 allocations: 0 bytes)

Default:
  7.375 ns (0 allocations: 0 bytes)
Sven:
  116.703 ns (4 allocations: 416 bytes)
Erik:
  4.833 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  75.111 ns (4 allocations: 416 bytes)
Erik:
  4.791 ns (0 allocations: 0 bytes)
```

I also verified that they are all correct:
```julia
@testset verbose=true "triangle plane intersection" begin
    cell_size = (1.0, 1.0, 1.0)
    min_corners = vec(SVector.(Tuple.(CartesianIndices((10, 10, 10)))))
    plane_points = rand(SVector{3}, 10) .* 10
    plane_normals = [2 * rand(SVector{3}) .- 1 for _ in 1:10]

    for point_on_plane in plane_points
        for plane_normal in plane_normals
            for cell_min_corner in min_corners
                cell_max_corner = cell_min_corner .+ cell_size

                default = TrixiParticles.triangle_plane_intersection(point_on_plane, plane_normal,
                                                                     cell_min_corner, cell_max_corner,
                                                                     cell_size)
                sven = TrixiParticles.triangle_plane_intersection_sven(point_on_plane, plane_normal,
                                                                       cell_min_corner, cell_max_corner,
                                                                       cell_size)
                erik = TrixiParticles.triangle_plane_intersection_erik(point_on_plane, plane_normal,
                                                                       cell_min_corner, cell_max_corner,
                                                                       cell_size)

                @test default == sven == erik
            end
        end
    end
end
```

src/preprocessing/particle_packing/system.jl
svchb: yes

src/preprocessing/particle_packing/nhs_faces.jl
svchb: ok I see I should have avoided creating the lists^^

efaulhaber: The trend still holds when I make all your vectors generators to avoid allocations:
```
Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  26.564 ns (0 allocations: 0 bytes)
Erik:
  4.833 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  26.538 ns (0 allocations: 0 bytes)
Erik:
  4.833 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  26.690 ns (0 allocations: 0 bytes)
Erik:
  4.833 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  26.564 ns (0 allocations: 0 bytes)
Erik:
  4.833 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  26.564 ns (0 allocations: 0 bytes)
Erik:
  4.833 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  26.591 ns (0 allocations: 0 bytes)
Erik:
  4.834 ns (0 allocations: 0 bytes)

Default:
  3.625 ns (0 allocations: 0 bytes)
Sven:
  26.564 ns (0 allocations: 0 bytes)
Erik:
  4.834 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  26.564 ns (0 allocations: 0 bytes)
Erik:
  4.833 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  26.553 ns (0 allocations: 0 bytes)
Erik:
  4.833 ns (0 allocations: 0 bytes)

Default:
  7.333 ns (0 allocations: 0 bytes)
Sven:
  26.580 ns (0 allocations: 0 bytes)
Erik:
  4.875 ns (0 allocations: 0 bytes)
```

svchb: It improved by ~500% though :D

LasNikas: cool! Thanks a lot

svchb:
Please add some tests.

src/preprocessing/geometries/io.jl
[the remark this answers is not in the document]
efaulhaber: So we now have each vertex twice (once at the end of an edge and once at the start of the next edge) with different normals?

src/preprocessing/particle_packing/signed_distance.jl
```diff
@@ -0,0 +1,273 @@
...
+    cond_1 = distances .== Inf
+
+    # Maximum inner distances
+    cond_2 = distances .< -max_signed_distance
+
+    # Keep "larger" signed distance field outside `boundary` to guarantee
+    # compact support for boundary particles.
+    cond_3 = distances .> sdf_factor * max_signed_distance
+
+    delete_indices = (cond_1 .|| cond_2 .|| cond_3)
```
efaulhaber, line 125:
```suggestion
    # Keep "larger" signed distance field outside `boundary` to guarantee
    # compact support for boundary particles.
    keep = -max_signed_distance .< distances .< sdf_factor * max_signed_distance
    delete_indices = .!keep
```

src/preprocessing/particle_packing/signed_distance.jl
```diff
@@ -0,0 +1,273 @@
...
+    (; positions, distances, normals) = signed_distance_field
+    positions = stack(signed_distance_field.positions)
+
+    trixi2vtk(positions, signed_distances=distances, normals=normals,
+              filename="signed_distance_field")
+end
+
+function calculate_signed_distances!(positions, distances, normals,
+                                     boundary, sdf_factor, max_signed_distance, nhs)
+    @threaded positions for point in eachindex(positions)
```
efaulhaber, line 101: How performance-relevant is all of this here? Is it only used once during initialization, or is it a bottleneck somewhere? We could probably speed this up by only looping over neighboring cells of non-empty cells.
But that only makes sense if this is not already fast enough.

src/preprocessing/particle_packing/signed_distance.jl
```diff
@@ -0,0 +1,273 @@
...
+        point_coords = positions[point]
+
+        for face in eachneighbor(point_coords, nhs)
+            # `sdf = (sign, distance, normal)`
+            sdf = signed_point_face_distance(point_coords, boundary, face)
+
+            if sdf[2] < distances[point]^2
+                distances[point] = sdf[1] ? -sqrt(sdf[2]) : sqrt(sdf[2])
```
efaulhaber, line 109:
```suggestion
                # Found a face closer than the previous closest face
                distances[point] = sdf[1] ? -sqrt(sdf[2]) : sqrt(sdf[2])
```

src/setups/complex_shape.jl
```diff
@@ -41,13 +41,25 @@ For more information about the method see [`WindingNumberJacobson`](@ref) or [`W
 !!! warning "Experimental Implementation"
     This is an experimental feature and may change in any future releases.
 """
-function ComplexShape(geometry::Union{TriangleMesh, Polygon}; particle_spacing, density,
+struct ComplexShape{S, IC, ICB, SDF, IG, WN, ELTYPE}
```
efaulhaber, line 44: Why is this a struct now? Especially since the packing system can also accept a plain `InitialCondition`.

efaulhaber (changes requested):
Examples and system remaining.

src/preprocessing/particle_packing/signed_distance.jl
```diff
@@ -0,0 +1,273 @@
...
+
+# Keywords
+- `max_signed_distance`: Maximum signed distance to be stored. That is, only particles with a
+                         distance of `abs(max_signed_distance)` to the surface of the shape
+                         will be sampled.
+- `point_grid`: Point grid in which the signed distance field is computed.
+                When set to `nothing` (default), the bounding box of the shape will be
+                sampled with a cartesian point grid.
+- `use_for_boundary_packing`: Set to `true` if [`SignedDistanceField`] is used to pack
+                              a boundary [`ParticlePackingSystem`](@ref).
```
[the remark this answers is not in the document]
LasNikas: For all other case where one needs a `SignedDistanceField`.  If `use_for_boundary_packing=true` the other signed distance field is larger then the inner one. This is a special case and `false` is the default case. Idk if this is to mention, no?

src/preprocessing/particle_packing/signed_distance.jl
```diff
@@ -0,0 +1,273 @@
...
+                                                    (max_corner .- min_corner) ./
+                                                    particle_spacing))
+
+            grid = rectangular_shape_coords(particle_spacing, n_particles_per_dimension,
+                                            min_corner; tlsph=true)
+
+            point_grid = reinterpret(reshape, SVector{NDIMS, ELTYPE}, grid)
+        end
+
+        positions = [SVector(position) for position in point_grid]
```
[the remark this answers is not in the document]
LasNikas: Yes, but `reinterpret`ed and not `resize`able

src/preprocessing/particle_packing/signed_distance.jl
LasNikas: It is only used once during initialization (so far). However, for big geometries it can be performance relevant. That is, the faster the better. 

> We could probably speed this up by only looping over neighboring cells of non-empty cells.

What exactly do you mean? Check if the point position is in a non-empty cell? But what if the face is in the neighboring cell and closest to this point position?

src/setups/complex_shape.jl
LasNikas: I'm not really happy with it yet either. I'm not really sure about the API. I think we should this discuss first. 
The issue is, that this holds also the boundary IC with the same initial particle grid and the same SDF without generating another SDF.

src/preprocessing/geometries/io.jl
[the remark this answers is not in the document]
LasNikas: Yes. This is to visualize the normals in Paraview

src/preprocessing/particle_packing/signed_distance.jl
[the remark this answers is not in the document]
efaulhaber: But then you can just `copy` it.

src/preprocessing/geometries/io.jl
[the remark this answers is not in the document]
efaulhaber: Then please add a comment like "add each vertex twice (once at the end of an edge and once at the start of the next edge) with corresponding normals to make ParaView work"

src/preprocessing/particle_packing/signed_distance.jl
LasNikas: Is looping over cells thread supported?
How can I loop over non-empty cells? 
How do I know which point index is in the non-empty cell?

src/preprocessing/particle_packing/signed_distance.jl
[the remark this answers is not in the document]
LasNikas: ah I see

src/preprocessing/particle_packing/signed_distance.jl
efaulhaber: Not out of the box. This would require some careful optimization in the NHS itself. If this is not a bottleneck, then it's not worth it.

src/preprocessing/particle_packing/signed_distance.jl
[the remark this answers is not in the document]
efaulhaber: ?

src/preprocessing/particle_packing/system.jl
```diff
@@ -0,0 +1,283 @@
...
+function kinetic_energy(v, u, t, system::ParticlePackingSystem)
+    (; initial_condition, is_boundary) = system
+
+    # Exclude boundary packing system
+    is_boundary && return zero(eltype(system))
+
+    # If `each_moving_particle` is empty (no moving particles), return zero
+    return sum(each_moving_particle(system), init=zero(eltype(system))) do particle
+        velocity = extract_svector(initial_condition.velocity, system, particle)
+        return 0.5 * initial_condition.mass[particle] * dot(velocity, velocity)
```
efaulhaber, line 150:
```suggestion
        return initial_condition.mass[particle] * dot(velocity, velocity) / 2
```

src/preprocessing/particle_packing/system.jl
```diff
@@ -0,0 +1,283 @@
...
+        end
+
+        if volume > eps()
+            distance_signed /= volume
+            normal_vector /= volume
+
+            # Store signed distance for visualization
+            system.signed_distances[particle] = distance_signed
+
+            constrain_particles!(u, system, particle, distance_signed, normal_vector)
```
efaulhaber, line 225:
```suggestion
            constrain_particle!(u, system, particle, distance_signed, normal_vector)
```
This is only constraining `particle`.

src/preprocessing/particle_packing/system.jl
```diff
@@ -0,0 +1,283 @@
...
+            u[dim, particle] -= shift[dim]
+        end
+    end
+
+    return u
+end
+
+@inline function update_transport_velocity!(system::ParticlePackingSystem, v_ode, semi)
+    v = wrap_v(v_ode, system, semi)
+    for particle in each_moving_particle(system)
```
efaulhaber, line 262: Threaded for GPU support.

src/preprocessing/particle_packing/system.jl
```diff
@@ -0,0 +1,283 @@
...
+        for i in 1:ndims(system)
+            # Use the initial condition to store the evolved velocity
+            system.initial_condition.velocity[i, particle] = v[i, particle]
```
efaulhaber, line 265: Why? This is not GPU-compatible. The initial condition lies on the CPU. IMO, it would make more sense to store this in a dedicated vector. Also, the usual argument holds that you cannot call `solve` twice on this system and get the same results. You always have to re-create the system, which is annoying and confusing when working interactively.

src/preprocessing/particle_packing/signed_distance.jl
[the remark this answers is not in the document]
LasNikas: The only reason is memory efficiency.
I don't find this less readable to be honest.
```distances[point] = sign_bit ? -sqrt(distance) : sqrt(distance)```

**Review by efaulhaber — CHANGES_REQUESTED** [review]

File: src/preprocessing/point_in_poly/winding_number_jacobson.jl [hunk 23]
```diff
@@ -86,15 +86,15 @@ function (point_in_poly::WindingNumberJacobson)(geometry, points;
...
-    inpoly = fill(false, size(points, 2))
+    inpoly = fill(false, length(points))
 
     winding_numbers = Float64[]
     store_winding_number && (winding_numbers = resize!(winding_numbers, length(inpoly)))
 
     divisor = ndims(geometry) == 2 ? 2pi : 4pi
 
-    @threaded points for query_point in axes(points, 2)
-        p = point_position(points, geometry, query_point)
```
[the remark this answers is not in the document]
LasNikas: I have removed it in `geometries.jl`

src/preprocessing/particle_packing/system.jl
```diff
@@ -0,0 +1,283 @@
...
+- `tlsph`:                 With the [`TotalLagrangianSPHSystem`](@ref), particles need to be placed
+                           on the boundary of the shape and not one particle radius away,
+                           as for fluids. When `tlsph=true`, particles will be placed
+                           on the boundary of the shape.
+- `is_boundary`:           When `is_boundary=true`, boundary particles will be sampled
+                           and packed in an offset surface of the `boundary`.
+                           The thickness of the boundary is specified by passing the
+                           [`SignedDistanceField`](@ref) of `boundary` with:
+                              - `use_for_boundary_packing=true`
+                              - `max_signed_distance=boundary_thickness`
```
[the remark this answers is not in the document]
LasNikas: If set to `true` the `ParticlePackingSystem` packs only boundary particles of the geometry.
If set to `false` (default) the `ParticlePackingSystem` packs the particles inside the geometry.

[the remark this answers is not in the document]
efaulhaber: So there is always only one system packed? Either inside or boundary? This is not clear to me from the docs.
And I still don't understand this description.
```suggestion
- `pack_boundary`:         When `pack_boundary=false`, the particles in `boundary`
                           are fixed and the particles in `shape` are packed.
                           When `pack_boundary=true`, the particles in `shape` are fixed
                           and the particles in `boundary` are fixed.
                           It often makes sense to first pack the `shape` against the `boundary`
                           and then in the next step `boundary` against the packed `shape`.
                           The thickness of the boundary is specified by passing the
                           [`SignedDistanceField`](@ref) of `boundary` with:
                              - `use_for_boundary_packing=true`
                              - `max_signed_distance=boundary_thickness`
```
Is this correct? Is it not possible to pack both at the same time?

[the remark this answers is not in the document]
LasNikas: No, that's not correct.
I have two different systems. One is filled with the particles inside the geometry (fluid) and one is filled with particles outside the geometry (boundary). These are independent systems that interact with each other.

The only difference between these systems is the shift condition used to constrain particles onto the surface.

For example:
```julia
packing_system = ParticlePackingSystem(shape_sampled; boundary=geometry,
                                       signed_distance_field, tlsph=tlsph,
                                       background_pressure)

boundary_system = ParticlePackingSystem(boundary_sampled; boundary=geometry,
                                        is_boundary=true, signed_distance_field,
                                        tlsph=tlsph, background_pressure)

semi = Semidiscretization(packing_system, boundary_system)
```

svchb (changes requested):
This still requires tests.

src/preprocessing/particle_packing/system.jl
```diff
@@ -0,0 +1,287 @@
...
+        # TODO: Let `Semidiscretization` handle this?
+        # Create neighborhood search
+        nhs_ = isnothing(neighborhood_search) ? TrivialNeighborhoodSearch{NDIMS}() :
+               neighborhood_search
+        nhs = copy_neighborhood_search(nhs_,
+                                       compact_support(smoothing_kernel, smoothing_length),
+                                       length(signed_distance_field.positions))
+
+        # Initialize neighborhood search with signed distances
+        PointNeighbors.initialize_grid!(nhs, stack(signed_distance_field.positions))
```
LasNikas, line 81: @efaulhaber should I use a full grid nhs here? The positions of the `signed_distance_field` are of course static. Or what is the best choice here?

efaulhaber: Yes, by default the nhs should be using a `FullGridCellList`. You need to set this in the kwarg.
This cell list is faster and GPU-compatible. Only downsides are 1. one needs to specify a bounding box and 2. the domain is constrained to this bounding box. Here you know the bounding box and particles should never leave it, right?

LasNikas: > Here you know the bounding box and particles should never leave it, right?

Yes, exactly. I also wondering if I use a precomputed NHS here. What do you think?

LasNikas:
needs https://github.com/trixi-framework/PointNeighbors.jl/pull/17

src/preprocessing/particle_packing/system.jl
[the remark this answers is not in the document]
efaulhaber: I still don't understand the description of the option. Let's work out a better formulation.

efaulhaber (changes requested):
I'll have to look at the NHS and the tests in more detail again, but here's my review so far.

src/preprocessing/particle_packing/system.jl
```diff
@@ -259,10 +257,9 @@ end
 
 @inline function update_transport_velocity!(system::ParticlePackingSystem, v_ode, semi)
     v = wrap_v(v_ode, system, semi)
-    for particle in each_moving_particle(system)
+    @threaded system for particle in each_moving_particle(system)
         for i in 1:ndims(system)
-            # Use the initial condition to store the evolved velocity
-            system.initial_condition.velocity[i, particle] = v[i, particle]
+            v[ndims(system) + i, particle] = v[i, particle]
```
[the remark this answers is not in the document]
LasNikas: No, the advection velocity (`v[ndims(system) + i, particle]`) is not constant during the stages. Thus, it must be integrated as well.

integrated:
image

constant:
image

src/general/initial_condition.jl
```diff
@@ -290,6 +290,36 @@ end
 
 Base.intersect(initial_condition::InitialCondition) = initial_condition
 
+function InitialCondition(sol::ODESolution, system, semi; use_final_velocity=false,
+                          max_particle_distance=system.initial_condition.particle_spacing /
```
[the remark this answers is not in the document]
LasNikas: > 1. Why max_distance? Isn't this the minimum allowed distance?

Yes, you're right.

> 2. Why do you even want to remove these particles? Shouldn't a simulation solution already have nicely distributed particles?

This is intended to be a check if particles come too close especially when the surface exhibits large curvature.

efaulhaber (changes requested):

test/preprocessing/packing/nhs_faces.jl
```diff
@@ -0,0 +1,92 @@
...
+
+    # Tested with Paraview and `save("triangle.stl", triangle)`, `trixi2vtk(stack([min_corner, max_corner]))`
+    @testset verbose=true "3D Cell Intersection" begin
+        # Axis aligned triangle
+        A = SVector(0.0, 1.0, 0.0)
+        B = zero(A)
+        C = SVector(1.0, 1.0, 0.0)
+
+        triangle_aligned = TrixiParticles.TriangleMesh([(A, B, C)],
+                                                       [TrixiParticles.cross(B - A, C - A)],
```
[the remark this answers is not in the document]
LasNikas: because this is the normal of the triangle ABC

src/preprocessing/particle_packing/nhs_faces.jl
```diff
@@ -0,0 +1,179 @@
...
+
+        ray_direction = vertices_list[3] - vertices_list[1]
+        ray_origin = vertices_list[1]
+
+        ray_intersection(min_corner, max_corner, ray_origin, ray_direction) && return true
+
+        # For 3D,  Check if triangle plane intersects cell (for very large triangles)
+        normal = face_normal(face, geometry)
+
+        return triangle_plane_intersection(ray_origin, normal, min_corner, cell_size)
```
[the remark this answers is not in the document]
LasNikas: see https://github.com/trixi-framework/TrixiParticles.jl/pull/699

src/general/initial_condition.jl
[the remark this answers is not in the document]
efaulhaber: > This is intended to be a check if particles come too close especially when the surface exhibits large curvature.

Can you please note this in a comment?

src/preprocessing/particle_packing/system.jl
```diff
@@ -159,8 +159,10 @@ end
     return system.initial_condition.mass[particle]
 end
 
+# Update from `UpdateCallback`
```
efaulhaber, line 162:
```suggestion
# Update from `UpdateCallback` (between time steps)
```
For people who don't know about this callback.

test/preprocessing/packing/nhs_faces.jl
```diff
@@ -0,0 +1,100 @@
...
+                    0.0 0.0 0.7 0.0]
+
+        # Only use the third edge of the triangle, i.e. the edge from [0.1, 0.0] to [0.0, 0.0]
+        edge_aligned = deleteat!(TrixiParticles.Polygon(triangle), [1, 2])
+        edge_id = 1 # Only one edge in `Polygon`
+
+        cell_sizes = [1.0 + sqrt(eps()), 0.1]
+
+        expected_ncells_bbox = [(1, 1), (11, 1)]
+        expected_ncells_filled = map(x -> prod(x .+ 2), expected_ncells_bbox)
```
efaulhaber, line 13:
```suggestion
        # One padding layer in each direction around the bounding box
        expected_ncells_filled = map(x -> prod(x .+ 2), expected_ncells_bbox)
```

efaulhaber (changes requested):

efaulhaber (approved):
LGTM!

docs/src/preprocessing/preprocessing.md
```diff
@@ -255,3 +255,12 @@ Pages = [joinpath("preprocessing", "point_in_poly", "winding_number_jacobson.jl"
...
 ```
+
+# Particle Packing
+
+TODO
+
+```@autodocs
+Modules = [TrixiParticles]
+Pages = map(file -> joinpath("preprocessing", "particle_packing", file), readdir(joinpath("..", "src", "preprocessing", "particle_packing")))
+```
```
svchb, line 266: Is there an issue for this?

src/preprocessing/particle_packing/signed_distance.jl
```diff
@@ -0,0 +1,290 @@
+"""
+    SignedDistanceField(geometry, particle_spacing;
+                        points=nothing,
+                        max_signed_distance=4 * particle_spacing,
+                        use_for_boundary_packing=false)
+
+Generate particles along a surface of a complex geometry holding the signed distances and normals
+to this surface.
```
svchb, line 8:
```suggestion
Generate particles along a surface of a complex geometry storing the signed distances and normals
to this surface.
```

svchb (changes requested):

docs/src/preprocessing/preprocessing.md
LasNikas: #701

efaulhaber (dismissed):

svchb (approved):

NEWS.md
```diff
@@ -3,51 +3,88 @@
...
+
+### Documentation
+
+- GPU Support Documentation was added (#660)
+- A new user tutorial was added (#514)
+
+### Fixes
+
+- Diverse Doc fixes (#663, #659, #637, #658, #664)
+- Simulations can be run with `Float32` (#662)
```
efaulhaber, line 30:
```suggestion
- Add particle packing based on STLs (#529)

### Compatibility Changes
- Dropped support for Julia 1.9

## Version 0.2.4

### Features

- A method to prevent penetration of fast moving particles with solids was added (#498)
- Added the callback `SteadyStateReachedCallback` to detect convergence of static simulations (#601)  
- Added ideal gas state equation (#607)
- Simulations can be run with `Float32` (#662)

### Documentation

- Documentation for GPU support was added (#660)
- A new user tutorial was added (#514)
- Several docs fixes (#663, #659, #637, #658, #664)

```
Thanks!

efaulhaber (dismissed):

svchb (dismissed):

efaulhaber (dismissed):
Fine by me. @LasNikas ?

svchb (approved):

PR #534, opened 2024-05-23
Rewrite `@threaded` macro to work with GPUs

efaulhaber:
See #484.

This is basically an alternative to #493.
Instead of defining a separate kernel for every threaded loop, I redefine the `@threaded` macro to run code multithreaded on the CPU or the GPU, depending on the type of the passed system.

To run a simulation on an NVidia GPU:
```julia
using CUDA, TrixiParticles

trixi_include("examples/fluid/dam_break_2d.jl", tspan=(0.0, 0.1), data_type=CuArray,
              saving_callback=nothing, saving_paper=nothing, neighborhood_search=nothing)
```


efaulhaber:
This PR is ready now. I still haven't added docs for `data_type` in `Semidiscretization` because the GPU feature is still not really useful without a GPU-compatible neighborhood search. I will add proper docs with a full GPU page once it's actually useful.

@lchristm It would be great if you could take a look at the macro stuff.

sloede:
> However, as we already briefly discussed before, the current approach basically uses the code optimized for CPUs and adds the KernelAbstractions.jl boilerplate around it to run it on a GPU which may not be optimal performance-wise. So it would be interesting to compare the performance of this approach to a highly optimized GPU implementation sometime in the future (e.g. for a single kernel) to see if there is a performance gap and how large it is.

Excellent point! I think in the future we may actually go the other way around: Optimize our code for GPU use and keep 🤞 it will be still OK-ish for CPUs :-)

efaulhaber:
We can still write explicit kernels for performance-relevant code. But the big question is: how do we optimize the code for the GPU? Through macros in the kernel signature? Then we need explicit kernels. Or maybe only through macros inside the kernel? Those might just work with the @threaded loops.

src/general/semidiscretization.jl
```diff
@@ -375,6 +375,29 @@ end
...
+@inline function wrap_v(v_ode, system, semi)
+    (; ranges_v) = semi
+
+    range = ranges_v[system_indices(system, semi)]
+
+    @boundscheck @assert length(range) == v_nvariables(system) * n_moving_particles(system)
+
+    return reshape(view(v_ode, range), (v_nvariables(system), n_moving_particles(system)))
+end
+
```
LasNikas, line 400: How about `return wrap_vector` in `wrap_u` and `wrap_v` and dispatch for `Array`s and non-`Array`s. 
`wrap_vector` doesn't care about `u_ode` or `v_ode`.

LasNikas (approved):
LGTM!

svchb (approved):

PR #539, opened 2024-05-29
Generalize surface normal calc

svchb:
Depends on #599 
extracted from #584


src/general/system.jl
```diff
@@ -151,3 +151,17 @@ end
...
 reset_callback_flag!(system) = system
+
+# Assuming a constant particle spacing one can calculate the number of neighbors within the
+# compact support for an undisturbed particle distribution.
+function number_density(::Val{D}, particle_spacing, compact_support) where {D}
+    throw(ArgumentError("Unsupported dimension: $D"))
+end
+
+@inline function number_density(::Val{2}, particle_spacing, compact_support)
+    return floor(Int64, pi * compact_support^2 / particle_spacing^2)
```
efaulhaber, line 162:
```suggestion
    return floor(Int, pi * compact_support^2 / particle_spacing^2)
```
Please use `Int` when `Int64` is not explicitly necessary. Also everywhere else in this PR.

src/schemes/boundary/dummy_particles/dummy_particles.jl
```diff
@@ -16,12 +16,12 @@ Boundary model for `BoundarySPHSystem`.
...
-- `correction`:     Correction method of the adjacent fluid system (see [Corrections](@ref corrections)).
-- `viscosity`:      Slip (default) or no-slip condition. See description below for further
-                    information.
-
+- `state_equation`:             This should be the same as for the adjacent fluid system
+                                (see e.g. [`StateEquationCole`](@ref)).
+- `correction`:                 Correction method of the adjacent fluid system (see [Corrections](@ref corrections)).
+- `viscosity`:                  Slip (default) or no-slip condition. See description below for further
+                                information.
+- `reference_particle_spacing`: The reference particle spacing used for weighting values at the boundary.
```
efaulhaber, line 24: 1. Please add this to the signature above.
2. This is completely unclear to me. Isn't this only used for surface tension? Then why is the surface tension not mentioned here?

src/schemes/boundary/dummy_particles/dummy_particles.jl
```diff
@@ -54,19 +55,33 @@ end
...
-             create_cache_model(correction, initial_density, NDIMS, n_particles)...)
+             create_cache_model(correction, initial_density, NDIMS, n_particles)...,
+             (; colorfield_bnd=zeros(ELTYPE, n_particles),
+              colorfield=zeros(ELTYPE, n_particles),
+              neighbor_count=zeros(ELTYPE, n_particles))...)
+
+    number_density_ = 0
+    if reference_particle_spacing > 0.0
+        number_density_ = number_density(Val(ndims(boundary_model)),
+                                         reference_particle_spacing,
```
efaulhaber, line 76: Please add a comment.

src/schemes/fluid/surface_tension.jl
```diff
@@ -1,4 +1,5 @@
-abstract type AkinciTypeSurfaceTension end
+abstract type SurfaceTensionModel end
```
efaulhaber, line 1:
```suggestion
abstract type SurfaceTension end
```

src/schemes/fluid/entropically_damped_sph/system.jl
```diff
@@ -58,6 +60,8 @@ struct EntropicallyDampedSPHSystem{NDIMS, ELTYPE <: Real, IC, M, DC, K, V, TV,
     pressure_acceleration_formulation :: PF
     transport_velocity                :: TV
     source_terms                      :: ST
+    surface_tension                   :: SRFT
+    surface_normal_method             :: SRFN
```
efaulhaber, line 64: Wouldn't it make sense to store the normal method inside the surface tension to avoid cluttering the struct?

src/schemes/fluid/entropically_damped_sph/system.jl
```diff
@@ -44,12 +44,14 @@ See [Entropically Damped Artificial Compressibility for SPH](@ref edac) for more
...
 struct EntropicallyDampedSPHSystem{NDIMS, ELTYPE <: Real, IC, M, DC, K, V, TV,
-                                   PF, ST, B, C} <: FluidSystem{NDIMS, IC}
+                                   PF, ST, SRFT, SRFN, B, C} <: FluidSystem{NDIMS, IC}
     initial_condition                 :: IC
     mass                              :: M # Vector{ELTYPE}: [particle]
     density_calculator                :: DC
     smoothing_kernel                  :: K
     smoothing_length                  :: ELTYPE
+    number_density                    :: Int64
+    color                             :: Int64
```
efaulhaber, line 54: Why do these need to be here? They are only used when using surface tension, no? Then they would be better off in the cache.

src/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,170 @@
...
+    nhs = get_neighborhood_search(system, neighbor_system, semi)
+
+    if smoothing_length != system.smoothing_length ||
+       smoothing_kernel !== system.smoothing_kernel
+        # TODO: this is really slow but there is no way to easily implement multiple search radia
+        search_radius = compact_support(smoothing_kernel, smoothing_length)
+        nhs = PointNeighbors.copy_neighborhood_search(nhs, search_radius,
+                                                      nparticles(system))
+        PointNeighbors.initialize!(nhs, system_coords, neighbor_system_coords)
+    end
```
efaulhaber, line 44: Do we want to support this with terrible performance, or rather not support it at all (yet)?

src/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,170 @@
+struct ColorfieldSurfaceNormal{ELTYPE, K}
```
efaulhaber, line 1: Please add docs.

efaulhaber (changes requested):
First review iteration, mostly focused on data structures.

src/general/system.jl
```diff
@@ -151,3 +151,17 @@ end
...
+end
+
+@inline @fastpow function number_density(::Val{3}, particle_spacing, compact_support)
+    return floor(Int64, 4.0 / 3.0 * pi * compact_support^3 / particle_spacing^3)
+end
```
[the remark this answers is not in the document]
svchb: Anyway i mixed the naming up. What about "ideal_neighbor_count"?

Its the number of particles within the compact support for an undisturbed particle distribution. This is used to find particles which have insufficient support. Here it is used to determine if particles are at the surface.

src/schemes/fluid/entropically_damped_sph/system.jl
svchb: While the surface normals are only used for surface tension, at this time, they are never actually called from the actual surface tension model. The surface normal calculation steps are always called from the systems.

src/schemes/fluid/surface_normal_sph.jl
svchb: I think this should be in here. Otherwise this just gets forgotten when this can be done more efficiently.

src/general/system.jl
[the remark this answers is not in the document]
efaulhaber: `ideal_neighbor_count` sounds very descriptive to me 👍

src/schemes/fluid/surface_normal_sph.jl
efaulhaber: Then please add an issue, so we don't forget.

src/schemes/boundary/dummy_particles/dummy_particles.jl
svchb: Its a general value that has nothing todo with surface tension beside being only used there at this time.

src/schemes/fluid/entropically_damped_sph/system.jl
svchb: They are currently only needed when a surface normal method is used.

svchb: The problem is though that these values are needed for all systems but not all systems have a surface normal or surface tension model activated.

src/schemes/fluid/entropically_damped_sph/system.jl
svchb: The surface normal calculations can be used in other things than the surface tension. 
1. Tracking of the interface 
2. Heatflux between phases
3. Free surface boundary conditions

Moving it to surface tension just makes it seem like it somehow depends on anything in surface tension but it actually doesn't. Also you need to actually declare surface normal methods for the solids as well but they don't have a surface tension. Declaring normal methods for solids is part of #666 since they are needed for the wetting model. These are than also different for elastic solids and rigid solids.

src/schemes/fluid/surface_normal_sph.jl
svchb: #668

src/schemes/fluid/entropically_damped_sph/system.jl
svchb: For example for a solid flat plate I declare:
boundary_system = BoundarySPHSystem(tank.boundary, boundary_model,
                                    adhesion_coefficient=1.0, surface_normal_method=StaticNormals((0.0, 1.0)))

svchb: So maybe we should rename this `normal_method` instead since this works better when applied to solid/boundaries?

src/schemes/fluid/surface_normal_sph.jl
```diff
@@ -1,11 +1,9 @@
...
-end
+@doc raw"""
+    ColorfieldSurfaceNormal()
 
-function ColorfieldSurfaceNormal(; smoothing_kernel, smoothing_length)
-    return ColorfieldSurfaceNormal(smoothing_kernel, smoothing_length)
-end
+Color field based computation of the interface normals.
+"""
+struct ColorfieldSurfaceNormal{} end
```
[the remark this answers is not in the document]
svchb: Ah I forgot on my version it is:
```
struct ColorfieldSurfaceNormal{ELTYPE}
    boundary_contact_threshold::ELTYPE
    interface_threshold::ELTYPE
    ideal_density_threshold::ELTYPE
end
```

src/schemes/fluid/entropically_damped_sph/system.jl
svchb: Currently there will be:
For fluids:
ColorfieldSurfaceNormal
And for Bnd:
StaticNormal (constant value for flat surfaces)
PrecalculatedNormal (curved surfaces)
ColorfieldSurfaceNormal (TLSPH)

efaulhaber: Then I agree it makes sense that each system has a `surface_normal_method`.

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,152 @@
...
+end
+
+@testset "Sphere Surface Normals" begin
+    # Test case 2: Particles arranged in a circle
+    particle_spacing = 0.25
+    radius = 1.0
+    center = (0.0, 0.0)
+    NDIMS = 2
+
+    # Create a SphereShape (which is a circle in 2D)
```
efaulhaber, line 99:
```suggestion
    # Create a `SphereShape`, which is a disk in 2D
```

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,152 @@
...
+    sphere_ic = SphereShape(particle_spacing, radius, center, 1000.0)
+
+    coordinates = sphere_ic.coordinates
+    velocity = zeros(NDIMS, size(coordinates, 2))
+    mass = sphere_ic.mass
+    density = sphere_ic.density
+
+    # To get some what accurate normals we increase the smoothing length unrealistically
```
efaulhaber, line 107:
```suggestion
    # To get somewhat accurate normals we increase the smoothing length unrealistically
```

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,152 @@
...
+    expected_normals = zeros(NDIMS, nparticles)
+    surface_particles = Int[]
+
+    # Compute expected normals and identify surface particles
+    for i in 1:nparticles
+        pos = coordinates[:, i]
+        r = pos - SVector(center...)
+        norm_r = norm(r)
+
+        if abs(norm_r - radius) < particle_spacing
```
efaulhaber, line 125:
```suggestion
        # If particle is on the surface of the circle
        if abs(norm_r - radius) < particle_spacing
```

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,152 @@
...
+    compute_and_test_surface_normals(system, semi, ode; NDIMS=NDIMS)
+
+    nparticles = size(coordinates, 2)
+    expected_normals = zeros(NDIMS, nparticles)
+    surface_particles = Int[]
+
+    # Compute expected normals and identify surface particles
+    for i in 1:nparticles
+        pos = coordinates[:, i]
+        r = pos - SVector(center...)
```
efaulhaber, line 122:
```suggestion
        r = pos .- center
```

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,152 @@
...
+        norm_computed = norm(computed_normals[:, i])
+        if norm_computed > 0
+            computed_normals[:, i] /= norm_computed
+        end
+    end
+
+    # Compare computed normals to expected normals for surface particles
+    for i in surface_particles
+        @test isapprox(computed_normals[:, i], expected_normals[:, i], atol=0.05)
+    end
```
efaulhaber, line 146:
```suggestion
    @test isapprox(computed_normals, expected_normals, atol=0.05)
```
The loop will generate a lot of tests.

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,152 @@
...
+
+    # Optionally, check that normals for interior particles are zero
+    # for i in setdiff(1:nparticles, surface_particles)
+    #     @test isapprox(norm(system.cache.surface_normal[:, i]), 0.0, atol=1e-4)
+    # end
```
efaulhaber, line 151: ?

efaulhaber (changes requested):

src/schemes/fluid/weakly_compressible_sph/system.jl
```diff
@@ -45,14 +45,17 @@ See [Weakly Compressible SPH](@ref wcsph) for more details on the method.
...
+       FluidSystem{NDIMS, IC}
     initial_condition                 :: IC
     mass                              :: MA     # Array{ELTYPE, 1}
     pressure                          :: P      # Array{ELTYPE, 1}
     density_calculator                :: DC
     state_equation                    :: SE
     smoothing_kernel                  :: K
     smoothing_length                  :: ELTYPE
+    ideal_neighbor_count              :: Int
+    color                             :: Int
```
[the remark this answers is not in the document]
svchb: Yes it will be used in #584

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,152 @@
...
+    v = TrixiParticles.wrap_v(v0_ode, system, semi)
+    u = TrixiParticles.wrap_u(u0_ode, system, semi)
+
+    # Compute the surface normals
+    TrixiParticles.compute_surface_normal!(system, surface_tension, v, u, v0_ode, u0_ode,
+                                           semi, 0.0)
+
+    # After computation, check that surface normals have been computed
+    @test all(isfinite.(system.cache.surface_normal))
+    @test all(isfinite.(system.cache.neighbor_count))
```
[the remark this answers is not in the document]
svchb: The inside will stay at 0.0. though.

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,152 @@
...
+
+    # Test the surface normals based on neighbor counts
+    for i in 1:nparticles
+        if system.cache.neighbor_count[i] < threshold
+            @test all(system.cache.surface_normal[:, i] .== 0.0)
+        else
+            # For the linear arrangement, surface normals may still be zero
+            # Adjust the test to account for this possibility
+            @test true
+        end
```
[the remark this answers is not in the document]
svchb: There is another condition that is used in #584. But I think I forgot to adjust the tests.

test/schemes/fluid/surface_normal_sph.jl
svchb: This is also part of #584

test/schemes/fluid/surface_normal_sph.jl
svchb: This doesn't work since the normals are not always exactly 0.0 on the inside.

test/schemes/fluid/surface_normal_sph.jl
svchb: I thought I used it there but this is rather tricky... Even with the special conditions that are implemented in #584

src/schemes/fluid/weakly_compressible_sph/system.jl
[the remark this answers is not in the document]
efaulhaber: As always, please remove it from this PR and add it when it's actually used.

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -96,18 +84,18 @@ end
     center = (0.0, 0.0)
     NDIMS = 2
 
-    # Create a SphereShape (which is a circle in 2D)
+    # Create a `SphereShape`, which is a circle in 2D
```
efaulhaber, line 87: A circle is the surface, a disk is a filled circle.

test/schemes/fluid/surface_normal_sph.jl
efaulhaber:
```julia
@test isapprox(computed_normals[:, surface_particles], expected_normals[:, surface_particles], atol=0.05)
```

Also, why are they not exactly zero on the inside? Does that make sense?

test/schemes/fluid/surface_normal_sph.jl
efaulhaber: And with a larger tolerance? There should be some test to ensure the inner values are not completely random.
And there should not be a commented out test in here.

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "TrixiParticles"
 uuid = "66699cd8-9c01-4e9d-a059-b96c86d16b3a"
 authors = ["erik.faulhaber <44124897+efaulhaber@users.noreply.github.com>"]
-version = "0.2.4-dev"
+version = "0.2.4"
```
efaulhaber, line 4: I think it makes more sense to wait for #529 before we release.

src/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,144 @@
...
+    # First we need to calculate the smoothed colorfield values
+    # TODO: move colorfield to extra step
+    # TODO: this is only correct for a single fluid
+    set_zero!(colorfield)
+
+    foreach_point_neighbor(system, neighbor_system,
+                           system_coords, neighbor_system_coords,
+                           nhs) do particle, neighbor, pos_diff, distance
+        colorfield[neighbor] += smoothing_kernel(system, distance)
+    end
```
efaulhaber, line 72:
```suggestion
    foreach_point_neighbor(neighbor_system, system,
                           neighbor_system_coords, system_coords,
                           nhs) do particle, neighbor, pos_diff, distance
        colorfield[particle] += smoothing_kernel(system, distance)
    end
```
Otherwise you will get race conditions on multiple threads.

src/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,144 @@
...
+
+    @threaded neighbor_system for bnd_particle in eachparticle(neighbor_system)
+        colorfield[bnd_particle] = colorfield[bnd_particle] / (colorfield[bnd_particle] +
```
efaulhaber, line 75: Why is `colorfield_bnd` initialized elsewhere and not here where `colorfield` is computed?

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,140 @@
...
+                                           semi, 0.0)
+
+    # After computation, check that surface normals have been computed and are not NaN or Inf
+    @test all(isfinite.(system.cache.surface_normal))
+    @test all(isfinite.(system.cache.neighbor_count))
+    @test size(system.cache.surface_normal, 1) == NDIMS
+
+    nparticles = size(u, 2)
+
+    # check the threshold has been applied correctly
```
efaulhaber, line 43:
```suggestion
    # Check that the threshold has been applied correctly
```
#133

test/schemes/fluid/surface_normal_sph.jl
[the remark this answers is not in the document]
efaulhaber: Can you please fix this here then?

src/schemes/fluid/fluid.jl
```diff
@@ -76,10 +76,27 @@ function calculate_dt(v_ode, u_ode, cfl_number, system::FluidSystem)
...
+    return nothing
+end
+
+@inline function surface_normal_method(system::FluidSystem)
+    return system.surface_normal_method
+end
+
+@inline function surface_normal_method(system)
+    return nothing
+end
```
efaulhaber, line 93: Never used.

src/schemes/boundary/system.jl
```diff
@@ -398,3 +398,29 @@ end
...
+                           neighborhood_search,
+                           points=eachparticle(system)) do particle, neighbor, pos_diff,
+                                                           distance
+        system.boundary_model.cache.colorfield_bnd[particle] += kernel(smoothing_kernel,
+                                                                       distance,
+                                                                       smoothing_length)
+        system.boundary_model.cache.neighbor_count[particle] += 1
+    end
+    return system
+end
```
efaulhaber, line 426: Please add a test.

src/general/system.jl
```diff
@@ -151,3 +151,17 @@ end
...
+# compact support for an undisturbed particle distribution.
+function ideal_neighbor_count(::Val{D}, particle_spacing, compact_support) where {D}
+    throw(ArgumentError("Unsupported dimension: $D"))
+end
+
+@inline function ideal_neighbor_count(::Val{2}, particle_spacing, compact_support)
+    return floor(Int, pi * compact_support^2 / particle_spacing^2)
+end
+
+@inline @fastpow function ideal_neighbor_count(::Val{3}, particle_spacing, compact_support)
```
efaulhaber, line 165: Please check the codecov report. This is not covered, but the 2D version is.

src/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,144 @@
...
+        cache.neighbor_count[particle] += 1
+    end
+
+    return system
+end
+
+# Section 2.2 in Akinci et al. 2013 "Versatile Surface Tension and Adhesion for SPH Fluids"
+# Note: This is the simplest form of normal approximation commonly used in SPH and comes
+# with serious deficits in accuracy especially at corners, small neighborhoods and boundaries
+function calc_normal_akinci!(system, neighbor_system::BoundarySystem, u_system, v,
```
efaulhaber, line 54: Please check Codecov report. This is missing tests.

efaulhaber (changes requested):

test/schemes/fluid/surface_normal_sph.jl
svchb: The problem is that not only the outer layer has the surface normal set but this goes some layers deep so the only possiblity is to eliminate several of the outer layers.

svchb: Than the inside is 0

src/schemes/fluid/weakly_compressible_sph/system.jl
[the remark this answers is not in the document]
svchb: Great unnecessary conflicts in 5 Branches...

test/schemes/fluid/surface_normal_sph.jl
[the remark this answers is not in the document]
svchb: what should I fix here?

Project.toml
svchb: Depends on how long that still will take...

test/schemes/fluid/surface_normal_sph.jl
svchb: That also doesn't work

svchb:
```
  Expression: isapprox(computed_normals[:, surface_particles], expected_normals[:, surface_particles], atol = 0.04)
   Evaluated: isapprox([0.3480639311878508 -0.3480639311878508 … 0.34806393118785095 -0.3480639311878508; 0.9374707994418061 0.9374707994418061 … -0.937470799441806 -0.937470799441806], [0.31622776601683794 -0.31622776601683794 … 0.31622776601683794 -0.31622776601683794; 0.9486832980505138 0.9486832980505138 … -0.9486832980505138 -0.9486832980505138]; atol = 0.04)
```

src/schemes/fluid/surface_normal_sph.jl
svchb: This is because the colorfield is constant on the boundary. This is changed in #666 where this is moved to a different function.

src/schemes/fluid/fluid.jl
svchb: This is used in #584

src/general/system.jl
svchb: A test for this exists in #584

src/schemes/boundary/system.jl
svchb: I have added a test but this test will change in all subsequent PRs

src/schemes/fluid/surface_normal_sph.jl
svchb: Testing this doesn't make much sense. Since there is no analytical value. And it will change in all subsequent PRs.

svchb: Let me restate that there is a analytical value but the results will not be close. Just closer than without it.

src/schemes/fluid/surface_normal_sph.jl
svchb: Doesn't work there will be no neighbors

test/schemes/fluid/surface_normal_sph.jl
efaulhaber: What happens between the surface and the inner layers where it's 0? Is this just a shorter normal then? Can you test that the normal either points in the correct direction or is almost zero?

src/schemes/fluid/weakly_compressible_sph/system.jl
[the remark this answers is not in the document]
efaulhaber: Rebasing should really eliminate all your problems.

test/schemes/fluid/surface_normal_sph.jl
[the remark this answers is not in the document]
efaulhaber: What I suggested.

test/schemes/fluid/surface_normal_sph.jl
efaulhaber: According to the docs of `isapprox`:
>  The norm keyword defaults to abs for numeric (x,y) and to LinearAlgebra.norm for arrays (where an alternative norm choice is sometimes useful).

In the loop, you checked the Linf norm of the difference, here you check the L2 norm, which becomes larger with the size of the vector when the Linf norm stays the same:
```julia
julia> norm(ones(100_000) - zeros(100_000))
316.22776601683796
```
For `isapprox`, this means:
```julia
julia> isapprox(ones(100_000), zeros(100_000), atol=300)
false

julia> isapprox(ones(100_000), zeros(100_000), atol=400)
true

julia> isapprox(ones(100_000), zeros(100_000), atol=1.0, norm=x -> norm(x, Inf))
true

julia> isapprox(ones(100_000), zeros(100_000), atol=0.9, norm=x -> norm(x, Inf))
false
```

src/schemes/fluid/fluid.jl
efaulhaber: Then add it there.

src/general/system.jl
efaulhaber: Can you add it here or does it require other code from #584?

src/schemes/boundary/system.jl
efaulhaber: I think that is better than reducing coverage here and potentially forgetting about it.

src/schemes/fluid/surface_normal_sph.jl
efaulhaber: There must be some useful way to test this. Maybe with a high resolution and very large smoothing length?

src/schemes/fluid/surface_normal_sph.jl
efaulhaber: `foreach_point_neighbor(neighbor_system, system, ..., particles=eachparticle(neighbor_system))`, or the loop will go over all moving particles, which are none for a boundary system.

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,152 @@
...
+
+@testset "Surface Normal Computation" begin
+    # Test case 1: Simple linear arrangement of particles
+    nparticles = 5
+    particle_spacing = 1.0
+    NDIMS = 2
+
+    coordinates = zeros(NDIMS, nparticles)
+    for i in 1:nparticles
+        coordinates[:, i] = [i * particle_spacing, 0.0]
```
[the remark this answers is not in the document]
svchb: Yes but in one of the follow up PRs please.

test/schemes/fluid/surface_normal_sph.jl
svchb: This is correct behavior. The normals should be set for 2-4 layers from the surface, but this is not constant. The more advanced methods in #680 fix this by doing an **additional** cleanup run across the particles that have normals set. This is done in this way since the method in #608 is like 4x slower and the calculated normals are less accurate but the surface is identified accurately.

src/schemes/fluid/fluid.jl
svchb: These changes added now 35 commits

src/schemes/boundary/system.jl
svchb: I am currently working on #680. There are only merge errors accumulating in the rest in the chain...

src/general/system.jl
svchb: Its indirectly tested since it is used in a tested function.

src/schemes/fluid/surface_normal_sph.jl
svchb: It should be noted that I am currently only working on a much different code. So I am completely confused. 
Anyway this function is renamed in a future version since it is merged with the other normal method and then split because it became too long. The condition that is used here is also changed.

What I mean here is that the results close to the wall in all versions are not reliable anyway. So the test would break down to testing that a nonzero value exists in this version.

docs/make.jl
```diff
@@ -140,8 +140,8 @@ makedocs(sitename="TrixiParticles.jl",
                      "Util" => joinpath("general", "util.md")
                  ],
                  "Systems" => [
-                     "Discrete Element Method (Solid)" => joinpath("systems",
-                                                                   "dem.md"),
+                     "Fluid Models" => joinpath("systems", "fluid.md"),
+                     "Discrete Element Method (Solid)" => joinpath("systems", "dem.md"),
```
efaulhaber, line 144: Would it make more sense to make it like this:
- Systems
  - Fluid Models
    - Overview (with the new content that you added in "fluid.md")
    - WCSPH
    - EDAC
  - DEM

Now we have "Fluid Models" on the same level as "WCSPH" and "EDAC", which is confusing to me, as these are fluid models.

docs/src/systems/fluid.md
```diff
@@ -0,0 +1,89 @@
...
+### Akinci-based intra-particle force surface tension and wall adhesion model
+The work by Akinci proposes three forces:
+- a cohesion force
+- a surface area minimization force
+- a wall adhesion force
+
+The classical model is composed of the curvature minimization and cohesion force.
+
+#### Cohesion force
+The model calculates the cohesion force based on the distance between particles and the support radius ``h_c``.
```
efaulhaber, line 40:
```suggestion
The model calculates the cohesion force based on the support radius ``h_c`` and the distance between particles.
```
I read this as "distance between (particles and support radius)".

docs/src/systems/fluid.md
```diff
@@ -0,0 +1,89 @@
...
+This force is determined using two distinct regimes within the support radius:
+- For particles closer than half the support radius,
+  a repulsive force is calculated to prevent particle clustering too tightly,
```
efaulhaber, line 43:
```suggestion
  a repulsive force is calculated to prevent particles from clustering too tightly,
```

examples/fluid/dam_break_oil_film_2d.jl
```diff
@@ -30,13 +30,16 @@ oil_viscosity = ViscosityMorris(nu=nu_sim_oil)
 trixi_include(@__MODULE__, joinpath(examples_dir(), "fluid", "dam_break_2d.jl"),
               sol=nothing, fluid_particle_spacing=fluid_particle_spacing,
               viscosity=ViscosityMorris(nu=nu_sim_water), smoothing_length=smoothing_length,
-              gravity=gravity, density_diffusion=nothing, sound_speed=sound_speed)
+              gravity=gravity, tspan=tspan, density_diffusion=nothing,
+              sound_speed=sound_speed, prefix="")
```
efaulhaber, line 34: I'm confused. Why is the `tspan` needed when `sol = nothing`?

examples/fluid/falling_water_spheres_2d.jl
```diff
@@ -40,7 +40,7 @@ sphere2 = SphereShape(fluid_particle_spacing, sphere_radius, sphere2_center,
 
 # ==========================================================================================
 # ==== Fluid
-fluid_smoothing_length = 1.0 * fluid_particle_spacing
+fluid_smoothing_length = 1.0 * fluid_particle_spacing - eps()
```
efaulhaber, line 43: Why is this necessary? It shouldn't be necessary, that would be confusing. Is this https://github.com/trixi-framework/PointNeighbors.jl/issues/19?

src/schemes/boundary/dummy_particles/dummy_particles.jl
```diff
@@ -16,12 +17,12 @@ Boundary model for `BoundarySPHSystem`.
...
-- `correction`:     Correction method of the adjacent fluid system (see [Corrections](@ref corrections)).
-- `viscosity`:      Slip (default) or no-slip condition. See description below for further
-                    information.
-
+- `state_equation`:             This should be the same as for the adjacent fluid system
+                                (see e.g. [`StateEquationCole`](@ref)).
+- `correction`:                 Correction method of the adjacent fluid system (see [Corrections](@ref corrections)).
+- `viscosity`:                  Slip (default) or no-slip condition. See description below for further
+                                information.
+- `reference_particle_spacing`: The reference particle spacing used for weighting values at the boundary, which currently is only needed when using surface tension.
```
efaulhaber, line 25:
```suggestion
- `reference_particle_spacing`: The reference particle spacing used for weighting values at the boundary,
                                which currently is only needed when using surface tension.
```

src/schemes/fluid/entropically_damped_sph/rhs.jl
```diff
@@ -48,8 +51,17 @@ function interact!(dv, v_particle_system, u_particle_system,
...
+                                                   particle, neighbor, pos_diff, distance)
+
+        dv_adhesion = adhesion_force(surface_tension_a, particle_system, neighbor_system,
+                                     particle, neighbor, pos_diff, distance)
+
         for i in 1:ndims(particle_system)
-            dv[i, particle] += dv_pressure[i] + dv_viscosity_[i] + dv_convection[i]
+            dv[i, particle] += dv_pressure[i] + dv_viscosity_[i] + dv_convection[i] +
+                               dv_surface_tension[i] +
+                               dv_adhesion[i]
```
efaulhaber, line 64:
```suggestion
                               dv_surface_tension[i] + dv_adhesion[i]
```

src/schemes/boundary/dummy_particles/dummy_particles.jl
```diff
@@ -38,35 +39,51 @@ BoundaryModelDummyParticles(AdamiPressureExtrapolation, ViscosityAdami)
...
-    viscosity          :: V
-    correction         :: COR
-    cache              :: C
+    pressure             :: VECTOR # Vector{ELTYPE}
+    hydrodynamic_mass    :: VECTOR # Vector{ELTYPE}
+    state_equation       :: SE
+    density_calculator   :: DC
+    smoothing_kernel     :: K
+    smoothing_length     :: ELTYPE
+    ideal_neighbor_count :: Int
```
efaulhaber, line 48: This field is never used? Neither for the other structs where you added it. Or am I missing something?

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,269 @@
+function create_boundary_system(coordinates, particle_spacing, state_equation, kernel,
```
efaulhaber, line 1:
```suggestion
# Create a platform below the fluid (at a distance `walldistance`)
function create_boundary_system(coordinates, particle_spacing, state_equation, kernel,
```

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,269 @@
...
+    v0_ode, u0_ode = ode.u0.x
+    v = TrixiParticles.wrap_v(v0_ode, system, semi)
+    u = TrixiParticles.wrap_u(u0_ode, system, semi)
+
+    TrixiParticles.compute_surface_normal!(system, surface_tension, v, u, v0_ode, u0_ode,
+                                           semi, 0.0)
+
+    # After computation, check that surface normals have been computed and are not NaN or Inf
+    @test all(isfinite.(system.cache.surface_normal))
+    @test all(isfinite.(system.cache.neighbor_count))
```
efaulhaber, line 93:
```suggestion
    @test all(isfinite, system.cache.surface_normal)
    @test all(isfinite, system.cache.neighbor_count)
```

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,269 @@
...
+    @test size(system.cache.surface_normal, 1) == NDIMS
+
+    nparticles = size(u, 2)
+
+    # Check that the threshold has been applied correctly
+    threshold = 2^ndims(system) + 1
+
+    # # Test the surface normals based on neighbor counts
```
efaulhaber, line 101:
```suggestion
    # Test the surface normals based on neighbor counts.
```

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,269 @@
...
+    # Test that surface normals are zero when there are not enough neighbors.
+    # For the linear arrangement, surface normals may still be zero
+    # when we have more neighbors than the threshold.
+    for i in 1:nparticles
+        @test system.cache.neighbor_count[i] >= threshold ||
+              all(system.cache.surface_normal[:, i] .== 0.0)
+    end
```
efaulhaber, line 108:
```suggestion
    @test all(i -> system.cache.neighbor_count[i] >= threshold || iszero(system.cache.surface_normal[:, i]), 1:nparticles)
```
This will generate only one test in the summary.

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,269 @@
...
+    mass = fill(1.0, nparticles)
+    fluid_density = 1000.0
+    density = fill(fluid_density, nparticles)
+
+    system, bnd_system, semi, ode = create_fluid_system(coordinates, velocity, mass,
+                                                        density,
+                                                        particle_spacing;
+                                                        NDIMS=NDIMS)
+
+    compute_and_test_surface_normals(system, semi, ode; NDIMS=NDIMS)
```
efaulhaber, line 131: This only tests that normals outside the threshold are zero, right?
Can you please add some test for the other normals, to check that they are not completely arbitrary?
Otherwise, this test only tests that everything runs without errors.

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,269 @@
...
+        norm_computed = norm(computed_normals[:, i])
+        if norm_computed > 0
+            computed_normals[:, i] /= norm_computed
+        end
+    end
+
+    # Compare computed normals to expected normals for surface particles
+    @test isapprox(computed_normals[:, surface_particles],
+                   expected_normals[:, surface_particles], norm=x -> norm(x, Inf),
+                   atol=0.05)
```
efaulhaber, line 191:
```suggestion
    @test isapprox.(computed_normals[:, surface_particles],
                   expected_normals[:, surface_particles], atol=0.05)
```
My earlier suggestions were stupid.

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,269 @@
...
+
+    # Optionally, check that normals for interior particles are zero
+    # for i in setdiff(1:nparticles, surface_particles)
+    #     @test isapprox(norm(system.cache.surface_normal[:, i]), 0.0, atol=1e-4)
+    # end
+end
+
+@testset "Sphere Surface Normals with wall" begin
```
efaulhaber, line 199:
```suggestion
@testset "Sphere Surface Normals with Wall" begin
```

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,269 @@
...
+    for i in surface_particles
+        norm_computed = norm(computed_normals[:, i])
+        if norm_computed > 0
+            computed_normals[:, i] /= norm_computed
+        end
+    end
+
+    # Boundary system
+    bnd_color = bnd_system.boundary_model.cache.colorfield_bnd
+    # this is only true since it assumed that the color is 1
```
efaulhaber, line 257:
```suggestion
    # This is only true since it assumed that the color is 1
```

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -0,0 +1,269 @@
...
+    @test all(bnd_color .>= 0.0)
+
+    # Compare computed normals to expected normals for surface particles (with wall we have a larger error)
+    @test isapprox(computed_normals[:, surface_particles],
+                   expected_normals[:, surface_particles], norm=x -> norm(x, Inf),
+                   atol=0.5)
```
efaulhaber, line 263:
```suggestion
    @test isapprox.(computed_normals[:, surface_particles],
                   expected_normals[:, surface_particles], atol=0.5)
```

test/schemes/fluid/surface_normal_sph.jl
efaulhaber: Can you add a comment then, please?

efaulhaber (changes requested):
A bunch of small remarks, otherwise this is ready to merge from my side.

examples/fluid/falling_water_spheres_2d.jl
svchb: Its not necessary but the result is better.

src/schemes/boundary/dummy_particles/dummy_particles.jl
svchb: I think it disappeared in this PR because of changes to the remove invalid normals function.

svchb:
```
    # We remove invalid normals (too few neighbors) to reduce the impact of underdefined normals
    for particle in each_moving_particle(system)
        # A corner has that many neighbors assuming a regular 2 * r distribution and a compact_support of 4r
        if cache.neighbor_count[particle] < 2^ndims(system) + 1
            cache.surface_normal[1:ndims(system), particle] .= 0
        end
    end
```

svchb: I have to check the other PR it makes sense to move this here.

test/schemes/fluid/surface_normal_sph.jl
svchb: Not sure what you try here? But this is wrong.

test/schemes/fluid/surface_normal_sph.jl
svchb: see above

src/schemes/boundary/dummy_particles/dummy_particles.jl
svchb: Hmm I checked not really

test/schemes/fluid/surface_normal_sph.jl
efaulhaber: Why is it wrong? Instead of comparing two vectors with a specific norm, we can just pointwise apply `isapprox` to check every element of the vector.

examples/fluid/falling_water_spheres_2d.jl
efaulhaber: Then please add a comment. Otherwise, it's very confusing why we need to subtract almost nothing from  the particle spacing.

src/schemes/boundary/dummy_particles/dummy_particles.jl
efaulhaber: So please remove it here.

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -89,50 +90,25 @@ function compute_and_test_surface_normals(system, semi, ode; NDIMS=2)
...
+    @test all(isfinite, system.cache.neighbor_count)
     @test size(system.cache.surface_normal, 1) == NDIMS
 
     nparticles = size(u, 2)
 
     # Check that the threshold has been applied correctly
     threshold = 2^ndims(system) + 1
 
-    # # Test the surface normals based on neighbor counts
+    # Test the surface normals based on neighbor counts
```
efaulhaber, line 102:
```suggestion
    # Test the surface normals based on neighbor counts.
```

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -191,12 +167,14 @@ end
                    atol=0.05)
 
     # Optionally, check that normals for interior particles are zero
```
efaulhaber, line 169:
```suggestion
    # Optionally, check that normals for interior particles are zero.
```

test/schemes/fluid/surface_normal_sph.jl
```diff
@@ -263,6 +241,8 @@ end
                    atol=0.5)
 
     # Optionally, check that normals for interior particles are zero
```
efaulhaber, line 243:
```suggestion
    # Optionally, check that normals for interior particles are zero.
```

test/schemes/fluid/surface_normal_sph.jl
efaulhaber: I meant
```suggestion
    @test all(isapprox.(computed_normals[:, surface_particles],
                   expected_normals[:, surface_particles], atol=0.05))
```
```
julia> isapprox(x, y, atol=10eps(), norm=x -> norm(x, Inf))
true

julia> isapprox(x, y, atol=10eps())
false

julia> all(isapprox.(x, y, atol=10eps()))
true
```

efaulhaber (changes requested):
LGTM! Just the two unresolved conversations left.
@LasNikas do you want to have a look as well?

test/schemes/fluid/surface_normal_sph.jl
svchb: Sphere Surface Normals: Error During Test at /media/svenb/Extreme SSD/GIT/TrixiParticles3/test/schemes/fluid/surface_normal_sph.jl:169
  Expression evaluated to non-Boolean

efaulhaber (approved):
Ready from my side.

src/schemes/boundary/dummy_particles/dummy_particles.jl
```diff
@@ -16,12 +17,13 @@ Boundary model for `BoundarySPHSystem`.
...
-- `viscosity`:      Slip (default) or no-slip condition. See description below for further
-                    information.
-
+- `state_equation`:             This should be the same as for the adjacent fluid system
+                                (see e.g. [`StateEquationCole`](@ref)).
+- `correction`:                 Correction method of the adjacent fluid system (see [Corrections](@ref corrections)).
+- `viscosity`:                  Slip (default) or no-slip condition. See description below for further
+                                information.
+- `reference_particle_spacing`: The reference particle spacing used for weighting values at the boundary,
+                                which currently is only needed when using surface tension.
```
LasNikas, line 26: If it is only needed when using surface tension, it might make more sense to initialize the surface tension objects with it, right? Or am I missing something?
Then we wouldn't need an extra field for all systems that use surface tension.

LasNikas (changes requested):

src/schemes/boundary/dummy_particles/dummy_particles.jl
```diff
@@ -54,19 +56,25 @@ end
...
+    ELTYPE = eltype(smoothing_length)
     n_particles = length(initial_density)
 
     cache = (; create_cache_model(viscosity, n_particles, NDIMS)...,
              create_cache_model(initial_density, density_calculator)...,
-             create_cache_model(correction, initial_density, NDIMS, n_particles)...)
+             create_cache_model(correction, initial_density, NDIMS, n_particles)...,
+             (; colorfield_bnd=zeros(ELTYPE, n_particles),
+              colorfield=zeros(ELTYPE, n_particles),
+              neighbor_count=zeros(ELTYPE, n_particles))...)
```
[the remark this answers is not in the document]
svchb: Because there is no simple way to make this happen without passing the surface tension model to all of the systems.

src/schemes/boundary/dummy_particles/dummy_particles.jl
svchb: This only works than if we would add additional initialisation code to semidiscretization. Also this would then need to be handled in the adaption and MPI implementation.

svchb: It should also be noted this is no longer clear in this PR why this causes issues because all of the problematic stuff that makes this necessary to be a system property was removed because Erik had me remove it because it is currently unused and is now in #584.

svchb: Also a discussion about this is already somewhere in this PR....

src/schemes/fluid/entropically_damped_sph/system.jl
```diff
@@ -69,7 +71,9 @@ struct EntropicallyDampedSPHSystem{NDIMS, ELTYPE <: Real, IC, M, DC, K, V, TV,
                                          alpha=0.5, viscosity=nothing,
                                          acceleration=ntuple(_ -> 0.0,
                                                              ndims(smoothing_kernel)),
-                                         source_terms=nothing, buffer_size=nothing)
+                                         source_terms=nothing, surface_tension=nothing,
+                                         surface_normal_method=nothing, buffer_size=nothing,
+                                         reference_particle_spacing=0.0)
```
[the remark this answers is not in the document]
svchb: Interesting it was there before...

src/schemes/boundary/dummy_particles/dummy_particles.jl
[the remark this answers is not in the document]
LasNikas: Okay I see. 
In 3D the number of boundary particles can be very high.
Is it possible to set these vectors to an empty vector and `resize!` them in `initialize!(boundary_system, nhs)`?
There you call the function `initialize_colorfield!` anyway

src/schemes/boundary/dummy_particles/dummy_particles.jl
LasNikas: In this PR `reference_particle_spacing` is never used.
However, `reference_particle_spacing` is the particle spacing in the initial configuration?
Then i might make more sense to add a function ?
```julia
reference_particle_spacing(system) = system.initial_condition.particle_spacing
```

svchb: Yes and no with adaption it would be the local particle spacing. 
I think Erik had me remove everything that used this...

LasNikas: can you please create an issue for this?

LasNikas (approved):

src/schemes/fluid/entropically_damped_sph/system.jl
```diff
@@ -22,26 +23,32 @@ See [Entropically Damped Artificial Compressibility for SPH](@ref edac) for more
...
+                                or `SVector` that is to be added to the acceleration of that particle.
+                                See, for example, [`SourceTermDamping`](@ref).
+                                Note that these source terms will not be used in the calculation of the
+                                boundary pressure when using a boundary with
+                                [`BoundaryModelDummyParticles`](@ref) and [`AdamiPressureExtrapolation`](@ref).
+                                The keyword argument `acceleration` should be used instead for
+                                gravity-like source terms.
+- `surface_tension`:            Surface tension model used for this SPH system. (default: no surface tension)
+- `surface_normal_method`:      The surface normal method to be used for this SPH system.
+                                (default: no surface normal method or ColorfieldSurfaceNormal() if a surface_tension model is used)
```
efaulhaber, line 48: Should be 
```
`ColorfieldSurfaceNormal()`
```

svchb: I have fixed this in #713

PR #549, opened 2024-06-19
Set to development version 0.1.4

svchb:


Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "TrixiParticles"
 uuid = "66699cd8-9c01-4e9d-a059-b96c86d16b3a"
 authors = ["erik.faulhaber <44124897+efaulhaber@users.noreply.github.com>"]
-version = "0.1.3-dev"
+version = "0.1.4-dev"
```
efaulhaber, line 4:
```suggestion
version = "0.1.4-pre"
```
Do we want to use this? Just like in Trixi.jl? Why did we use `dev`?

svchb: Michael wrote I should use `-dev` when I pushed the 0.1.1 release.

PR #550, opened 2024-06-21
Use new v0.3 release of PointNeighbors.jl

efaulhaber:


svchb (dismissed):

src/general/semidiscretization.jl
```diff
@@ -1,21 +1,18 @@
 """
-    Semidiscretization(systems...; neighborhood_search=GridNeighborhoodSearch,
-                       periodic_box_min_corner=nothing, periodic_box_max_corner=nothing,
-                       threaded_nhs_update=true)
+    Semidiscretization(systems...; neighborhood_search=GridNeighborhoodSearch{NDIMS}())
```
LasNikas, line 2: Please remove the keywords for the periodic box in the docs.

Or maybe, IMO, it is better to keep this keywords and create the nhs internally with the bounding box. Or only pass the `PeriodicBox`? Is there a reason why you changed it this way?

svchb (dismissed):

svchb (changes requested):
Waiting for further breaking changes.

efaulhaber:
More breaking changes with https://github.com/trixi-framework/PointNeighbors.jl/pull/42.

PR #561, opened 2024-06-27
Release v0.2

efaulhaber:
Waiting on breaking changes in https://github.com/trixi-framework/PointNeighbors.jl/pull/42.
Based on #562.


src/general/semidiscretization.jl
```diff
@@ -285,18 +270,18 @@ function semidiscretize(semi, tspan; reset_threads=true, data_type=nothing)
...
-            # Get the neighborhood search for this system
-            neighborhood_search = get_neighborhood_search(system, semi)
+    foreach_system(semi) do system
+        # Get the neighborhood search for this system
+        neighborhood_search = get_neighborhood_search(system, semi)
 
-            # Initialize this system
-            initialize!(system, neighborhood_search)
+        # Initialize this system
+        initialize!(system, neighborhood_search)
```
LasNikas, line 281: I was confused why we pass `neighborood_search` here. 
This is a system initialization and the only initialized system here is the tlsph system.
The nhs is initialized above, so can we change this to `initialize!(system)`?

LasNikas: or better `initialize!(system, semi)`

efaulhaber: We need a neighborhood search to initialize the TLSPH system (compute correction matrix and stuff).
And passing the neighborhood search is more convenient than passing the semidiscretization when writing benchmarks in PointNeighbors.jl, where we don't necessarily have a semidiscretization.

Issue #488, opened 2024-04-02
Move viscosity from BoundaryModel to BoundarySPHSystem


svchb:

svchb:
For the current setups it doesn't make sense since it would add viscosity to TLSPH.

PR #564, opened 2024-06-27
Move viscosity to boundary sph system

svchb:
#488


src/schemes/boundary/dummy_particles/dummy_particles.jl
```diff
@@ -12,15 +12,11 @@
                        See description above for more information.
 - `density_calculator`: Strategy to compute the hydrodynamic density of the boundary particles.
                         See description below for more information.
-- `smoothing_kernel`: Smoothing kernel should be the same as for the adjacent fluid system.
-- `smoothing_length`: Smoothing length should be the same as for the adjacent fluid system.
```
efaulhaber, line 16: These arguments are still there.

src/schemes/boundary/dummy_particles/dummy_particles.jl
```diff
@@ -30,21 +26,19 @@ boundary_model = BoundaryModelDummyParticles(densities, masses, AdamiPressureExt
 
 # No-slip condition
```
efaulhaber, line 27: This example doesn't make sense anymore.

src/schemes/boundary/dummy_particles/dummy_particles.jl
```diff
@@ -306,15 +272,22 @@ end
     boundary_model.pressure[particle] = max(boundary_model.state_equation(density), 0.0)
 end
 
+reset_wall_velocity!(viscosity, system) = system
+
+@inline function reset_wall_velocity!(viscosity::ViscosityAdami, system)
+    set_zero!(system.cache.wall_velocity)
+    return system.cache
+end
```
efaulhaber, line 280:
```suggestion
@inline function reset_wall_velocity!(system, viscosity::ViscosityAdami)
    set_zero!(system.cache.wall_velocity)
    return system
end
```
`system` is the argument that is modified.

src/schemes/boundary/system.jl
```diff
@@ -1,5 +1,6 @@
 """
-    BoundarySPHSystem(initial_condition, boundary_model; movement=nothing, adhesion_coefficient=0.0)
+    BoundarySPHSystem(initial_condition, boundary_model; movement=nothing, adhesion_coefficient=0.0,
+                      viscosity=nothing, smoothing_kernel=nothing, smoothing_length=0.0)
```
efaulhaber, line 3:
```suggestion
                      viscosity=nothing)
```

src/schemes/boundary/system.jl
```diff
@@ -12,49 +13,56 @@ The interaction between fluid and boundary particles is specified by the boundar
 - `movement`: For moving boundaries, a [`BoundaryMovement`](@ref) can be passed.
 - `adhesion_coefficient`: Coefficient specifying the adhesion of a fluid to the surface.
    Note: currently it is assumed that all fluids have the same adhesion coefficient.
+- `viscosity`:  Free-slip (default) or no-slip condition. See description above for further
+                information.
```
efaulhaber, line 17:
```suggestion
- `viscosity`:  Free-slip (default) or no-slip condition.
```
There is no further information in the description. I'm wondering where this comes from, as the models (where you copied this from) also don't contain any additional information in the description.

src/schemes/boundary/system.jl
```diff
@@ -161,13 +169,24 @@ function create_cache_boundary(::BoundaryMovement, initial_condition)
...
 end
 
+create_cache_model(viscosity, n_particles, n_dims) = (;)
+
+function create_cache_model(viscosity::ViscosityAdami, n_particles, n_dims)
+    ELTYPE = eltype(viscosity.epsilon)
+
+    wall_velocity = zeros(ELTYPE, n_dims, n_particles)
+
+    return (; wall_velocity)
```
efaulhaber, line 179: This should be inside `create_cache_boundary`. @LasNikas I think we decided a while ago that we want to have only one `create_boundary` call in the constructor, right?

efaulhaber (changes requested):
Good idea to do that before v0.2!

src/schemes/boundary/system.jl
LasNikas: Yes, I wanted to remark that as well

LasNikas:
I haven't looked in detail but this PR is a very good idea.  However, how do we handle FSI when the viscosity is stored in the system and not in the model?

svchb:
Doesn't seem to make sense than.

PR #569, opened 2024-06-28
Fix viscosity calculation when using two different fluids

### LINKED ISSUE #540: Fix viscosity for two different fluids
Also see #488

svchb:
Fix #540


test/schemes/fluid/viscosity.jl
```diff
@@ -1,3 +1,4 @@
+include("../../test_util.jl")
```
LasNikas, line 1: why is this necessary?

examples/fluid/dam_break_two_fluids_2d.jl
```diff
@@ -0,0 +1,127 @@
...
+extra_callback = nothing
+
+use_reinit = false
+density_reinit_cb = use_reinit ?
+                    DensityReinitializationCallback(semi.systems[1], interval=10) :
+                    nothing
+stepsize_callback = StepsizeCallback(cfl=0.9)
+
+callbacks = CallbackSet(info_callback, saving_callback, stepsize_callback, extra_callback,
+                        density_reinit_cb, saving_paper)
```
LasNikas, line 123: do we need this here?

examples/fluid/dam_break_two_fluids_2d.jl
```diff
@@ -0,0 +1,127 @@
+# 2D dam break simulation with an oil film on top
```
LasNikas, line 1: maybe call this file `dam_break_oil_film_2d.jl`?

src/schemes/fluid/viscosity.jl
```diff
@@ -4,20 +4,7 @@ function dv_viscosity(particle_system, neighbor_system,
...
-                        sound_speed, m_a, m_b, rho_a, rho_b, grad_kernel)
-end
-
-function dv_viscosity(particle_system, neighbor_system::OpenBoundarySPHSystem,
-                      v_particle_system, v_neighbor_system,
-                      particle, neighbor, pos_diff, distance,
-                      sound_speed, m_a, m_b, rho_a, rho_b, grad_kernel)
-    # No viscosity in the open boundary system. Use viscosity of the fluid system.
-    viscosity = viscosity_model(particle_system)
+    viscosity = viscosity_model(neighbor_system, particle_system)
```
LasNikas, line 7: nice, I think that fixes the weird `include` order in `src/schemes/schemes.jl`.
is the following possible?
```julia
include("fluid/fluid.jl")
include("boundary/boundary.jl")
include("boundary/open_boundary/boundary_zones.jl")
include("boundary/open_boundary/system.jl")
```
Or even better: Put the open boundaries in `boundary/boundary.jl`?

LasNikas (approved):

src/schemes/fluid/viscosity.jl
```diff
@@ -181,12 +169,15 @@ end
 end
 
 @inline function (viscosity::ViscosityMorris)(c, v_diff, pos_diff, distance, rho_mean,
-                                              rho_a, rho_b, h, grad_kernel)
-    (; epsilon, nu) = viscosity
+                                              rho_a, rho_b, h, grad_kernel, particle_system,
+                                              neighbor_system)
```
efaulhaber, line 173: I don't like that the viscosity, which is supposed to be an elementary function, needs systems. Can you instead pass `nu_a` and `nu_b`?

src/schemes/fluid/viscosity.jl
efaulhaber, line 7: Why is this using the viscosity model of the neighbor system? Can you please add a comment?

examples/fluid/dam_break_oil_film_2d.jl
```diff
@@ -0,0 +1,114 @@
+# 2D dam break simulation with an oil film on top
```
efaulhaber, line 1: Can't you do something like
```julia
trixi_include("dam_break_2d.jl", sol=nothing)

oil_system = ...

semi = Semidiscretization(fluid_system, oil_system, boundary_system)
ode = semidiscretize(semi, tspan)

sol = solve(ode, CarpenterKennedy2N54(williamson_condition=false),
            dt=1.0, # This is overwritten by the stepsize callback
            save_everystep=false, callback=callbacks);
```

efaulhaber (changes requested):

src/schemes/fluid/viscosity.jl
svchb: I don't know this was like this before. So changing that again changes the validation files again :D.

src/schemes/fluid/viscosity.jl
svchb: Then we need to dispatch on the viscosity models in interaction.

src/schemes/fluid/viscosity.jl
efaulhaber: I think because we need to use the boundary viscosity for fluid-boundary interaction, for example.
I guess it would then make sense to change this to `viscosity_model(particle_system, neighbor_system)` and let `viscosity_model` return the viscosity of the `neighbor_system` by default, except when this is an open boundary system.

src/schemes/fluid/viscosity.jl
```diff
@@ -4,7 +4,7 @@ function dv_viscosity(particle_system, neighbor_system,
                       v_particle_system, v_neighbor_system,
                       particle, neighbor, pos_diff, distance,
                       sound_speed, m_a, m_b, rho_a, rho_b, grad_kernel)
-    viscosity = viscosity_model(neighbor_system, particle_system)
+    viscosity = viscosity_model(particle_system, neighbor_system)
```
LasNikas, line 7: wait, why is this now the `particle_system`? This should be the neighbors system, since we use the viscosity of the interacting system. When I want to interact with a boundary system with slip-condition, it should be no viscosity.

svchb: because this dispatches to the neighbor_system viscosity for a boundary system

LasNikas: Is this correct? IMO the default should be the viscosity model of the neighbor system, no?

svchb: you are right

examples/fluid/falling_water_spheres_2d.jl
```diff
@@ -90,6 +90,6 @@ callbacks = CallbackSet(info_callback, saving_callback)
 
 # Use a Runge-Kutta method with automatic (error based) time step size control.
 sol = solve(ode, RDPK3SpFSAL35(),
-            abstol=1e-6, # Default abstol is 1e-6
+            abstol=1e-7, # Default abstol is 1e-6
```
LasNikas, line 93: why do you need to change this?
What do the falling spheres have to do with the viscosity? I'm curious

svchb: This case uses two fluid systems which interact.

svchb: It also uses adami and Monaghan viscosity at the same time. So the solution changes.

LasNikas: Ah sorry I confused it with the falling spheres example

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -484,3 +484,4 @@ end
 
 @inline viscosity_model(system::OpenBoundarySPHSystem, neighbor_system::FluidSystem) = neighbor_system.viscosity
 @inline viscosity_model(system::OpenBoundarySPHSystem, neighbor_system::BoundarySystem) = neighbor_system.boundary_model.viscosity
+@inline viscosity_model(system, neighbor_system::OpenBoundarySPHSystem) = system.viscosity
```
efaulhaber, line 487:
```suggestion
# When available, use the viscosity of the neighbor system in the interaction.
# So, e.g. for fluid-boundary interaction, we use the boundary's viscosity, which can be used
# to model no-slip boundary conditions.
@inline viscosity_model(system::OpenBoundarySPHSystem, neighbor_system::FluidSystem) = neighbor_system.viscosity
@inline viscosity_model(system::OpenBoundarySPHSystem, neighbor_system::BoundarySystem) = neighbor_system.boundary_model.viscosity
@inline viscosity_model(system, neighbor_system::OpenBoundarySPHSystem) = system.viscosity
```
From the perspective of a new user, it would be great if you could add this comment here and in all other locations where you define this function.
Or would it make sense to define all of them together somewhere?

efaulhaber (changes requested):
Looks good! Just one minor remark.

src/schemes/boundary/open_boundary/system.jl
svchb: Only if we want to move these functions to viscosity.jl.

efaulhaber: Is this possible or do we not have access to the system types there?

svchb: ah yes that would be an issue with the **OpenBoundarySPHSystem**

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -482,6 +482,8 @@ function wrap_reference_function(constant_vector_, ::Val{NDIMS}) where {NDIMS}
     return constant_vector(coords, t) = SVector{NDIMS}(constant_vector_)
 end
 
+# The open boundary system either uses the fluids viscosity model or at the boundary the model
+# of the boundary which is the normal behavior to incorporate the boundaries impact in the viscosity term of the rhs.
```
efaulhaber, line 486:
```suggestion
# The open boundary system either uses the fluid's viscosity model or, at the boundary, the model
# of the boundary, which is the normal behavior to incorporate the boundary's impact in the viscosity term of the rhs.
```
Can you please use the same comment in the other two locations as well?

efaulhaber (changes requested):

src/schemes/boundary/open_boundary/system.jl
svchb: why the other two locations have nothing to do with the open boundary system?

efaulhaber: This comment has nothing to do with the open boundary system. It's the reason why we use the neighbor's viscosity and not the system's viscosity. But maybe we need to reformulate it to make this clearer?

svchb: The first part of the sentence is only relevant for the open boundary system?

efaulhaber: Ah, oops. I didn't see that. Then my original suggestion?
```
# When available, use the viscosity of the neighbor system in the interaction.
# So, e.g. for fluid-boundary interaction, we use the boundary's viscosity, which can be used
# to model no-slip boundary conditions.
```
Plus the other suggestion below for the open boundary system.

efaulhaber (approved):

PR #498, opened 2024-04-18
Some changes to how AdamiPressureExtrapolation is calculated

svchb:
- add optional offset
- use Bernoulli Equation to obtain dynamic pressure contribution

PR #581, opened 2024-07-24
Make Adami loop flipping optimization optional

efaulhaber:
Depends on #498.

I now get an error of `0.0` for the dam break validation between a single threads and 128 threads on an AMD Threadripper 3990X. So we can generate all validation files on as many threads as we have available.


efaulhaber:
@svchb do you remember what we did with this? Did we implement it somewhere else? Or did we just forget about it?

svchb:
The optimization is there because there are cases where the larger system is looped over and there are no/only a small number of neighbors between systems. In my simulations this made up to a 20% difference in run time.

One side effect is though that if you run the validation cases with different numbers of threads the results change.

efaulhaber:
Yes, I know. This PR is based on your PR, so the diff is messy. But did we make the flipping optional and disable it for validiation? Or did we forget about it? Or did we decide not to do this for some reason?

efaulhaber:
Just rebased to make the diff useful.

svchb:
No this was not merged. You also did not create all the validation files correctly. I think this was the latest on this PR.

efaulhaber:
> You also did not create all the validation files correctly.

These have to be recreated in #775 anyway, so I suggest merging this PR as is, and then I only have to recreate them once.

LasNikas (approved):

svchb (approved):

PR #582, opened 2024-07-25
Add argument `t`  to `source_terms`

LasNikas:


src/schemes/boundary/open_boundary/system.jl
```diff
@@ -165,8 +165,6 @@ end
 
 update_callback_used!(system::OpenBoundarySPHSystem) = system.update_callback_used[] = true
 
-@inline source_terms(system::OpenBoundarySPHSystem) = nothing
-
```
efaulhaber, line 169: Why is this not needed anymore?

efaulhaber (changes requested):

src/schemes/boundary/open_boundary/system.jl
LasNikas: It was never needed. Dispatch handles this case

efaulhaber (approved):

svchb (approved):

PR #589, opened 2024-08-06
Prepare release 0.2.1

svchb:


svchb:
wait for #586 to merge

Project.toml
```diff
@@ -41,8 +41,8 @@ GPUArraysCore = "0.1"
 JSON = "0.21"
 KernelAbstractions = "0.9"
 MuladdMacro = "0.2"
-PointNeighbors = "0.4.2"
-Polyester = "0.7.5"
+PointNeighbors = "0.4.5"
+Polyester = "0.7.15"
```
efaulhaber, line 45: But do we need this version? Or is it running just fine with 0.7.5?

Project.toml
```diff
@@ -41,8 +41,8 @@ GPUArraysCore = "0.1"
 JSON = "0.21"
 KernelAbstractions = "0.9"
 MuladdMacro = "0.2"
-PointNeighbors = "0.4.2"
-Polyester = "0.7.5"
+PointNeighbors = "0.4.5"
```
efaulhaber, line 44: 0.4.2 is enough. We don't need to require 0.4.5. No new features of 0.4.5 are used in TrixiP.

efaulhaber (changes requested):

Project.toml
svchb: There was an error in 0.7.6 that's why we put 0.7.5 here. We can revert to 0.7 if you want?

Project.toml
svchb: Should we put 0.4 than instead?

Project.toml
svchb: But I don't see any drawback by requiring the newer version since there are no downstream package changes (I checked)

Project.toml
efaulhaber: No, we get unsatisfiable compat requirements with 0.4.1 and TrixiParticles. We need at least 0.4.2.

Project.toml
svchb: Ok so lets progress over the error fix than to 0.7.8?

efaulhaber: Did we? `"0.7.5"` matches every `0.7.X` version since `0.7.5`, so also `0.7.6`.

svchb: Than this is wrong anyway?

efaulhaber: If there was indeed a bug in 0.7.6 and it was fixed in 0.7.8, then I'm fine with setting this to 0.7.8.

svchb: So 0.7.10 than like trixi.jl? Since there is another error fixed in 0.7.10 that is relevant?

efaulhaber: Then 0.7.10.

efaulhaber (approved):

PR #592, opened 2024-08-06
Cleanup examples

svchb:


examples/fluid/hydrostatic_water_column_2d.jl
```diff
@@ -31,7 +31,8 @@ tank = RectangularTank(fluid_particle_spacing, initial_fluid_size, tank_size, fl
 smoothing_length = 1.2 * fluid_particle_spacing
 smoothing_kernel = SchoenbergCubicSplineKernel{2}()
 
-viscosity = ArtificialViscosityMonaghan(alpha=0.02, beta=0.0)
+alpha = 0.02
+viscosity = ArtificialViscosityMonaghan(alpha=alpha, beta=0.0)
```
efaulhaber, line 35: Why?

svchb: Because alpha is used in a trixi_include.

test/examples/examples.jl
```diff
@@ -2,41 +2,87 @@
...
-                                           joinpath(examples_dir(), "fluid",
-                                                    "hydrostatic_water_column_2d.jl"))
-            @test sol.retcode == ReturnCode.Success
-            @test count_rhs_allocations(sol, semi) == 0
-        end
+            hydrostatic_water_column_tests = Dict(
+                "default" => (),
+                "with source term damping" => (source_terms=SourceTermDamping(damping_coefficient=1e-4),),
+                "with SummationDensity" => (fluid_density_calculator=SummationDensity(),
+                                            clip_negative_pressure=true),
```
LasNikas, line 10: The artificial viscosity is missing, no?

test/examples/examples.jl
```diff
@@ -2,41 +2,87 @@
...
+                          joinpath(examples_dir(), "fluid",
+                                   "hydrostatic_water_column_2d.jl"),
+                          fluid_system=nothing, sol=nothing, semi=nothing, ode=nothing)
+
+            viscosity = ViscosityAdami(nu=alpha * smoothing_length * sound_speed / 8)
+            fluid_system = EntropicallyDampedSPHSystem(tank.fluid, smoothing_kernel,
+                                                       smoothing_length, sound_speed,
+                                                       viscosity=viscosity,
+                                                       density_calculator=ContinuityDensity(),
+                                                       acceleration=(0.0, -gravity))
```
LasNikas, line 68: can you please add a test with `SummationDensity`?

LasNikas: In the original paper, they use the `SummationDensity`.  But it's also possible with the `ContinuityDensity` (and way faster)

test/examples/examples.jl
```diff
@@ -2,6 +2,121 @@
...
+                "with ViscosityAdami and SummationDensity" => (
+                                                               # from 0.02*10.0*1.2*0.05/8
+                                                               viscosity=ViscosityAdami(nu=0.0015),
+                                                               fluid_density_calculator=SummationDensity(),
+                                                               clip_negative_pressure=true),
+                "with ViscosityMorris and SummationDensity" => (
+                                                                # from 0.02*10.0*1.2*0.05/8
+                                                                viscosity=ViscosityMorris(nu=0.0015),
+                                                                fluid_density_calculator=SummationDensity(),
+                                                                clip_negative_pressure=true),
```
LasNikas, line 90: EDAC doesn't need a state equation, so `clip_negaive_pressure` is not necessary

Also, you define the system above with viscosity and density calculator and pass it to `trixi_include`. The `kwargs` from the `Dict` are then never used in the example, right?

LasNikas: How about one test set called `hydrostatic water column`. Above you can then create the dict for both edac and wcsph.

LasNikas (approved):

src/general/semidiscretization.jl
```diff
@@ -715,19 +715,27 @@ function update_nhs!(neighborhood_search,
             points_moving=(true, neighbor.ismoving[]))
 end
 
+# If declared as a union with the `TotalLagrangianSPHSystem` version will be ambiguous with another dispatch
```
efaulhaber, line 718:
```suggestion
# This function is the same as the one below to avoid ambiguous dispatch when using `Union`
```

src/general/semidiscretization.jl
```diff
@@ -715,19 +715,27 @@ function update_nhs!(neighborhood_search,
...
+    #
+    # Boundary coordinates only change over time when `neighbor.ismoving[]`.
+    # The current coordinates of fluids and solids change over time.
+    update!(neighborhood_search, system,
+            current_coordinates(u_system, system),
+            current_coordinates(u_neighbor, neighbor),
+            points_moving=(system.ismoving[], true))
 end
 
+# If declared as a union with the `FluidSystem` version will be ambiguous with another dispatch
```
efaulhaber, line 735:
```suggestion
# This function is the same as the one above to avoid ambiguous dispatch when using `Union`
```

test/examples/examples.jl
```diff
@@ -2,6 +2,96 @@
...
+            # import variables into scope
+            trixi_include(@__MODULE__,
+                          joinpath(examples_dir(), "fluid",
+                                   "hydrostatic_water_column_2d.jl"),
+                          fluid_system=nothing, sol=nothing, semi=nothing, ode=nothing)
+
+            hydrostatic_water_column_tests = Dict(
+                "(WCSPH) default" => (),
+                "(WCSPH) with source term damping" => (source_terms=SourceTermDamping(damping_coefficient=1e-4),),
+                "(WCSPH) with SummationDensity" => (fluid_density_calculator=SummationDensity(),
```
efaulhaber, line 15:
```suggestion
                "WCSPH default" => (),
                "WCSPH with source term damping" => (source_terms=SourceTermDamping(damping_coefficient=1e-4),),
                "WCSPH with SummationDensity" => (fluid_density_calculator=SummationDensity(),
```
Why the parentheses?

test/examples/examples.jl
```diff
@@ -2,6 +2,96 @@
 # but without checking the correctness of the solution.
 @testset verbose=true "Examples" begin
     @testset verbose=true "Fluid" begin
+        @trixi_testset "fluid/hydrostatic_water_column_2d.jl" begin
+            # import variables into scope
```
efaulhaber, line 6:
```suggestion
            # Import variables into scope
```

test/examples/examples.jl
```diff
@@ -2,6 +2,96 @@
...
+                    @test_nowarn_mod trixi_include(@__MODULE__,
+                                                   joinpath(examples_dir(), "fluid",
+                                                            "hydrostatic_water_column_2d.jl");
+                                                   kwargs...)
+
+                    @test sol.retcode == ReturnCode.Success
+                    @test count_rhs_allocations(sol, semi) == 0
+                end
+            end
+        end
```
efaulhaber, line 93: This is an excellent idea! A very elegant solution.

test/examples/examples.jl
```diff
@@ -77,34 +131,86 @@
...
+                                           tspan=(0.0, 0.05),
+                                           surface_tension=SurfaceTensionAkinci(surface_tension_coefficient=0.025),
+                                           fluid_particle_spacing=0.5 *
+                                                                  fluid_particle_spacing,
+                                           smoothing_kernel=SchoenbergCubicSplineKernel{2}(),
+                                           smoothing_length=0.5 * fluid_particle_spacing,
+                                           correction=AkinciFreeSurfaceCorrection(fluid_density),
+                                           density_diffusion=nothing,
+                                           adhesion_coefficient=0.05,
+                                           sound_speed=100.0) [
```
efaulhaber, line 201: Can't you combine these with the test above?

efaulhaber (changes requested):

efaulhaber (approved):
Have you checked that all these `trixi_include` commands give you the simulations you want? It's easy to miss something and end up with an incorrect simulation setup.

svchb:
> Have you checked that all these `trixi_include` commands give you the simulations you want? It's easy to miss something and end up with an incorrect simulation setup.

yes I have just scrolled through again. As far as I know everything should be fine. At least all of the values that are defined are used.

PR #594, opened 2024-08-07
Fix threaded winding number calculation

efaulhaber:
> Due to its packed storage format, concurrent access to the elements of a BitArray where at least one of them is a write is not thread-safe.

https://docs.julialang.org/en/v1/base/arrays/#Base.BitArray

```julia
julia> function f!(x)
           Threads.@threads :static for i in eachindex(x)
               x[i] = true
           end
       end
f! (generic function with 1 method)

julia> x = falses(1000);

julia> f!(x)

julia> count(x)
1000

julia> x = falses(1000);

julia> f!(x)

julia> count(x)
931
```


svchb (changes requested):
add also an entry to news.md so that this version is not empty

LasNikas (approved):
good catch! Thanks a lot

svchb (approved):

PR #601, opened 2024-08-14
Steady state callback

LasNikas:
I tested the callback with the  `pipe_flow_2d.jl` example. It terminates after `~0.8sec`.
I consider the change in kinetic energy in a `0.002*10` interval.
```julia
pp_callback = PostprocessCallback(; dt=0.02, kinetic_energy)

steady_state = SteadyStateCallback(; dt=0.002, interval_size=10, abstol=1.0e-8, reltol=1.0e-6)
```

image


src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+falls below the threshold specified by `abstol, reltol`.
+
+# Keywords
+- `interval=0`:     Check steady state condition every `interval` time steps.
+- `dt=0.0`:         Check steady state condition  in regular intervals of `dt` in terms
+                    of integration time by adding additional `tstops`.
+- `interval_size`:  The interval in which the change of the kinetic energy is considered.
+                    `interval_size` is a (integer) multiple of `interval` or `dt`.
+- `abstol`:         Absolute tolerance.
+- `reltol`:         Relative tolerance.
```
svchb, line 15: Add here the actual condition with the mse in form of an equation.

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+"""
+mutable struct SteadyStateCallback{I, RealT <: Real}
```
svchb, line 17:
```suggestion
mutable struct SteadyStateCallback{I, ELTYPE<: Real}
```

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+    interval      :: I
+    abstol        :: RealT
+    reltol        :: RealT
+    previous_ekin :: Vector{Float64}
```
svchb, line 21:
```suggestion
    abstol        :: ELTYPE
    reltol        :: ELTYPE
    previous_ekin :: Vector{ELTYPE}
```

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+    ekin = 0.0
+    foreach_system(semi) do system
+        v = wrap_v(v_ode, system, semi)
+
+        # Calculate kintetic energy
+        for particle in each_moving_particle(system)
+            velocity = current_velocity(v, system, particle)
+            ekin += 0.5 * system.mass[particle] * dot(velocity, velocity)
+        end
+    end
```
svchb, line 140: use the function from custom_quantities

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+        end
+
+        mse /= interval_size
+
+        threshold = abstol + reltol * ekin
+
+        terminate = mse <= threshold
+    end
+
+    return terminate
```
svchb, line 162:
```suggestion
    return false
```

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+        mse = 0.0
+        for index in 1:interval_size
+            mse += (previous_ekin[index] - ekin)^2
+        end
+
+        mse /= interval_size
+
+        threshold = abstol + reltol * ekin
+
+        terminate = mse <= threshold
```
svchb, line 159:
```suggestion
        if mse <= abstol + reltol * ekin
          return true
       
       # Pop old kinetic energy
       popfirst!(previous_ekin)
```

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+
+        # Calculate MSE only over the `interval_size`
+        mse = 0.0
+        for index in 1:interval_size
+            mse += (previous_ekin[index] - ekin)^2
+        end
+
+        mse /= interval_size
+
+        threshold = abstol + reltol * ekin
```
svchb, line 157:
```suggestion
```

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+
+    terminate = false
+
+    if interval_size_condition(cb, integrator)
+        popfirst!(previous_ekin)
+
+        # Calculate MSE only over the `interval_size`
+        mse = 0.0
+        for index in 1:interval_size
+            mse += (previous_ekin[index] - ekin)^2
```
svchb, line 152:
```suggestion
            mse += (previous_ekin[index] - ekin)^2/interval_size
```

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+        end
+
+        mse /= interval_size
```
svchb, line 155:
```suggestion
```

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+            velocity = current_velocity(v, system, particle)
+            ekin += 0.5 * system.mass[particle] * dot(velocity, velocity)
+        end
+    end
+
+    push!(previous_ekin, ekin)
+
+    terminate = false
+
+    if interval_size_condition(cb, integrator)
```
svchb, line 146:
```suggestion
    if length(previous_ekin) == interval_size
```

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+        # Calculate kintetic energy
+        for particle in each_moving_particle(system)
+            velocity = current_velocity(v, system, particle)
+            ekin += 0.5 * system.mass[particle] * dot(velocity, velocity)
+        end
+    end
+
+    push!(previous_ekin, ekin)
+
+    terminate = false
```
svchb, line 144:
```suggestion
```

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+            mse += (previous_ekin[index] - ekin)^2
+        end
+
+        mse /= interval_size
+
+        threshold = abstol + reltol * ekin
+
+        terminate = mse <= threshold
+    end
+
```
svchb, line 161:
```suggestion
# Add current kinetic energy       
push!(previous_ekin, ekin)
```

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+        v = wrap_v(v_ode, system, semi)
+
+        # Calculate kintetic energy
+        for particle in each_moving_particle(system)
+            velocity = current_velocity(v, system, particle)
+            ekin += 0.5 * system.mass[particle] * dot(velocity, velocity)
+        end
+    end
+
+    push!(previous_ekin, ekin)
```
svchb, line 142:
```suggestion
```

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
...
+
+    terminate = false
+
+    if interval_size_condition(cb, integrator)
+        popfirst!(previous_ekin)
```
svchb, line 147:
```suggestion
```

svchb (changes requested):
There is one mistake in the condition since you pop/push the current ekin one value is always 0.0 which will effectively reduce the interval size by one.

src/callbacks/steady_state.jl
efaulhaber: Why do you want to do this after computing the MSE?
Why not
```
if too large
    popfirst!(previous_ekin)
end

push!(previous_ekin, ekin)
```

efaulhaber (changes requested):

src/callbacks/steady_state.jl
svchb: Because we don't have to compare the current value with the current value because that will be 0.

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,175 @@
+"""
+    SteadyStateCallback(; interval::Integer=0, dt=0.0, interval_size::Integer=10,
```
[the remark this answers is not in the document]
svchb: ConvergenceConditionCallBack
ConvergenceTerminationCallBack
TerminationConditionCallBack

src/callbacks/steady_state.jl
```diff
@@ -0,0 +1,163 @@
...
+end
+
+# `affect!` (`PeriodicCallback`)
+function (cb::SteadyStateCallback)(integrator)
+    steady_state_condition!(cb, integrator) || return nothing
+
+    print_summary(integrator)
+
+    # `terminate!(integrator)` terminates the simulation immediately and might cause an error message.
+    integrator.opts.maxiters = integrator.iter
```
LasNikas, line 111: This causes a warning message.
A more elegant solution would be to initially (in `initialize!`)  sort the `CallbackSet` in such a way that the `SteadyStateCallback` is last.  
The set is stored in `integrator.opts.callback`. However, I have no idea how to check if the actual callback is the last element or not.

Any idea @efaulhaber  and @svchb ?

src/callbacks/steady_state.jl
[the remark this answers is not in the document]
LasNikas: `SteadyStateDetector`
`SteadyStateTerminator`

src/callbacks/steady_state.jl
LasNikas: Another (simpler) solution:
`@suppress integrator.opts.maxiters = integrator.iter`

from https://github.com/JuliaIO/Suppressor.jl

svchb: I guess the only way would be to implement a custom solve function which wraps the ODE solve?

LasNikas: My prefered approach is something like:

```julia
function initialize_cb!(cb::DiscreteCallback{<:Any,
                                             <:PeriodicCallbackAffect{<:SteadyStateCallback}},
                        u, t, integrator)
    callback_set = integrator.opts.callback.discrete_callbacks

    last(callback_set) isa typeof(cb) && return nothing

    # Make sure that the last callback is a `SteadyStateCallback` to avoid an error message when terminating
    new_set = (filter(i -> !(i isa typeof(cb), callback_set))...,
               filter(i -> i isa typeof(cb), callback_set)...)

    integrator.opts.callback.discrete_callbacks = new_set

    return nothing
end
```

However, I think we can't just mutate the `CallbackSet`.

@efaulhaber what's your opinion?

svchb: Otherwise a solution would be to adjust the CallBacks upstream?

efaulhaber: Why not use `terminate!`? This is the recommended option:
https://docs.sciml.ai/DiffEqDocs/stable/features/callback_functions/#Example-2:-Terminating-an-Integration
grafik

LasNikas: See comment in code above

src/callbacks/steady_state.jl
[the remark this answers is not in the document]
efaulhaber: As pointed out by @gregorgassner, "convergence" might be misleading. I also think the name should end with "callback".
So we have:
- `SteadyStateCallback`
- `TerminationConditionCallback`
- `SteadyStateReachedCallback` (by @gregorgassner)

`TerminationConditionCallback` is more general, but the callback is not. This could be a good name if we want to make the callback more general in the future.
`SteadyStateReachedCallback` makes the current use case very clear.
As I said before, I'm not a fan of `SteadyStateCallback`.

What do you think, @LasNikas , @svchb ?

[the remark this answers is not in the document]
LasNikas: `SteadyStateReachedCallback` sounds good for me

src/callbacks/steady_state.jl
efaulhaber: `terminate!` was fixed in https://github.com/SciML/DiffEqCallbacks.jl/pull/229.
This is the better solution, which does show the misleading warning.
However, before we can use the latest version of DiffEqCallbacks, we first have to wait for https://github.com/JuliaSIMD/StrideArrays.jl/issues/92.

test/callbacks/steady_state_reached.jl
```diff
@@ -0,0 +1,69 @@
...
+    end
+
+    @testset verbose=true "every `interval`" begin
+        steady_state_reached = SteadyStateReachedCallback(; interval=1, interval_size=10,
+                                                          abstol=1.0e-5, reltol=1.0e-4)
+        trixi_include(@__MODULE__, joinpath(examples_dir(), "fluid", "pipe_flow_2d.jl"),
+                      extra_callback=steady_state_reached, tspan=(0.0, 1.5))
+
+        @test sol.t[end] < 1.0
+    end
```
svchb, line 18: This should be in example. Otherwise this will generate a too high coverage report.

svchb (changes requested):

svchb (approved):

efaulhaber (approved):

PR #606, opened 2024-08-20
Skip empty systems

svchb:
With open boundaries a system can be empty. Lets skip them in the IO step.

Extracted from #604


efaulhaber (approved):
Doesn't look like a great solution to me, as VTK files are missing for some steps. But considering that VTK writing currently doesn't work for 0 particles, it's at least better than before.

LasNikas (approved):

PR #584, opened 2024-07-29
Implement Morris surface tension model

svchb:

PR #607, opened 2024-08-20
Add ideal gas equation

svchb:
depends on #599 
extracted from #584 
output (1)
output (2)


examples/fluid/dam_break_2d.jl
```diff
@@ -35,7 +35,8 @@ tank_size = (floor(5.366 * H / boundary_particle_spacing) * boundary_particle_sp
 fluid_density = 1000.0
 sound_speed = 20 * sqrt(gravity * H)
 state_equation = StateEquationCole(; sound_speed, reference_density=fluid_density,
-                                   exponent=1, clip_negative_pressure=false)
+                                   exponent=1, clip_negative_pressure=false,
+                                   background_pressure=0.0)
```
efaulhaber, line 39: This is a free surface simulation. Why would we use a background pressure there? That would just cause the fluid to expand or contract at the beginning of the simulation.

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
```diff
@@ -49,3 +49,65 @@ function inverse_state_equation(state_equation::StateEquationCole, pressure)
 
     return reference_density * tmp^(1 / exponent)
 end
+
+@inline sound_speed(eos) = eos.sound_speed
+
+@doc raw"""
+    StateEquationIdealGas(; gas_constant, temperature, gamma)
+Equation of state to describe the relationship between pressure and density
```
efaulhaber, line 57:
```suggestion
    StateEquationIdealGas(; gas_constant, temperature, gamma)

Equation of state to describe a proportional relationship between pressure and density
```

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
```diff
@@ -49,3 +49,65 @@ function inverse_state_equation(state_equation::StateEquationCole, pressure)
...
+of a gas using the Ideal Gas Law.
+# Keywords
+- `gas_constant`: Specific gas constant (R) for the gas, typically in J/(kg*K).
```
efaulhaber, line 60:
```suggestion

# Keywords
- `gas_constant`: Specific gas constant ``R`` for the gas, typically in ``\mathrm{J/(kg \, K)}``.
```

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
```diff
@@ -49,3 +49,65 @@ function inverse_state_equation(state_equation::StateEquationCole, pressure)
...
+- `temperature` : Absolute temperature of the gas in Kelvin.
+- `gamma`       : Heat-capacity ratio
+This struct calculates the pressure of a gas from its density using the formula:
```
efaulhaber, line 63:
```suggestion

This struct calculates the pressure of a gas from its density using the formula:
```
Also, the equations below don't render correctly.

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
```diff
@@ -49,3 +49,65 @@ function inverse_state_equation(state_equation::StateEquationCole, pressure)
...
+\[ P = \rho \cdot R \cdot T \]
+where \( P \) is pressure, \( \rho \) is density, \( R \) is the gas constant, and \( T \) is temperature.
+Note:
+For basic WCSPH this boils down to the assumption of a linear pressure-density relationship.
```
efaulhaber, line 67: I don't understand this sentence. This state equation is always linear, even when not using basic WCSPH.

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
```diff
@@ -49,3 +49,65 @@ function inverse_state_equation(state_equation::StateEquationCole, pressure)
...
+    function StateEquationIdealGas(; gas_constant, temperature, gamma)
+        new{typeof(gas_constant)}(gas_constant, temperature, gamma)
+    end
+end
+
+function (state_equation::StateEquationIdealGas)(density)
+    (; gas_constant, temperature) = state_equation
+    pressure = density * gas_constant * temperature
+    return pressure
+end
```
efaulhaber, line 83: 1. Why is there no background pressure? The SE by Cole is also designed to work with a reference density, at which the pressure is zero. For this state equation to be useful for SPH where we work with relative pressures, it should read `pressure = (density - reference_density) * gas_constant * temperature + background_pressure`.
2. Why does this state equation use the gas constant and doesn't calculate it from the speed of sound like the SE by Cole? This makes it very annoying to switch between state equation, as one has to look up the equation to convert sound speed to gas constant.
3. Please add tests. Codecov reports all of this as uncovered.

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
```diff
@@ -49,3 +49,65 @@ function inverse_state_equation(state_equation::StateEquationCole, pressure)
...
+
+# This version is for simulations that include a temperature.
+function (state_equation::StateEquationIdealGas)(density, internal_energy)
+    (; gamma) = state_equation
+    pressure = (gamma - 1.0) * density * internal_energy
+    return pressure
+end
```
efaulhaber, line 90: This is never used in TrixiP?

src/schemes/fluid/weakly_compressible_sph/system.jl
```diff
@@ -215,7 +215,10 @@ end
     return system.pressure[particle]
 end
 
-@inline system_sound_speed(system::WeaklyCompressibleSPHSystem) = system.state_equation.sound_speed
+@inline system_sound_speed(system::WeaklyCompressibleSPHSystem) = sound_speed(system.state_equation)
+@inline system_sound_speed(system::WeaklyCompressibleSPHSystem, pressure, density) = sound_speed(system.state_equation,
```
efaulhaber, line 219: Also never used.

examples/fluid/dam_break_2phase_2d.jl
```diff
@@ -0,0 +1,95 @@
...
+using OrdinaryDiffEq
+
+# Size parameters
+H = 0.6
+W = 2 * H
+
+gravity = 9.81
+tspan = (0.0, 2.0)
+
+# Resolution
```
efaulhaber, line 13:
```suggestion
# ==========================================================================================
# ==== Resolution
```
Please follow the conventions of other example files.
For example, `gravity` and `tspan` should go below this under "Experiment Setup".

src/schemes/fluid/weakly_compressible_sph/system.jl
svchb: Would be needed if we actually used a method with compressibility.

examples/fluid/dam_break_2d.jl
svchb: I think I forgot to remove it again.

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
svchb: 1. I don't know anymore.
2. c^2/(gamma * T) I think i forgot that i could also do this

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
svchb: I can remove it.

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
svchb: No I know again why I did that I wanted to avoid setting the physical sound speed.

svchb: But than I forgot to change it when I decided against doing that.

examples/fluid/dam_break_2phase_2d.jl
```diff
@@ -63,21 +67,18 @@ end
...
 air_system = union(air_system, air_system2)
 
-# We use background_pressure here to prevent negative pressure regions in the gas phase.
 air_system_system = WeaklyCompressibleSPHSystem(air_system, fluid_density_calculator,
                                                 StateEquationCole(; sound_speed,
                                                                   reference_density=air_density,
                                                                   exponent=1,
                                                                   clip_negative_pressure=false,
-                                                                  background_pressure=1000),
+                                                                  background_pressure=0),
```
efaulhaber, line 75: Is it ever useful to overwrite the background pressure here? This is unphysical if we don't do the same for the fluid, no?

examples/fluid/dam_break_2phase_2d.jl
```diff
@@ -0,0 +1,96 @@
...
+                                                                  reference_density=air_density,
+                                                                  exponent=1,
+                                                                  clip_negative_pressure=false,
+                                                                  background_pressure=0),
+                                                smoothing_kernel, smoothing_length,
+                                                viscosity=air_viscosity,
+                                                acceleration=(0.0, -gravity))
+
+# air_system_system = WeaklyCompressibleSPHSystem(air_system, fluid_density_calculator,
+#                                          StateEquationIdealGas(; sound_speed, reference_density=air_density, gamma=1.4, background_pressure=0),
```
efaulhaber, line 81: This here is identical apart from the state equation, isn't it? Can you just define the state equation before the system and add the ideal gas equation commented out?

test/schemes/fluid/weakly_compressible_sph/state_equation.jl
```diff
@@ -1,3 +1,4 @@
+include("../../../test_util.jl")
```
efaulhaber, line 1: This is already included.

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
```diff
@@ -49,3 +49,49 @@ function inverse_state_equation(state_equation::StateEquationCole, pressure)
 
     return reference_density * tmp^(1 / exponent)
 end
+
+@doc raw"""
+    StateEquationIdealGas(;sound_speed, reference_density, gamma, background_pressure=0.0)
```
efaulhaber, line 54:
```suggestion
    StateEquationIdealGas(; sound_speed, reference_density, gamma, background_pressure=0.0)
```
And can you please include `clip_negative_pressure`?

efaulhaber (changes requested):

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
```diff
@@ -49,3 +49,50 @@ function inverse_state_equation(state_equation::StateEquationCole, pressure)
...
+    (; reference_density, sound_speed, gamma, background_pressure) = state_equation
+    pressure = (density - reference_density) * sound_speed^2 / gamma + background_pressure
+
+    # This is determined statically and has therefore no overhead
+    if clip_negative_pressure(state_equation)
+        return max(0.0, pressure)
+    end
+
+    return pressure
+end
```
LasNikas, line 91: This function is almost identical with `StateEquationCole` and `exponent=1` (described here in the [docs](https://trixi-framework.github.io/TrixiParticles.jl/dev/systems/weakly_compressible_sph/#equation_of_state)) except the `gamma` in the bulk modulus.

Or Am I completely missing something?

I don't want to hold back the review process, but why not adding a kwarg `gamma` to the `StateEquationCole` instead of creating a new object?. 
The default is then `gamma=exponent` but can be overwritten.

svchb: While this is true I am not sure this makes a lot of sense. 
1. Everyone expects there to be an ideal gas state equation when you calculate a gas.
2. This is really quirky because it would require some error checks. E.g. if gamma and exponent are set etc...
3. As soon as there is heat flux the equation changes to versions with temperature (which were removed because they are unused at this time)

efaulhaber: Agreed. From a user's perspective, this is more intuitive.

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
```diff
@@ -49,3 +49,50 @@ function inverse_state_equation(state_equation::StateEquationCole, pressure)
...
+                           clip_negative_pressure=false)
+
+Equation of state to describe the relationship between pressure and density
+of a gas using the Ideal Gas Law.
+
+# Keywords
+- `sound_speed`            : Artificial speed of sound.
+- `reference_density`      : Reference density of the fluid.
+- `gamma`                  : Heat-capacity ratio
+- `background_pressure=0.0`: Background pressure.
```
efaulhaber, line 64: Can you please add `clip_negative_pressure` here in the list as well? And it seems that is also missing in the SE by Cole.

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
```diff
@@ -49,3 +49,50 @@ function inverse_state_equation(state_equation::StateEquationCole, pressure)
 
     return reference_density * tmp^(1 / exponent)
 end
+
+@doc raw"""
+    StateEquationIdealGas(;sound_speed, reference_density, gamma, background_pressure=0.0,
```
efaulhaber, line 54:
```suggestion
    StateEquationIdealGas(; sound_speed, reference_density, gamma, background_pressure=0.0,
```

efaulhaber (changes requested):

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
LasNikas: Alright, thanks for the explanation.

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
```diff
@@ -10,6 +10,7 @@ of water up to high pressures.
 - `reference_density`:       Reference density of the fluid.
 - `exponent`:                A value of `7` is usually used for most simulations.
 - `background_pressure=0.0`: Background pressure.
+- `clip_negative_pressure=false`: Negative pressure values are clipped to 0 which is necessary for `SummationDensity`.
```
efaulhaber, line 13:
```suggestion
- `clip_negative_pressure=false`: Negative pressure values are clipped to 0, which prevents spurious surface tension with `SummationDensity` but allows unphysical rarefaction of the fluid.
```

efaulhaber (changes requested):

src/schemes/fluid/weakly_compressible_sph/state_equations.jl
```diff
@@ -10,7 +10,7 @@ of water up to high pressures.
 - `reference_density`:       Reference density of the fluid.
 - `exponent`:                A value of `7` is usually used for most simulations.
 - `background_pressure=0.0`: Background pressure.
-- `clip_negative_pressure=false`: Negative pressure values are clipped to 0 which is necessary for `SummationDensity`.
+- `clip_negative_pressure=false`: Negative pressure values are clipped to 0, which prevents spurious surface tension with `SummationDensity` but allows unphysical rarefaction of the fluid.
```
efaulhaber, line 13: And for the other state equation as well, please.

efaulhaber (changes requested):

efaulhaber (approved):

efaulhaber:
Blocked by https://github.com/trixi-framework/TrixiParticles.jl/pull/669.

PR #610, opened 2024-08-22
Remove unsafe bitsets

svchb:
As in #594


test/preprocessing/geometries/geometries.jl
```diff
@@ -137,6 +137,9 @@
             end
         end
 
-        @test TrixiParticles.unique_sorted(copy(x)) == sort(unique(x))
+        sorted_unique_x = sort(unique(x))
+        sorted_unique_copy_x = TrixiParticles.unique_sorted(copy(x))
+
+        @test isapprox(sorted_unique_copy_x, sorted_unique_x; atol=eps())
```
efaulhaber, line 143: I don't understand why this is necessary. Can you please explain?

`unique_sorted` works with an absolute tolerance of `1e-14`. If we have two distinct elements within this tolerance, one will be removed by `unique_sorted`, but not by `unique`. This test here only works when all values of `x` (before adding duplicates) are not within `1e-14` of each other, which is apparently the case with the seed above.
If there are elements that are almost identical, the test would also fail with `isapprox`, as the number of elements is not correct.

efaulhaber (dismissed):

test/preprocessing/geometries/geometries.jl
svchb: Can't remember anymore could be the case that I tried this first before fixing it by removing the unsafe bitsets.

efaulhaber: Can you please try to remove it? I think it's a bit misleading, suggesting that this test would work with approximately equal numbers.

src/general/buffer.jl
```diff
@@ -4,7 +4,8 @@ struct SystemBuffer{V}
     buffer_size     :: Int
 
     function SystemBuffer(active_size, buffer_size::Integer)
-        active_particle = vcat(trues(active_size), falses(buffer_size))
+        # We cannot use a `BitVector` here, as writing to a `BitVector` is not thread-safe
+        active_particle = vcat(fill(true, active_size), fill(false, buffer_size))
```
LasNikas, line 8: please change `active_particle :: BitVector` to `active_particle :: Vector{Bool}` above

LasNikas (changes requested):

LasNikas (approved):

PR #614, opened 2024-09-16
Add YouTube video of JuliaCon talk to README

efaulhaber:


svchb:
Why don't you use the embed video option?
Otherwise I think it would be more obvious this is a link to a video if you added a play button on the image.

efaulhaber:
> Why don't you use the embed video option?
I googled, and apparently, one can only embed a video **file**, but we want to embed the official JuliaCon YouTube video.
Please tell me if you know a way to embed the YouTube video?

> Otherwise I think it would be more obvious this is a link to a video if you added a play button on the image.
That's a good idea!

svchb:
> > Why don't you use the embed video option?
> > I googled, and apparently, one can only embed a video **file**, but we want to embed the official JuliaCon YouTube video.
> > Please tell me if you know a way to embed the YouTube video?
NVM Ah I thought it would work.


> > Otherwise I think it would be more obvious this is a link to a video if you added a play button on the image.
> > That's a good idea!

svchb (approved):

PR #617, opened 2024-09-22
[docs] Point to downloads page to install julia

giordano:
I don't think https://julialang.org/downloads/platform/ is a particularly user-friendly page to point people to, it's more for those who for some reasons can't use [juliaup](https://github.com/JuliaLang/juliaup), which is nowadays the generally recommended method.

### DIFF
--- a/docs/src/install.md
+++ b/docs/src/install.md
@@ -1,8 +1,8 @@
 # [Installation](@id installation)
 
 ## Setting up Julia
-If you have not yet installed Julia, please [follow the instructions for your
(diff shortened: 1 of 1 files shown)


svchb (approved):
Thanks that is a good point.

giordano:
Uhm, sounds like this will never be merged as things stand, because the "run tests" jobs are mandatory to pass but they'll never run: https://github.com/trixi-framework/TrixiParticles.jl/blob/33f6f8683aa30e7b16ce6d2538c8929c177e7841/.github/workflows/ci.yml#L17-L27

Issue #629, opened 2024-10-21
[Docs] Getting started page claims a plot is automatically diplayed, but it isn't


giordano:
The example at https://trixi-framework.github.io/TrixiParticles.jl/v0.2.2/getting_started/#Running-an-Example claims that a plot is automatically displayed, but that's not the case, and I don't see how that'd happen, since there's nothing about plotting in the code at https://github.com/trixi-framework/TrixiParticles.jl/blob/b51cd2664b02d4dc98600dd8aeadf61a6c66e125/examples/fluid/hydrostatic_water_column_2d.jl.

Reproducer:
```julia
using Pkg
Pkg.add(["TrixiParticles", "OrdinaryDiffEq", "Plots"])
using TrixiParticles
trixi_include(joinpath(examples_dir(), "fluid", "hydrostatic_water_column_2d.jl"))
```

efaulhaber:
Thanks for catching this! The `plot(sol)` got lost somewhere.

LasNikas:
Thanks again for catching this. We'll fix it ASAP.
The bug is in the order of the parametes: https://github.com/trixi-framework/TrixiParticles.jl/blob/b51cd2664b02d4dc98600dd8aeadf61a6c66e125/src/visualization/recipes_plots.jl#L2

PR #632, opened 2024-10-22
Fix plotting of `ODESolution`

efaulhaber:
This is caused by a different order of parameters in newer versions of OrdinaryDiffEq.
We now restrict SciMLBase to the versions with this order of parameters.

See #629.


Project.toml
```diff
@@ -47,7 +47,7 @@ PointNeighbors = "0.4.2"
 Polyester = "0.7.10"
 RecipesBase = "1"
 Reexport = "1"
-SciMLBase = "1, 2"
+SciMLBase = "2.45"
```
svchb, line 50:
```suggestion
SciMLBase = "2.45.1"
```

svchb (changes requested):

svchb (approved):

LasNikas:
It would be nice to have tests in case the order of parameter types changes again in the future.

PR #649, opened 2024-11-07
Resize semi v2

LasNikas:


src/general/resize.jl
```diff
@@ -0,0 +1,206 @@
...
+    system.cache.additional_capacity[] = 0
+    system.cache.values_conserved[] = false
+
+    return system
+end
+
+resize_density!(system, n::Int, ::SummationDensity) = resize!(system.cache.density, n)
+resize_density!(system, n::Int, ::ContinuityDensity) = system
+
+function resize_cache!(system::WeaklyCompressibleSPHSystem, n::Int)
```
svchb, line 119: This is missing the resizing for all optional models?

src/general/resize.jl
```diff
@@ -0,0 +1,206 @@
...
+    # TODO
+    # resize!(system.cache.smoothing_length, n)
+
+    return system
+end
+
+function resize_cache!(system::EntropicallyDampedSPHSystem, n::Int)
```
svchb, line 126: This is missing the resizing for all optional models?

src/general/resize.jl
```diff
@@ -0,0 +1,206 @@
...
+            set_particle_density!(v, system, particle, density_keep)
+
+            for dim in 1:ndims(system)
+                v[dim, particle] = vel_keep[dim]
+                u[dim, particle] = pos_keep[dim]
+            end
+
+            delete_counter += 1
+        end
+    end
```
svchb, line 192: Please implement a copy function for each system so that optional models that have per particle values can be handled.

src/general/resize.jl
```diff
@@ -0,0 +1,206 @@
...
+
+    resize_density!(system, capacity_system, density_calculator)
+
+    resize_cache!(system, capacity_system)
+
+    system.cache.additional_capacity[] = 0
+    system.cache.values_conserved[] = false
+
+    return system
+end
```
svchb, line 114: Move common parts to a helper function

svchb (changes requested):

src/general/resize.jl
```diff
@@ -0,0 +1,206 @@
+# TODO: Make sure that the following functions are only called from a callback
```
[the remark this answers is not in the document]
LasNikas: Because we modify the integrator array, which is only allowed in a callback where you can call `u_modified!(integrator, true)`

src/general/resize.jl
LasNikas: What do you mean with "optional models"?

src/general/resize.jl
```diff
@@ -0,0 +1,206 @@
...
+
+    return system
+end
+
+resize_density!(system, n::Int, ::SummationDensity) = resize!(system.cache.density, n)
+resize_density!(system, n::Int, ::ContinuityDensity) = system
+
+function resize_cache!(system::WeaklyCompressibleSPHSystem, n::Int)
+    # TODO
+    # resize!(system.cache.smoothing_length, n)
```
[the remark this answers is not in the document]
LasNikas: This is for adaptive particle refinement

Issue #448, opened 2024-03-08
Storage for docs images


efaulhaber:
Like #438, but for docs images.

> Since I am not a fan of storing binary blobs in our repo, I create this issue to collect such logos. Feel free to add new ones at the bottom as a reply to this (permanently closed) issue, but with only a single logo per post please. If an existing logo is deprecated and to be replaced by another one, please mark it **boldly** in the original post that a new logo has been added and link to the new post, but _keep the old one_ such that existing links continue to work.

Posted by @sloede in https://github.com/trixi-framework/Trixi.jl/issues/1390.

efaulhaber:
Plotted with 200 dpi. That seems high enough to be perfectly sharp in the docs, but still has a reasonably small file size.
```
https://github.com/trixi-framework/TrixiParticles.jl/assets/44124897/95821154-577d-4323-ba57-16ef02ea24e0
```
![default_example_plot](https://github.com/trixi-framework/TrixiParticles.jl/assets/44124897/95821154-577d-4323-ba57-16ef02ea24e0)

svchb:
`https://github.com/trixi-framework/TrixiParticles.jl/assets/10238714/8466dbed-b1b1-4fce-ab2d-8b93b5cda833`
![osc_beam 0247](https://github.com/trixi-framework/TrixiParticles.jl/assets/10238714/8466dbed-b1b1-4fce-ab2d-8b93b5cda833)

svchb:
'https://github.com/trixi-framework/TrixiParticles.jl/assets/10238714/e05ace63-e330-441a-a391-eda3d2764074`
![osc_beam 0247](https://github.com/trixi-framework/TrixiParticles.jl/assets/10238714/e05ace63-e330-441a-a391-eda3d2764074)

svchb:
`https://github.com/trixi-framework/TrixiParticles.jl/assets/10238714/d7c6eedd-7173-4879-b62e-3e6d4bc5091f`
![296170691-bb5283d8-07fe-4fe7-b7f1-2c00fd166ee8](https://github.com/trixi-framework/TrixiParticles.jl/assets/10238714/d7c6eedd-7173-4879-b62e-3e6d4bc5091f)

svchb:
`https://github.com/trixi-framework/TrixiParticles.jl/assets/10238714/4c4c537a-8585-4456-9a1b-8b5adbf68533`
[...]

PR #659, opened 2024-11-11
Fix image path

### LINKED ISSUE #654: Images do not render in deployed docs
Issue related to revision https://github.com/openjournals/joss-reviews/issues/7044

The images linked in the visu part of the docs:

https://github.com/trixi-framework/TrixiParticles.jl/blob/550b95b5c088841b3bda2232f6e1b9310bae46bd/docs/src/visualization.md?plain=1#L19

and

https://github.com/trixi-framework/TrixiParticles.jl/blob/550b95b5c088841b3bda2232f6e1b9310bae46bd/docs/src/visualization.md?plain=1#L23

does not render in the deployed docs.

svchb:
Fix #654

### DIFF
--- a/docs/src/visualization.md
+++ b/docs/src/visualization.md
@@ -16,11 +16,11 @@ Follow these steps to view the exported VTK files in ParaView:
 5. Hold the left mouse button to move the solution around.
 
 You will now see the following:
-![image](https://github.com/svchb/TrixiParticles.jl/assets/10238714/45c90fd2-984b-4eee-b130-e691cefb33ab)
+![image](https://github.com/user-attachments/assets/108a55c5-888c-457c-8308-eb4a533d0833)
 
 To now view the result variables **first** make sure you have "fluid_1.pvd" highlighted in the "Pipeline Browser" then select them in the variable selection combo box (see picture below).
 Let's, for example, pick "density". To now view the time progression of the result hit the "play button" (see picture below).
-![image](https://github.com/svchb/TrixiParticles.jl/assets/10238714/7565a13f-9532-4a69-9f81-e79505400b1c)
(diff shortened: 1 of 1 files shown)


docs/src/visualization.md
```diff
@@ -16,11 +16,11 @@ Follow these steps to view the exported VTK files in ParaView:
 5. Hold the left mouse button to move the solution around.
 
 You will now see the following:
-![image](https://github.com/svchb/TrixiParticles.jl/assets/10238714/45c90fd2-984b-4eee-b130-e691cefb33ab)
+![image](https://github.com/user-attachments/assets/108a55c5-888c-457c-8308-eb4a533d0833)
```
efaulhaber, line 19: I find this confusing. There is no axis grid and particles are all the same color when first opening ParaView.

efaulhaber (changes requested):

efaulhaber:
I resetted my ParaView preferences and took some screenshots of how it looks like when a user first opens ParaView:
https://github.com/trixi-framework/TrixiParticles.jl/issues/448#issuecomment-2470091183

I suggest the following:
```
It is useful to make the particles larger.
For this, **first** make sure you have "fluid_1.pvd" highlighted in the "Pipeline Browser" then in the "Properties" window in the bottom left change "Point Size" to a larger value.
```
and then add my second screenshot above.
It would be great if you could add the yellow arrows in my first screenshot.

docs/src/visualization.md
svchb: These pictures are not new. They have just not been accessible since we switched from a private repo to a public one.

svchb: Yes we did but it would be better if you would say then when we are at it lets replace it with these better images.
Note: I cannot change it at all since it is themed through the OS.

efaulhaber (approved):

Issue #627, opened 2024-10-17
Improve documentation


efaulhaber:
This issue is to collect feedback that I got from students regarding the documentation.
I will continue this issue in the future with more feedback from more students/users.

- [x] `using Pkg; Pkg.add("OrdinaryDiffEq", "Plots")` doesn't work.
- [ ] Add images of shapes to the docs (rectangle, sphere, etc.).
- [ ] Add docs on how to visualize initial conditions for easier simulation setups.
- [ ] Continue the method theory page and make it easier for users to find out which methods and corrections, density diffusion, boundary conditions are best for which use case.
- [ ] Add more "developing TrixiParticles" information like best practices and conventions (format, comments, etc.).
- [x] Make clear that API reference is inside the components docs (#660).
- [ ] Clarify update steps. Why is `compute_pressure` separate from `compute_quantities`?
- [ ] Mention in the docs that `test_util.jl` has to be included before running an individual test file.
- [ ] Mention that density of fluids is stored either in `v` or in `cache.density`.

PR #660, opened 2024-11-12
Add docs for GPU support

efaulhaber:
I also added an example test for the hydrostatic water column with the `FullGridCellList`, which tests the code that I added in the docs.
And I addressed point 7 of #627.

Find the rendered docs for this PR here:
https://trixi-framework.github.io/TrixiParticles.jl/previews/PR660/


docs/src/gpu.md
```diff
@@ -0,0 +1,57 @@
+# GPU Support
+
+We are currently actively working on GPU support.
```
svchb, line 3:
```suggestion
GPU support is still an experimental feature that is actively being worked on.
```

docs/src/gpu.md
```diff
@@ -0,0 +1,57 @@
...
+As of now, the [`WeaklyCompressibleSPHSystem`](@ref) and the [`BoundarySPHSystem`](@ref)
+are supported on GPUs.
+We have tested this on GPUs by Nvidia and AMD.
+
+To run a simulation on a GPU, we need to use the [`FullGridCellList`](@ref)
+as cell list inside the [`GridNeighborhoodSearch`](@ref).
```
svchb, line 9:
```suggestion
To run a simulation on a GPU, we need to use the [`FullGridCellList`](@ref)
as cell list for the used [`GridNeighborhoodSearch`](@ref).
```

docs/src/gpu.md
```diff
@@ -0,0 +1,57 @@
...
+to obtain the bounding box as follows.
+```jldoctest gpu; output=false, setup=:(using TrixiParticles; trixi_include(@__MODULE__, joinpath(examples_dir(), "fluid", "hydrostatic_water_column_2d.jl"), sol=nothing))
+search_radius = TrixiParticles.compact_support(smoothing_kernel, smoothing_length)
+min_corner = minimum(tank.boundary.coordinates, dims=2) .- search_radius
+max_corner = maximum(tank.boundary.coordinates, dims=2) .+ search_radius
+cell_list = TrixiParticles.PointNeighbors.FullGridCellList(; min_corner, max_corner)
+
+# output
+PointNeighbors.FullGridCellList{PointNeighbors.DynamicVectorOfVectors{Int32, Matrix{Int32}, Vector{Int32}, Base.RefValue{Int32}}, Nothing, SVector{2, Float64}, SVector{2, Float64}}(Vector{Int32}[], nothing, [-0.24500000000000002, -0.24500000000000002], [1.245, 1.245])
+```
```
svchb, line 22: Can't we default to setting this bounding box?

docs/src/gpu.md
```diff
@@ -0,0 +1,57 @@
...
+│ ══════════════════                                                                               │
+│ #spatial dimensions: ………………………… 2                                                                │
+│ #systems: ……………………………………………………… 2                                                                │
+│ neighborhood search: ………………………… GridNeighborhoodSearch                                           │
+│ total #particles: ………………………………… 636                                                              │
+└──────────────────────────────────────────────────────────────────────────────────────────────────┘
+```
+
+At this point, we should run the simulation and make sure that it still works and that
+the bounding box is large enough.
```
svchb, line 42: Can you give an example when the bounding box would not be large enough?

docs/src/gpu.md
```diff
@@ -0,0 +1,57 @@
...
+ode = semidiscretize(semi, tspan, data_type=CuArray)
+```
+On an AMD GPU, we use:
+```julia
+using AMDGPU
+ode = semidiscretize(semi, tspan, data_type=ROCArray)
+```
+Then, we can run the simulation as usual.
+All data is transferred to the GPU during initialization and all loop over particles
+and neighbors are launched as GPU kernels.
```
svchb, line 56:
```suggestion
All data is transferred to the GPU during initialization and all loops over particles
and neighbors will be executed on the GPU.
```

docs/src/gpu.md
```diff
@@ -0,0 +1,57 @@
...
+Data is only copied to the CPU for saving VTK files with the [`SolutionSavingCallback`](@ref).
```
svchb, line 57:
```suggestion
Data is only copied to the CPU for saving VTK files via the [`SolutionSavingCallback`](@ref).
```

svchb, line 57: Probably postprocessing callback as well?

docs/src/overview.md
```diff
@@ -1,10 +1,15 @@
 # Overview
+The actual API reference is not listed on a single page, like in most Julia packages,
+but instead is split into multiple sections that follow a similar structure
+as the code files themselves.
+In these sections, API docs are integrated into theoretical background of the methods.
```
svchb, line 5:
```suggestion
In these sections, API docs are combined with explanations of the theoretical background of these methods.
```

svchb (changes requested):

docs/src/gpu.md
efaulhaber: Not really. What if we don't use a tank? Or if the tank doesn't have a closed top? 
I think it makes more sense to work on cell lists that support unbounded domains.

svchb: I am not saying to remove the option to overwrite the default bounding box. But using the minimum/maximum as a default seems to me to be a reasonable choice.

docs/src/gpu.md
svchb: What I mean is to improve the wording so that you mention cases where a larger bounding box should be used like an open tank etc.

docs/src/gpu.md
efaulhaber: I think it's clearer for people if we mention GPU kernels. Or they will be asking "how is it executed on the GPU? Don't we need a kernel?"

docs/src/gpu.md
efaulhaber: I think that's not working yet.

docs/src/gpu.md
efaulhaber: The problem is that this is an option of the neighborhood search. The neighborhood search for fluid-fluid only sees the coordinates of the fluid, so we can't set the bounding box to the maximum and minimum of the boundary coordinates.

docs/src/gpu.md
svchb: How about this than:
All data is transferred to the GPU during initialization, and computations looping over particles and their neighbors are processed directly on the GPU, utilizing GPU kernels generated via KernelAbstractions.jl.

docs/src/gpu.md
svchb: Yes it would need to be set in semidiscretization I guess.

efaulhaber: Discussion in https://github.com/trixi-framework/TrixiParticles.jl/issues/665.

docs/src/gpu.md
```diff
@@ -40,6 +40,10 @@ semi = Semidiscretization(fluid_system, boundary_system,
 
 At this point, we should run the simulation and make sure that it still works and that
 the bounding box is large enough.
+For some simulations where particles move outside the initial tank coordiantes,
```
svchb, line 43: coordinates

svchb (changes requested):

svchb (approved):

LasNikas (approved):

PR #662, opened 2024-11-12
Make simulations run with `Float32`

efaulhaber:
With this PR, replacing all `Float64` literals in a simulation file will run the full simulation on `Float32` only. This is much faster on consumer GPUs.


src/general/smoothing_kernels.jl
```diff
@@ -3,7 +3,8 @@ abstract type SmoothingKernel{NDIMS} end
 @inline Base.ndims(::SmoothingKernel{NDIMS}) where {NDIMS} = NDIMS
 
 @inline function kernel_grad(kernel, pos_diff, distance, h)
-    distance < sqrt(eps()) && return zero(pos_diff)
+    # Use `eps` relative to `h` and with the type of `h` to allow for `Float32`
+    distance < sqrt(eps(h)) && return zero(pos_diff)
```
svchb, line 7: Shouldn't this be also typeof(h)?

src/general/smoothing_kernels.jl
```diff
@@ -400,7 +401,7 @@ end
 end
 
 @inline normalization_factor(::WendlandC2Kernel{2}, h) = 7 / (pi * h^2)
-@inline normalization_factor(::WendlandC2Kernel{3}, h) = 21 / (2pi * h^3)
+@inline normalization_factor(::WendlandC2Kernel{3}, h) = 21 / (pi * h^3 * 2)
```
svchb, line 404: Can you please keep the multiplication in front of pi this is written like this everywhere. I am not sure why this was changed to remove the '*'...

src/general/smoothing_kernels.jl
```diff
@@ -461,7 +462,7 @@ end
 end
 
 @inline normalization_factor(::WendlandC4Kernel{2}, h) = 9 / (pi * h^2)
-@inline normalization_factor(::WendlandC4Kernel{3}, h) = 495 / (32pi * h^3)
+@inline normalization_factor(::WendlandC4Kernel{3}, h) = 495 / (pi * h^3 * 32)
```
svchb, line 465: see above

svchb (changes requested):

src/general/smoothing_kernels.jl
efaulhaber: See the comment. This is using `eps` relative to `h`. So when scaling a problem down by a factor of a million, this will also scale down the `eps` accordingly.

src/general/smoothing_kernels.jl
efaulhaber:
```julia
julia> typeof(2 * pi * 1f0)
Float64

julia> typeof(pi * 1f0 * 2)
Float32
```

src/general/smoothing_kernels.jl
svchb: But is this what we want? This will result in different cutoffs with variable smoothing_length.
```
julia> eps(Float32)
1.1920929f-7

julia> sqrt(eps(Float32(0.1)))
8.6316744f-5

sqrt(eps(Float32(100)))
0.0027621358f0
```

src/general/smoothing_kernels.jl
svchb: Than add a comment please... because this is stupid

src/general/smoothing_kernels.jl
LasNikas: If you have particle refinement in mind, then we restrict the variable smoothing length to 5...20% in the neighborhood. I think scaling `eps` is reasonable then.

efaulhaber: > But is this what we want? This will result in different cutoffs with variable smoothing_length.

Well, this is eps. So the rounding error for `distance2` will be in the order of `eps(smoothing_length)`. When two larger particles interact, the rounding error will be larger because everything is scaled, so it's natural that we need a larger tolerance.

svchb (dismissed):

svchb (dismissed):

svchb (dismissed):

svchb (approved):

PR #664, opened 2024-11-13
Test docs on win

svchb:
Fix #638 
Fix the remaining path issue

PR #609, opened 2024-08-22
Fixes for Open Boundary Systems for free surfaces

svchb:
Depends on #610 
Depends on #608
Depends on #606 
Depends on #605
Depends on #603 
Depends on #599

PR #605, opened 2024-08-20
Remove soundspeed from openboundary system

svchb:
Saving sound speed in the openboundary system makes no sense especially in the context of multiple fluids that can be at the openboundary.
As such remove sound speed and reuse the information from the fluid system(s).

Extracted from #604

PR #666, opened 2024-11-14
Wetting model for CSF Surface Tension Models

svchb:
Depends on #664 
Depends on #498
Depends on #539 
Depends on #604 
Depends on #605 
Depends on #606 
Depends on #607
Depends on #608
Depends on #609
Depends on #599 
Depends on #584


svchb:
has been reimplemented

PR #670, opened 2024-11-20
Another try on #669

svchb:


src/callbacks/solution_saving.jl
```diff
@@ -209,99 +209,61 @@ function (finalize::SolutionSavingCallback)(c, u, t, integrator)
...
-            "save initial solution" => solution_saving.save_initial_solution ?
-                                       "yes" : "no",
-            "save final solution" => solution_saving.save_final_solution ? "yes" :
-                                     "no",
-            "output directory" => abspath(solution_saving.output_directory),
-            "prefix" => solution_saving.prefix
-        ]
-        summary_box(io, "SolutionSavingCallback", setup)
+        # Fallback to default show method for AbstractCallback
+        invoke(Base.show, Tuple{IO, AbstractCallback}, io, cb)
```
efaulhaber: What if this is dispatched elsewhere? Then you overwrite this to instead use the generic function for `AbstractCallback`. Even if we don't do this in TrixiParticles, just loading TrixiParticles with another package that dispatches on a subtype of `DiscreteCallback` would break this other package.

src/callbacks/solution_saving.jl
```diff
@@ -209,99 +209,75 @@ function (finalize::SolutionSavingCallback)(c, u, t, integrator)
...
-    print(io, "SolutionSavingCallback(save_times=", solution_saving.save_times, ")")
-end
-
-function Base.show(io::IO, ::MIME"text/plain",
-                   cb::DiscreteCallback{<:Any, <:SolutionSavingCallback})
-    @nospecialize cb # reduce precompilation time
-
-    if get(io, :compact, false)
-        show(io, cb)
+function Base.show(io::IO, cb::DiscreteCallback)
```
efaulhaber, line 212: In general, this might break other functionality, even of other packages, as core functionality of SciMLBase is overwritten.

efaulhaber: We never want to do type piracy, unless we have a very good reason to do so. In main and #669, all dispatches contain our custom type somewhere, so it's not type piracy.

svchb:
Closed for #669

PR #680, opened 2024-12-04
Improved normal calculation methods

svchb:
Based on #666


svchb:
Closing as superseded by the current surface-tension and surface-normal stacks.
Boundary normals are now provided through InitialCondition.normals from #1036, which the new WettedAreaContactAngle model requires directly. The old wall_tangential helper is unnecessary because tangential directions are derived from the fluid and boundary normals.

PR #684, opened 2024-12-12
Use Asciicast.jl to display console output

efaulhaber:
Depends on https://github.com/ericphanson/Asciicast.jl/pull/35.

Find the rendered docs for this PR here:
https://trixi-framework.github.io/TrixiParticles.jl/previews/PR684/


LasNikas (dismissed):
Nice ! Thanks a lot

LasNikas:
I'm wondering if you can set the autoplay to false?

efaulhaber:
I don't think that's possible without another PR at Asciicast.jl. But do we even want to do that? I feel like most users would not notice that it's a video and not click on it.
Alternatively, how about we make it a loop?

LasNikas:
loop sounds good

svchb (dismissed):

efaulhaber:
Okay, so while I like the loop, it doesn't make sense if we're not able to read the timer output because the next loop iteration starts immediately. I reverted the change for now and created an issue here: https://github.com/ericphanson/Asciicast.jl/issues/38

ericphanson:
this is very cool btw! I think this kind of iterative-solver-output is a great use-case for Asciicast, where the thing you want to show the user isn't a nice plot/graphic (which is better done by emitting rich HTML), nor something static (which can be done more simply with just an `@example` or `@repl` block), but rather something dynamic in-terminal.

My motivating use-case for writing Asciicast was very similar, I wanted to show off the intermediate solutions generated during the solve by [TravelingSalesmanExact.jl](https://ericphanson.github.io/TravelingSalesmanExact.jl/dev/#Example), my toy package for TSP problems.

efaulhaber:
Alright, @ericphanson solved this problem in https://github.com/ericphanson/Asciicast.jl/pull/39, and we now have enough time to read the timer output. Ready for re-review @LasNikas @svchb.

LasNikas (approved):

PR #693, opened 2025-01-03
Implement the time integration method used in DualSPHysics

efaulhaber:


Project.toml
```diff
@@ -31,6 +31,12 @@ TimerOutputs = "a759f4b9-e2f1-59dc-863e-4aeb61b1ea8f"
 TrixiBase = "9a0f1c46-06d5-4909-a5a3-ce25d3fa3284"
 WriteVTK = "64499a7a-5c06-52f2-abe2-ccb03c286192"
 
+[weakdeps]
+OrdinaryDiffEqCore = "bbf590c4-e513-4bbe-9b18-05decba2e5d8"
```
efaulhaber, line 35: Unfortunatly, 
```julia
using TrixiParticles, OrdinaryDiffEq
```
doesn't make this extension available. I have to explicitly add and import `OrdinaryDiffEqCore`.
How do I make this work with just `using OrdinaryDiffEq`?
In the extension, I need to import `OrdinaryDiffEqCore`.
@sloede ?

src/general/time_integration.jl
```diff
@@ -0,0 +1,13 @@
+# Time integration is handles by the package OrdinaryDiffEq.jl.
```
efaulhaber, line 1:
```suggestion
# Time integration is handled by the package OrdinaryDiffEq.jl.
```

Project.toml
sloede: Well, then you should have OrdinaryDiffEq as the trigger for the pkgext. I am sure you can finagle OrdinaryDiffEq from it, either by then doing
```julia
using OrdinaryDiffEq.OrdinaryDiffEqCore
```
or if this does not work, I am sure one of the dependencies of OrdinaryDiffEq will have the Core package available as a name.

efaulhaber:
Superseded by #716.

PR #698, opened 2025-01-14
[Proof of Concept] Very simple prototype for an MPI implementation

efaulhaber:
Required https://github.com/trixi-framework/PointNeighbors.jl/pull/92.

This is a very simple prototype for a first MPI implementation. We can run a "static" simulation on multiple ranks, by which I mean that particles are not moving out of their rank's domain partition. Sending particles from one rank to another is not implemented yet, neither on the TrixiParticles nor the PointNeighbors side.


sloede:
Very cool! What's the major challenge to implementing full simulations (i.e., with particles migrating across MPI subdomain boundaries)?

efaulhaber:
> What's the major challenge to implementing full simulations (i.e., with particles migrating across MPI subdomain boundaries)?

Implementing it and spending all the time to debug everything and to make it ready for a clean PR. We changed a lot of internals here.
And we will have to implement a monolithic NHS that manages the domain partitioning and then stores all cell lists for individual systems. Currently we have one NHS for each pair of systems.
Also, resizing of the systems when particles change ranks is required.

sloede:
> > What's the major challenge to implementing full simulations (i.e., with particles migrating across MPI subdomain boundaries)?
> 
> Implementing it and spending all the time to debug everything and to make it ready for a clean PR. We changed a lot of internals here. And we will have to implement a monolithic NHS that manages the domain partitioning and then stores all cell lists for individual systems. Currently we have one NHS for each pair of systems. Also, resizing of the systems when particles change ranks is required.

Thanks. There are some incoming master students, one or two of whom might be up for a challenge. I'll reach out to you if I found someone who is interested and then we can see if it is feasible or not.

efaulhaber:
I think it's feasible for a good student. I would say it's similar to what I did in my thesis. Although we would have to work closely with the student to figure out how to cleanly integrate this into our current code.

sloede:
> I think it's feasible for a good student. I would say it's similar to what I did in my thesis.

Thanks for your estimate!

> Although we would have to work closely with the student to figure out how to cleanly integrate this into our current code.

Absolutely, I agree.

PR #689, opened 2024-12-23
Make GPUs actually usable

efaulhaber:
Depends on https://github.com/trixi-framework/PointNeighbors.jl/pull/90 and https://github.com/trixi-framework/TrixiBase.jl/pull/35.

This PR contains the following changes:
- TLSPH (and therefore FSI examples) now works on GPUs.
- TLSPH now works with `Float32` (and therefore on Apple GPUs that don't support double precision).
- Plotting now works on the GPU.
- ~I added a new macro `trixi_include_changeprecision(Float32, example_file)` to conveniently change all `Float64` literals in an example file to `Float32`, which is enough to make the whole simulation run with single precision (that's how we designed the code).~ Moved to https://github.com/trixi-framework/TrixiBase.jl/pull/35.

Find the rendered docs for this PR here:
https://trixi-framework.github.io/TrixiParticles.jl/previews/PR689/

PR #723, opened 2025-02-20
Use JuliaGPU Buildkite CI for GPU tests

efaulhaber:
Based on #689.

I added a GPU test suite and configured Buildkite on the public JuliaGPU runners.
These resources are very limited, so we should not start a build on every single push. Instead, I added a GitHub action to trigger GPU CI by commenting `/run-gpu-tests` on a PR. About 30 seconds later, the Buildkite jobs should appear in the list of checks for the PR. Also, every push on main will trigger a GPU CI build.
Note that especially for AMD the lack of resources requires us to wait for about an hour for a free runner.

While writing the GPU tests, I also found the following problems (and fixed some of them):
- [x] Depending on the number of fluid and boundary particles and the number of CPU threads, we sometimes end up in the serial implementation of the Adami pressure extrapolation, which is then still executed in parallel on the GPU, but leads to nondeterministic behavior due to race conditions.
- [x] The `RectangularTank` setup does not work properly with `Float32`.
- [x] Pressure clipping in the state equations is hardcoded in `Float64`.
- [x] EDAC is not GPU-compatible.
- [x] EDAC doesn't work with `Float32`.
- [x] Moving boundaries are not GPU-compatible.
- [ ] The `SchoenbergQuarticSplineKernel` does not work on GPUs (https://github.com/JuliaGPU/CUDA.jl/issues/2681 and https://github.com/JuliaGPU/Metal.jl/issues/550).
- [ ] Some weird Heisenbug breaking the 3D dam break on Metal (https://github.com/JuliaGPU/Metal.jl/issues/549). But only on my machine, apparently.
- [x] The default value for `acceleration` in the WCSPH constructor is in `Float64`.
- [ ] The open boundary system is not GPU-compatible (https://github.com/trixi-framework/PointNeighbors.jl/issues/20).
- [ ] Monaghan-Kajtar BC don't work with `Float32` (https://github.com/JuliaGPU/Metal.jl/issues/552).
- [ ] Intel GPUs just report a segfault in the first example.


sloede:
With https://github.com/JuliaConcurrent/Atomix.jl/pull/52 merged, at least the oneAPI should be able to run now...

efaulhaber:
> With https://github.com/JuliaConcurrent/Atomix.jl/pull/52 merged, at least the oneAPI should be able to run now...

It's still sh*t and gives a segfault in the first test.

svchb (changes requested):
Highlight support for GPUs by adding it to readme.md

svchb (approved):

LasNikas (approved):

PR #728, opened 2025-02-25
Code to reproduce Heisenbug with Metal.jl

efaulhaber:
This PR is to reproduce the Heisenbug with Metal.jl, which makes certain simulations fail on Metal.
I reduced it to this minimal example, but unfortunately, it only triggers the bug when I replace the actual simulation code (the function `kick!`) with the call to `test1` and then run this simulation in `examples/test2.jl`. Maybe the GPU allocations there are necessary to trigger the bug?

To trigger this bug, I include this file a hundred times by copying this into the REPL:
```
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
include("examples/test2.jl");
```
These are 50 lines, so the file is included 50 times. 
I often have to run it several hundred times to trigger the bug, but then I often get it multiple times in a row.
When the bug is triggered, it prints something like this and then the output stops for 5 seconds:
```
┌ Error: 
│   dv[:, 1] =
│    4-element MtlVector{Float32, Metal.PrivateStorage}:
│     123.78716
│     123.78716
│     123.78716
│       0.0
└ @ TrixiParticles ~/git/TrixiParticles.jl/src/general/semidiscretization.jl:457
```
This should be almost zero, and certainly not 123.78716.

### DIFF
--- a/examples/fluid/dam_break_3d.jl
+++ b/examples/fluid/dam_break_3d.jl
@@ -57,7 +57,7 @@ boundary_system = BoundarySPHSystem(tank.boundary, boundary_model)
 # ==========================================================================================
 # ==== Simulation
 semi = Semidiscretization(fluid_system, boundary_system)
-ode = semidiscretize(semi, tspan)
+ode = semidiscretize(semi, tspan, data_type=nothing)
 
 info_callback = InfoCallback(interval=10)
 saving_callback = SolutionSavingCallback(dt=0.02, prefix="")
--- a/examples/test2.jl
+++ b/examples/test2.jl
@@ -1,0 +1,23 @@
+using Metal, TrixiParticles
+
+# Import variables into scope
+trixi_include_changeprecision(Float32, @__MODULE__,
+joinpath(examples_dir(), "fluid",
+         "dam_break_3d.jl"),
+fluid_particle_spacing=0.1,
+sol=nothing, ode=nothing)
+
+# Neighborhood search with `FullGridCellList` for GPU compatibility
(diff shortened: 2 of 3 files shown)


efaulhaber:
Cannot reproduce anymore.

PR #732, opened 2025-03-11
Rework particle packing 

LasNikas:


svchb:
@LasNikas The test coverage is too low this should also cover these functions:
+ function volume(polygon::Polygon)
+ function volume(mesh::TriangleMesh)

efaulhaber:
Tests are failing.

src/preprocessing/geometries/triangle_mesh.jl
```diff
@@ -232,3 +242,11 @@ function unique_sorted(vertices)
 
     return vertices_sorted[keep]
 end
+
+function volume(mesh::TriangleMesh)
```
efaulhaber, line 246: Source? Just add a wikipedia link or so.

src/preprocessing/particle_packing/nhs_faces.jl
```diff
@@ -24,6 +24,8 @@ function faces_in_cell(cell, neighborhood_search)
     return PointNeighbors.points_in_cell(cell, neighborhood_search)
 end
 
+@inline eachneighbor(coords, nhs::TrivialNeighborhoodSearch) = nhs.eachpoint
```
efaulhaber, line 27:
```suggestion
@inline eachneighbor(coords, nhs::TrivialNeighborhoodSearch) = PointNeighbors.eachneighbor(coords, nhs)
```
Don't use internals.

src/preprocessing/particle_packing/system.jl
```diff
@@ -29,90 +30,135 @@ For more information on the methods, see description below.
                               - `use_for_boundary_packing=true`
                               - `max_signed_distance=boundary_thickness`
                            See [`SignedDistanceField`](@ref).
+- `fixed_system`:          When set to `true`, the system remains static, meaning particles
+                           will not move and the `InitialCondition` will stay unchanged.
+                           This is useful when the system is packed together with another
+                           (non-fixed) `ParticlePackingSystem`.
+                           In this case, no `SignedDistanceField` is required for both
+                           the fixed and non-fixed system.
```
efaulhaber, line 38:
```suggestion
                           the fixed and non-fixed system (use `nothing` as signed distance field).
```

src/preprocessing/particle_packing/system.jl
```diff
@@ -29,90 +30,135 @@ For more information on the methods, see description below.
...
 - `signed_distance_field`: To constrain particles onto the surface, the information about
                            the signed distance from a particle to a face is required.
                            The precalculated signed distances will be interpolated
                            to each particle during the packing procedure.
 - `smoothing_kernel`:      Smoothing kernel to be used for this system.
                            See [Smoothing Kernels](@ref smoothing_kernel).
-- `smoothing_length`:      Smoothing length to be used for this system.
+- `smoothing_length`:      Smoothing length to be used for the gradient estimation.
                            See [Smoothing Kernels](@ref smoothing_kernel).
+- `smoothing_length_interpolation`: Smoothing length to be used for intrepolating the `SignedDistanceField` information.
```
efaulhaber, line 47:
```suggestion
- `smoothing_length_interpolation`: Smoothing length to be used for interpolating the `SignedDistanceField` information.
```

src/preprocessing/particle_packing/system.jl
```diff
@@ -29,90 +30,135 @@ For more information on the methods, see description below.
...
+    if ndims(smoothing_kernel) != NDIMS
+        throw(ArgumentError("smoothing kernel dimensionality must be $NDIMS for a $(NDIMS)D problem"))
+    end
+
+    # Create neighborhood search for `ParticlePackingSystem` with the positions
+    # of the `SignedDistanceField` as neighbors.
+    # This is an intern NHS and is thus not organized by `Semidiscretization`.
+    if isnothing(signed_distance_field)
+        nhs = nothing
+        @info "No `SignedDistanceField` provided. Particles will not be constraint onto a geoemtric surface."
```
efaulhaber, line 119:
```suggestion
        @info "No `SignedDistanceField` provided. Particles will not be constraint onto a geometric surface."
```

src/preprocessing/particle_packing/system.jl
```diff
@@ -151,14 +209,33 @@ end
 
 update_callback_used!(system::ParticlePackingSystem) = system.update_callback_used[] = true
 
-function write2vtk!(vtk, v, u, t, system::ParticlePackingSystem, semi; write_meta_data=true)
+function write2vtk!(vtk, v, u, t, system::ParticlePackingSystem; write_meta_data=true)
+    vtk["velocity"] = [current_velocity(system.initial_condition.velocity, system,
```
efaulhaber, line 213: Why do you return the initial velocity?

src/preprocessing/particle_packing/system.jl
```diff
@@ -169,21 +246,14 @@ function kinetic_energy(system::ParticlePackingSystem, v_ode, u_ode, semi, t)
 
     # If `each_moving_particle` is empty (no moving particles), return zero
     return sum(each_moving_particle(system), init=zero(eltype(system))) do particle
-        velocity = advection_velocity(v, system, particle)
+        velocity = current_velocity(system.initial_condition.velocity, system, particle)
```
efaulhaber, line 249: ?

src/preprocessing/particle_packing/system.jl
```diff
@@ -287,12 +361,15 @@ function constrain_particle!(u, system, particle, distance_signed, normal_vector
...
+# Skip for fixed systems
+@inline update_transport_velocity!(system::ParticlePackingSystem{<:Any, true}, v_ode, semi) = system
+
 # Update from `UpdateCallback` (between time steps)
 @inline function update_transport_velocity!(system::ParticlePackingSystem, v_ode, semi)
     v = wrap_v(v_ode, system, semi)
     @threaded semi for particle in each_moving_particle(system)
         for i in 1:ndims(system)
-            v[ndims(system) + i, particle] = v[i, particle]
+            system.initial_condition.velocity[i, particle] = v[i, particle]
```
efaulhaber, line 372: I see, so you misuse the initial condition velocity as advection velocity? Why?

src/setups/complex_shape.jl
```diff
@@ -70,7 +70,7 @@ function ComplexShape(geometry; particle_spacing, density,
     # This is most likely only useful for debugging. Note that this is not public API.
     if store_winding_number
         return (initial_condition=initial_condition, winding_numbers=winding_numbers,
```
efaulhaber, line 72:
```suggestion
        return (; initial_condition, winding_numbers, grid)
```

efaulhaber (changes requested):

src/preprocessing/geometries/polygon.jl
```diff
@@ -139,3 +139,13 @@ end
 end
 
 @inline face_normal(edge, geometry::Polygon) = geometry.edge_normals[edge]
+
```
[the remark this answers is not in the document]
LasNikas: I think it then makes the code more complicated. I can't dispatch for polygon. Thus,  it would require a separate function `polygon_area`. Also, it would require the edges, which I don't have in the constructor.

src/preprocessing/geometries/triangle_mesh.jl
```diff
@@ -129,9 +129,19 @@ struct TriangleMesh{NDIMS, ELTYPE}
         min_corner = SVector([minimum(v[i] for v in vertices) for i in 1:NDIMS]...)
         max_corner = SVector([maximum(v[i] for v in vertices) for i in 1:NDIMS]...)
 
+        for i in eachindex(edge_normals)
+            # Skip zero normals to avoid unstable packing.
```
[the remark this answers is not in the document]
LasNikas: Because I have the `isfinite` check in `incident_angle`

src/preprocessing/particle_packing/system.jl
LasNikas: I don't actually need a momentum velocity (`v[1:ndims]`), since it is set to zero anyway.
Thus, I only need a place where I can store the advection (respective transport) velocity to evolve the particle position.
Since we extract the `InitialCondition` from the final solution, I don't think this is really "missusing", no?

src/preprocessing/particle_packing/system.jl
LasNikas: see https://github.com/trixi-framework/TrixiParticles.jl/pull/732#discussion_r2068355970

examples/preprocessing/packing_2d.jl
```diff
@@ -87,3 +91,9 @@ packed_boundary_ic = InitialCondition(sol, boundary_system, semi)
 
 trixi2vtk(packed_ic, filename="initial_condition_packed")
 trixi2vtk(packed_boundary_ic, filename="initial_condition_boundary_packed")
+
+shape = Shape(stack(geometry.vertices)[1, :], stack(geometry.vertices)[2, :])
```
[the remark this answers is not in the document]
efaulhaber: That's confusing. Please make it `Plots.Shape`.

src/preprocessing/particle_packing/system.jl
efaulhaber: It is. As a user, I would expect this to work:
```julia
ic = ComplexShape(...)
ic2 = pack(ic)

trixi_include("mysimulation.jl", initial_condition=ic, prefix="unpacked")
trixi_include("mysimulation.jl", initial_condition=ic2, prefix="packed")
```
I would not expect the simulation to screw with my initial condition.

src/preprocessing/geometries/polygon.jl
```diff
@@ -139,3 +139,16 @@ end
...
+function volume(polygon::Polygon)
```
svchb, line 143:
```suggestion
function area(polygon::Polygon)
```

LasNikas: Technically, I agree that "volume" is more commonly associated with 3D objects. However, there are two reasons why I chose to keep the function name `volume`:
- By using the same function name for both 2D and 3D cases, we can seamlessly handle both cases without needing to differentiate between dimensions in the calling code (in the source code or when using `trixi_include`).
- It's not public API

src/preprocessing/particle_packing/system.jl
```diff
@@ -29,90 +30,135 @@ For more information on the methods, see description below.
...
-                                   background_pressure, tlsph=true)
-        NDIMS = ndims(shape)
-        ELTYPE = eltype(shape)
-
-        particle_refinement = nothing
-
-        if ndims(smoothing_kernel) != NDIMS
-            throw(ArgumentError("smoothing kernel dimensionality must be $NDIMS for a $(NDIMS)D problem"))
-        end
+struct ParticlePackingSystem{S, F, NDIMS, ELTYPE <: Real, PR, C,
```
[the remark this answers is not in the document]
LasNikas: Do we? What about the TLSPH system and so on? 
It would make the code more complex (since I can dispatch with the first parameter typ) without any functional benefit.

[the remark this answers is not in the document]
LasNikas: If it's a fix system or not. The ndims dispatch is with the abstract system anyway.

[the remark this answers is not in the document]
svchb: This just a style/consistency question.
@efaulhaber what is your opinion about this?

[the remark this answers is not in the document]
LasNikas: Making it consistent would require subsequent changes (TLPSH, Open boundaries, boundary system, ...) and more discussion about other structs where we have NDIMS, ELTYPE etc....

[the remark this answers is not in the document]
efaulhaber: We usually put `NDIMS, ELTYPE` first for easy dispatch, but in cases like TLSPH and boundaries where we often wanted to dispatch on the boundary model, we changed it to put the boundary model in front to avoid a lot of these extra functions:
```julia
f(system::BoundarySPHSystem) = f(system, system.boundary_model)
f(system, ::BoundaryModelDummyParticles) = ...
```

src/preprocessing/particle_packing/system.jl
```diff
@@ -246,7 +254,7 @@ function kinetic_energy(system::ParticlePackingSystem, v_ode, u_ode, semi, t)
 
     # If `each_moving_particle` is empty (no moving particles), return zero
     return sum(each_moving_particle(system), init=zero(eltype(system))) do particle
-        velocity = current_velocity(system.initial_condition.velocity, system, particle)
+        velocity = current_velocity(system.advection_velocity, system, particle)
```
efaulhaber, line 257: It's confusing to misuse `current_velocity` like this. Please create a separate wrapper function
```julia
@inline advection_velocity(system::PackingSystem, particle) = extract_svector(system.advection_velocity, system, particle)
```

efaulhaber (changes requested):

src/preprocessing/particle_packing/system.jl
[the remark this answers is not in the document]
svchb: @efaulhaber And what is your opinion in this case?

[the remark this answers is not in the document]
efaulhaber: I'm fine with switching the order if that simplifies dispatch. That was also the reason to put `NDIMS` first most of the time, to simplify dispatch on `NDIMS`.

efaulhaber (approved):

svchb (approved):

PR #735, opened 2025-03-15
Implement the Particle Shifting Technique (PST)

efaulhaber:
Implements the Particle Shifting Technique by Sun et al. 2017 ([10.1016/j.cma.2016.10.028](https://www.sciencedirect.com/science/article/pii/S0045782516309112)).

grafik

Note that the strength of the shifting is directly scaled by the smoothing length.
Low Mach numbers (high speed of sound) stabilize the simulation, but high Mach numbers make the simulation very sensitive to the scaling factor for the shifting.
For Mach 0.2, we get this with the Wendland C2 kernel:
grafik

For the Schoenberg Cubic Spline kernel, everything looks alright:
grafik

Reducing the shifting by a factor of 2 yields this for the Wendland:
grafik

and this for the Cubic Spline Kernel:
grafik

So it is evident that our scaling for the Wendland kernels is incorrect and should be changed to be half of the current smoothing length.

Note that this is without shifting:
grafik


efaulhaber:
It doesn't really make sense to add a test at this point. I will add a vortex street example later, which will use PST.

LasNikas (changes requested):

LasNikas (approved):

svchb (approved):

PR #737, opened 2025-03-24
write2json

marcelschurer:
This PR adds the new function `trixi2json`, which exports simulation metadata to a dedicated JSON file at the start of the simulation. Previously, metadata was stored within the VTK output files. The structure and design of `write2json.jl` closely follow the existing implementation in `write2vtk.jl` for consistency.

Remark: Tests for the new functionality will be added in a follow-up commit.

### DIFF
--- a/src/callbacks/solution_saving.jl
+++ b/src/callbacks/solution_saving.jl
@@ -133,6 +133,11 @@ function initialize_save_cb!(solution_callback::SolutionSavingCallback, u, t, in
     solution_callback.latest_saved_iter = -1
     solution_callback.git_hash[] = compute_git_hash()
 
+    # Write metadata to JSON-file
+    if integrator.iter == 0
+        trixi2json(solution_callback, integrator)
+    end
+
     # Save initial solution
     if solution_callback.save_initial_solution
         # Update systems to compute quantities like density and pressure
--- a/src/callbacks/solution_saving.jl
+++ b/src/callbacks/solution_saving.jl
@@ -2,8 +2,7 @@
     SolutionSavingCallback(; interval::Integer=0, dt=0.0, save_times=Array{Float64, 1}([]),
                            save_initial_solution=true, save_final_solution=true,
                            output_directory="out", append_timestamp=false, prefix="",
-                           verbose=false, write_meta_data=true, max_coordinates=2^15,
-                           custom_quantities...)
+                           verbose=false, max_coordinates=2^15, custom_quantities...)
 
 
 Callback to save the current numerical solution in VTK format in regular intervals.
@@ -28,7 +27,6 @@ To ignore a custom quantity for a specific system, return `nothing`.
 - `append_timestamp=false`:     Append current timestamp to the output directory.
 - 'prefix=""':                  Prefix added to the filename.
 - `custom_quantities...`:       Additional user-defined quantities.
-- `write_meta_data=true`:       Write meta data.
 - `verbose=false`:              Print to standard IO when a file is written.
 - `max_coordinates=2^15`:       The coordinates of particles will be clipped if their
                                 absolute values exceed this threshold.
@@ -70,7 +68,6 @@ mutable struct SolutionSavingCallback{I, CQ}
     save_times            :: Vector{Float64}
     save_initial_solution :: Bool
     save_final_solution   :: Bool
-    write_meta_data       :: Bool
     verbose               :: Bool
     output_directory      :: String
     prefix                :: String
@@ -84,8 +81,8 @@ function SolutionSavingCallback(; interval::Integer=0, dt=0.0,
                                 save_times=Float64[],
                                 save_initial_solution=true, save_final_solution=true,
                                 output_directory="out", append_timestamp=false,
-                                prefix="", verbose=false, write_meta_data=true,
-                                max_coordinates=Float64(2^15), custom_quantities...)
+                                prefix="", verbose=false, max_coordinates=Float64(2^15),
+                                custom_quantities...)
     if (dt > 0 && interval > 0) || (length(save_times) > 0 && (dt > 0 || interval > 0))
         throw(ArgumentError("Setting multiple save times for the same solution " *
                             "callback is not possible. Use either `dt`, `interval` or `save_times`."))
@@ -101,8 +98,8 @@ function SolutionSavingCallback(; interval::Integer=0, dt=0.0,
 
     solution_callback = SolutionSavingCallback(interval, Float64.(save_times),
                                                save_initial_solution, save_final_solution,
-                                               write_meta_data, verbose, output_directory,
-                                               prefix, max_coordinates, custom_quantities,
+                                               verbose, output_directory, prefix,
+                                               max_coordinates, custom_quantities,
                                                -1, Ref("UnknownVersion"))
 
     if length(save_times) > 0
@@ -162,8 +159,8 @@ end
 
 # `affect!`
 function (solution_callback::SolutionSavingCallback)(integrator)
-    (; interval, output_directory, custom_quantities, write_meta_data, git_hash,
-    verbose, prefix, latest_saved_iter, max_coordinates) = solution_callback
+    (; interval, output_directory, custom_quantities, git_hash, verbose,
+    prefix, latest_saved_iter, max_coordinates) = solution_callback
 
     vu_ode = integrator.u
     semi = integrator.p
@@ -186,8 +183,8 @@ function (solution_callback::SolutionSavingCallback)(integrator)
 
     @trixi_timeit timer() "save solution" trixi2vtk(vu_ode, semi, integrator.t;
                                                     iter, output_directory, prefix,
-                                                    write_meta_data, git_hash=git_hash[],
-                                                    max_coordinates, custom_quantities...)
+                                                    git_hash=git_hash[], max_coordinates,
+                                                    custom_quantities...)
 
     # Tell OrdinaryDiffEq that `u` has not been modified
     u_modified!(integrator, false)
--- a/examples/fluid/falling_water_spheres_2d.jl
+++ b/examples/fluid/falling_water_spheres_2d.jl
@@ -87,8 +87,7 @@ semi = Semidiscretization(sphere_surface_tension, sphere, boundary_system)
 ode = semidiscretize(semi, tspan)
 
 info_callback = InfoCallback(interval=50)
-saving_callback = SolutionSavingCallback(dt=0.01, output_directory="out",
-                                         prefix="", write_meta_data=true)
+saving_callback = SolutionSavingCallback(dt=0.01, output_directory="out", prefix="")
 
 callbacks = CallbackSet(info_callback, saving_callback)
 
--- a/examples/fsi/falling_sphere_3d.jl
+++ b/examples/fsi/falling_sphere_3d.jl
@@ -10,5 +10,4 @@ trixi_include(@__MODULE__,
               solid_smoothing_kernel=WendlandC2Kernel{3}(),
               sphere_type=RoundSphere(),
               output_directory="out", prefix="",
-              write_meta_data=false, # Files with meta data can't be read by meshio
               tspan=(0.0, 1.0), abstol=1e-6, reltol=1e-3)
--- a/examples/fsi/falling_spheres_2d.jl
+++ b/examples/fsi/falling_spheres_2d.jl
@@ -117,8 +117,7 @@ semi = Semidiscretization(fluid_system, boundary_system, solid_system_1, solid_s
 ode = semidiscretize(semi, tspan)
 
 info_callback = InfoCallback(interval=10)
-saving_callback = SolutionSavingCallback(dt=0.02, output_directory="out", prefix="",
-                                         write_meta_data=true)
+saving_callback = SolutionSavingCallback(dt=0.02, output_directory="out", prefix="")
 
 callbacks = CallbackSet(info_callback, saving_callback)
 
--- a/examples/n_body/n_body_system.jl
+++ b/examples/n_body/n_body_system.jl
@@ -110,7 +110,7 @@ end
 
 TrixiParticles.vtkname(system::NBodySystem) = "n-body"
 
-function TrixiParticles.write2vtk!(vtk, v, u, t, system::NBodySystem; write_meta_data=true)
+function TrixiParticles.write2vtk!(vtk, v, u, t, system::NBodySystem)
     (; mass) = system
 
     vtk["velocity"] = v
--- a/src/TrixiParticles.jl
+++ b/src/TrixiParticles.jl
@@ -52,6 +52,7 @@ include("schemes/schemes.jl")
 include("general/semidiscretization.jl")
 include("general/gpu.jl")
 include("visualization/write2vtk.jl")
+include("visualization/write2json.jl")
 include("visualization/recipes_plots.jl")
 include("preprocessing/preprocessing.jl")
 
@@ -77,7 +78,7 @@ export BoundaryModelMonaghanKajtar, BoundaryModelDummyParticles, AdamiPressureEx
 
 export BoundaryMovement
 export examples_dir, validation_dir
-export trixi2vtk
+export trixi2vtk, trixi2json
 export RectangularTank, RectangularShape, SphereShape, ComplexShape
 export ParticlePackingSystem, SignedDistanceField
 export WindingNumberHormann, WindingNumberJacobson
--- a/src/preprocessing/particle_packing/system.jl
+++ b/src/preprocessing/particle_packing/system.jl
@@ -142,10 +142,8 @@ end
 
 update_callback_used!(system::ParticlePackingSystem) = system.update_callback_used[] = true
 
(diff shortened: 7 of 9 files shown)


src/io/io.jl
```diff
@@ -1,2 +1,141 @@
...
+function write_meta_data(system; system_name, output_directory, prefix, git_hash)
+    mkpath(output_directory)
+
+    meta_data = Dict{String, Any}(
+        "solver_name" => "TrixiParticles.jl",
+        "solver_version" => git_hash,
+        "julia_version" => string(VERSION)
+    )
+
+    get_meta_data!(meta_data, system)
```
LasNikas, line 33:
```suggestion
    add_meta_data!(meta_data, system)
```
IMO this is a better function name

src/callbacks/post_process.jl
```diff
@@ -216,6 +216,7 @@ function initialize_postprocess_callback!(cb::PostprocessCallback, u, t, integra
     # Apply the callback
     cb(integrator)
 
+    write_meta_data(cb, integrator)
```
LasNikas, line 219:
```suggestion
    write_meta_data(cb, integrator)
    
```

LasNikas (approved):

efaulhaber (changes requested):

src/io/io.jl
```diff
@@ -1,2 +1,147 @@
...
+
+function add_meta_data!(meta_data, system::FluidSystem)
+    meta_data["acceleration"] = system.acceleration
+    meta_data["viscosity"] = type2string(system.viscosity)
+    add_meta_data!(meta_data, system.viscosity)
+    meta_data["smoothing_kernel"] = type2string(system.smoothing_kernel)
+    meta_data["smoothing_length_factor"] = system.cache.smoothing_length_factor
+    meta_data["density_calculator"] = type2string(system.density_calculator)
+
+    if system isa WeaklyCompressibleSPHSystem
```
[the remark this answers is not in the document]
LasNikas: I think this is copy pasted from write2vtk. But I agree anyway

src/callbacks/post_process.jl
```diff
@@ -216,6 +216,8 @@ function initialize_postprocess_callback!(cb::PostprocessCallback, u, t, integra
...
```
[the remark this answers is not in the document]
LasNikas: > Do they both create the file and the second overwrites the first?
Yes

svchb:
@marcelschurer You need to fix the failing tests

src/io/io.jl
```diff
@@ -1,2 +1,265 @@
...
+
+    git_hash = callback.git_hash
+    prefix = hasproperty(callback, :prefix) ? callback.prefix : ""
+    semi = integrator.p
+    names = system_names(semi.systems)
+
+    info = Dict{String, Any}()
+    info["solver_name"] = "TrixiParticles.jl"
+    info["solver_version"] = git_hash[]
+    info["julia_version"] = string(VERSION)
```
LasNikas, line 17:
```suggestion
# Time integration
time_integrator = Dict(
    "Start time" => first(integrator.sol.prob.tspan),
    "Final time" => last(integrator.sol.prob.tspan),
    "time integrator" => integrator.alg |> typeof |> nameof,
    "adaptive" => integrator.opts.adaptive
)

if integrator.opts.adaptive
    time_integrator["abstol"] = integrator.opts.abstol
    time_integrator["reltol"] = integrator.opts.reltol
    time_integrator["controller"] = integrator.opts.controller
end

# Technical details
technical_setup = Dict(
    "parallelization backend" => semi.parallelization_backend |> typeof |> nameof,
    "#threads" => Threads.nthreads()
)
```

can you please add this to the file somehow/somewhere

src/io/io.jl
```diff
@@ -1,2 +1,265 @@
...
+    open(json_file, "w") do file
+        JSON.print(file, simulation_meta_data, 2)
+    end
+end
+
+add_meta_data!(meta_data, ::Nothing) = meta_data
+
+function add_meta_data!(meta_data, data)
+    ArgumentError("Unsupported data type: $(data). This data type is not implemented yet.")
+end
```
LasNikas, line 49:
```suggestion
function add_meta_data!(meta_data, data)
    throw(ArgumentError("Method for $(typeof(data)) not implemented. " *
                       "Please add a method `add_meta_data!(meta_data, ::$(typeof(data)))`."))
end
```

src/io/io.jl
```diff
@@ -1,2 +1,265 @@
...
+    mkpath(output_directory)
+
+    json_file = joinpath(output_directory, "simulation_meta_data.json")
+
+    open(json_file, "w") do file
+        JSON.print(file, simulation_meta_data, 2)
+    end
+end
+
+add_meta_data!(meta_data, ::Nothing) = meta_data
```
LasNikas, line 45:
```suggestion
# Skip metadata addition for `Nothing`
add_meta_data!(meta_data, data::Nothing) = meta_data
```

src/io/write_vtk.jl
```diff
@@ -388,6 +343,11 @@ function write2vtk!(vtk, v, u, t, system::TotalLagrangianSPHSystem; write_meta_d
                            initial_coords(system, particle)
                            for particle in eachparticle(system)]
 
+    vtk["lame_lambda"] = system.lame_lambda
+    vtk["lame_mu"] = system.lame_mu
+    vtk["young_modulus"] = system.young_modulus
+    vtk["poisson_ratio"] = system.poisson_ratio
```
svchb, line 349: why are these written to vtk?

src/io/io.jl
```diff
@@ -1,2 +1,265 @@
...
+
+    simulation_meta_data = Dict{String, Any}()
+    simulation_meta_data["info"] = info
+    simulation_meta_data["systems"] = systems
+
+    # write JSON-file
+    output_directory = callback.output_directory
+    mkpath(output_directory)
+
+    json_file = joinpath(output_directory, "simulation_meta_data.json")
```
svchb, line 38: Filename should be settable.

svchb (changes requested):

src/io/write_vtk.jl
LasNikas: Because these are per-particle values

LasNikas: Yes, in biomechanical applications

src/io/io.jl
```diff
@@ -1,2 +1,265 @@
...
+        systems[name] = meta_data
+    end
+
+    simulation_meta_data = Dict{String, Any}()
+    simulation_meta_data["info"] = info
+    simulation_meta_data["systems"] = systems
+
+    # write JSON-file
+    output_directory = callback.output_directory
+    mkpath(output_directory)
```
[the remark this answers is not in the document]
marcelschurer: But without `mkpath()`, it doesn't work: `SystemError: opening file "out\\simulation_meta_data.json": No such file or directory"`

src/callbacks/post_process.jl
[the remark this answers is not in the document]
marcelschurer: I changed the structure. The `PostprocessCallback` now calls the `add_system_data!` function in `io.jl` to write the data to its own file. This means that when both callbacks are called, each file contains the Systemdata, but this solves the problem of data overwriting and file naming.

src/io/io.jl
```diff
@@ -1,2 +1,311 @@
...
+    filename = callback.filename
+
+    meta_data = create_meta_data_dict(callback, integrator)
+
+    # write JSON-file
+    output_directory = callback.output_directory
+    mkpath(output_directory)
+    json_file = joinpath(output_directory,
+                         "meta_data" * add_opt_str_post(prefix) *
+                         add_opt_str_post(filename) * ".json")
```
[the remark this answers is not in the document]
LasNikas: as discussed on slack, I would suggest to use a static filename `meta_`

src/callbacks/solution_saving.jl
```diff
@@ -2,8 +2,7 @@
     SolutionSavingCallback(; interval::Integer=0, dt=0.0, save_times=Array{Float64, 1}([]),
                            save_initial_solution=true, save_final_solution=true,
                            output_directory="out", append_timestamp=false, prefix="",
-                           verbose=false, write_meta_data=true, max_coordinates=2^15,
```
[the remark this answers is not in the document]
efaulhaber: 1. Isn't the metadata an addition to other files? Like you get a CSV and additional metadata for that. Or a VTK and additional metadata. Would you ever need just a metadata file without any other data from a simulation?
2. I wouldn't want to add a new callback just for this. This would complicate simulation files unnecessarily.
3. I'm fine with always writing this. It doesn't affect the runtime and one extra file doesn't hurt. This is especially useful when one wants to restart from VTK. It would avoid situations where you forgot to turn metadata on, and then you have to run it again to be able to restart.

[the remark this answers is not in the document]
efaulhaber: I think the only reason why we made metadata optional was that it was crashing ParaView sometimes.

src/io/io.jl
```diff
@@ -1,2 +1,309 @@
...
+
+# handle "_" on optional prefix strings
+add_opt_str_post(str) = (str === "" ? "" : "_$(str)")
+
+function write_meta_data(callback::SolutionSavingCallback, integrator)
+    prefix = callback.prefix
+
+    meta_data = create_meta_data_dict(callback, integrator)
+
+    # write JSON-file
```
efaulhaber, line 12:
```suggestion
    # Write JSON file
```

src/io/io.jl
```diff
@@ -1,2 +1,309 @@
...
+    meta_data = Dict{String, Any}()
+
+    info = Dict{String, Any}()
+    add_simulation_info!(info, git_hash, integrator)
+    meta_data["simulation_info"] = info
+
+    systems = Dict{String, Any}()
+    foreach_system(semi) do system
+        idx = system_indices(system, semi)
+        name = prefix * add_opt_str_post("$(names[idx])")
```
efaulhaber, line 38: This is not correct. This function is for optional postfixes, not prefixes, which means you end up with `_fluid_1`.
Use this:
```suggestion
        name = add_opt_str_pre(prefix) * names[idx]
```
with
```julia
add_opt_str_pre(str) = (str === "" ? "" : "$(str)_")
```
It probably makes sense to rename this function to something more descriptive like `add_underscore_to_optional_prefix`.

src/io/io.jl
```diff
@@ -1,2 +1,309 @@
 include("write_vtk.jl")
 include("read_vtk.jl")
+
+# handle "_" on optional prefix strings
+add_opt_str_post(str) = (str === "" ? "" : "_$(str)")
```
efaulhaber, line 5:
```suggestion
# Handle "_" on optional prefix strings
add_underscore_to_optional_prefix(str) = (str === "" ? "" : "$(str)_")
# Same for optional postfix strings
add_underscore_to_optional_postfix(str) = (str === "" ? "" : "_$(str)")
```
Can you please remove these original functions from `trixi2vtk`, now that they are defined globally here?

src/io/io.jl
```diff
@@ -1,2 +1,309 @@
...
+function add_simulation_info!(info, git_hash, integrator)
+    info["solver_name"] = "TrixiParticles.jl"
+    info["solver_version"] = git_hash[]
+    info["julia_version"] = string(VERSION)
+
+    info["time_integrator"] = Dict{String, Any}()
+    info["time_integrator"]["integrator_type"] = type2string(integrator.alg)
+    info["time_integrator"]["start_time"] = first(integrator.sol.prob.tspan)
+    info["time_integrator"]["final_time"] = last(integrator.sol.prob.tspan)
+    info["time_integrator"]["adaptive"] = integrator.opts.adaptive
```
efaulhaber, line 59: `dt` is missing for fixed time step sizes.

src/io/io.jl
```diff
@@ -1,2 +1,309 @@
...
+        info["time_integrator"]["reltol"] = integrator.opts.reltol
+        info["time_integrator"]["controller"] = type2string(integrator.opts.controller)
+    end
+
+    info["technical_setup"] = Dict{String, Any}()
+    info["technical_setup"]["parallelization_backend"] = type2string(integrator.p.parallelization_backend)
+    info["technical_setup"]["#threads"] = Threads.nthreads()
+end
+
+# Skip systemdata addition for `Nothing`
```
efaulhaber, line 71:
```suggestion
```
Redundant comment

src/io/io.jl
```diff
@@ -1,2 +1,309 @@
...
+add_system_data!(system_data, data::Nothing) = system_data
+
+function add_system_data!(system_data, data)
+    throw(ArgumentError("Method for $(typeof(data)) not implemented. " *
+                        "Please add a method `add_system_data!(system_data, ::$(typeof(data)))`."))
+end
```
efaulhaber, line 77:
```suggestion
```
We usually don't do this in Julia, as we automatically get an error like this.

src/io/io.jl
```diff
@@ -1,2 +1,309 @@
...
+function add_system_data!(system_data, system::FluidSystem)
+    system_data["system_type"] = type2string(system)
+    system_data["particle_spacing"] = particle_spacing(system, 1)
+    system_data["density_calculator"] = type2string(system.density_calculator)
+    system_data["smoothing_kernel"] = type2string(system.smoothing_kernel)
+    system_data["smoothing_length"] = system.cache.smoothing_length
+    system_data["acceleration"] = system.acceleration
+    system_data["sound_speed"] = system_sound_speed(system)
+    system_data["pressure_acceleration_formulation"] = nameof(system.pressure_acceleration_formulation)
+    add_system_data!(system_data, system.shifting_technique)
```
efaulhaber, line 88:
```suggestion
    add_system_data!(system_data, shifting_technique(system))
```

efaulhaber (changes requested):

sloede:
Out of curiosity - what's the purpose of having JSON files? Is this for some post processing or just to make simulations "reproducible"? If the postprocessing is done by Julia, why not serialize the relevant TrixiP objects into a JLD2 file or something similar?

I'm just asking since before, JSON would also have been my go-to move, but given that it can never fully capture what we can do in Julia itself (e.g., it can never accurately represent an initial condition we only supply as a function), it is also somewhat limiting.

sloede:
Why not follow the Trixi.jl strategy for restarts:

Just store the elixir along with your results, and to restart, you replace the part that creates the ODE problem with some initial conditions by a call to something like `load_restart_file` that loads the simulation state from a restart file.

efaulhaber:
We had this discussion a while ago, but I don't remember the reason. @svchb?

I think the reasoning was the following. If you modified an example file to run a simulation, you would have to do the exact same modification again in order to load the particle data to do interpolation and/or restart the simulation. With the metadata approach, all you need to load/restart is the output data.

Maybe we need to discuss this again, @LasNikas?

sloede:
If it's really just about storing some meta data for further processing (which not necessarily takes place in Julia), then writing a JSON file seems reasonable - especially if it is *not* intended to be used as the mechanism for restarting simulations 👍

However, I think it's a IMHO a common misconception to think that JSON is a human-readable file format. Technically it is, yes, but it clearly wasn't designed for that purpose, and if data gets just a little bit more complex, *actual*  human readability gets shot in a second.

If you want human readability, why not go for JSON? For example, the `meta.json` from above would look like this, which IMHO is much nicer to parse as a human being (plus you can add comments!):
```toml
[simulation_info]
julia_version = "1.11.5"
solver_version = "v0.2.6-67-g64fd67424"
solver_name = "TrixiParticles.jl"

    [simulation_info.technical_setup]
    "#threads" = 8
    parallelization_backend = "PolyesterBackend"

    [simulation_info.time_integrator]
    start_time = 0.0
    reltol = 0.001
    integrator_type = "RDPK3SpFSAL35"
    adaptive = true
    final_time = 1.0
    controller = "PIDController"
    abstol = 1.0e-6

[system_data._fluid_1]
acceleration = [0.0, -9.81]
pressure_acceleration_formulation = "pressure_acceleration_continuity_density"
density_calculator = "ContinuityDensity"
particle_spacing = 0.05
system_type = "WeaklyCompressibleSPHSystem"
smoothing_length = 0.06
sound_speed = 10.0
smoothing_kernel = "SchoenbergCubicSplineKernel"

    [system_data._fluid_1.state_equation]
    model = "StateEquationCole"
    reference_density = 1000.0
    background_pressure = 0.0
    exponent = 7.0

    [system_data._fluid_1.viscosity_model]
    model = "ArtificialViscosityMonaghan"
    alpha = 0.02
    epsilon = 0.01
    beta = 0.0

[system_data._boundary_1]
particle_spacing = 0.05
system_type = "BoundarySPHSystem"
adhesion_coefficient = 0.0

    [system_data._boundary_1.boundary_model]
    density_calculator = "AdamiPressureExtrapolation"
    model = "BoundaryModelDummyParticles"
    smoothing_length = 0.06
    smoothing_kernel = "SchoenbergCubicSplineKernel"

        [system_data._boundary_1.boundary_model.state_equation]
        model = "StateEquationCole"
        reference_density = 1000.0
        background_pressure = 0.0
        exponent = 7.0
```

PR #759, opened 2025-04-09
Fix Inflow Boundary

svchb:


src/schemes/boundary/open_boundary/system.jl
```diff
@@ -346,6 +346,25 @@ end
     return fluid_system
 end
 
+# Fluid particle enters from the simulation domain into an inflow boundary so we reimpose the reference velocity.
```
efaulhaber, line 349:
```suggestion
# Fluid particle enters from the simulation domain into an inflow boundary so we reimpose the reference velocity
```

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -346,6 +346,25 @@ end
...
+                                   boundary_zone::BoundaryZone{InFlow}, particle,
+                                   v, u, v_fluid, u_fluid, t)
+    (; reference_velocity) = system
+
+    particle_coords = current_coords(u_fluid, fluid_system, particle)
+
+    v_particle = current_velocity(v_fluid, fluid_system, particle)
+    v_ref = reference_value(reference_velocity, v_particle, particle_coords, t)
+
+    # Update the particle’s velocity to match the prescribed inlet profile.
```
efaulhaber, line 360:
```suggestion
    # Update the particle’s velocity to match the prescribed inlet profile
```

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -346,6 +346,25 @@ end
...
+    @inbounds for dim in 1:ndims(system)
+        v_fluid[dim, particle] = v_ref[dim]
+    end
+
+    return fluid_system
+end
```
efaulhaber, line 366: Can you please explain why this is needed? And why is only the velocity updated but not the other values?

efaulhaber (changes requested):

svchb:
The followup of this is #764 but #764 is not working correctly

PR #762, opened 2025-04-10
Extrapolate density in `BoundaryModelTafuni`

LasNikas:
The current approach (on main) is that we calculate the density for the buffer particles with the inverse equation of state.
In this PR, I extrapolated the density (similar to Tafuni et al.).

I simulated a pipe-flow with a prescribed velocity profile (black line in diagram)
Next, I measured the velocity on the green line with a Shepard interpolation (from ParaView).
The red particles are buffer particles and the blue are fluid particles.

The blue line in the diagram shows the averaged measured velocity with the `main` branch and the red line with this PR.
The simulation runs longer with this PR. However, the simulation also reaches an instability at ~11 sec.

Is it worth to merge this PR? I mean the current approach is a bit more performant.

image

image


svchb:
fluid/pipe_flow_2d.jl - BoundaryModelTafuni (WCSPH)        |    1     2      3     23.6s
failed

efaulhaber:
> Next, I measured the velocity on the green line with a Shepard interpolation (from ParaView).
Can you please use our interpolation? I don't trust ParaView.

> I mean the current approach is a bit more performant.

Why is it less performant? I don't see any major differences that would significantly hurt performance.

> However, the simulation also reaches an instability at ~11 sec.

Why? Is this supposed to be stable? If so, there must be a bug somewhere, and it would make sense to sort this out first.

src/schemes/boundary/open_boundary/mirroring.jl
```diff
@@ -134,7 +139,10 @@ function extrapolate_values!(system, v_open_boundary, v_fluid, u_open_boundary,
             density[particle] = reference_value(reference_density, density[particle],
                                                 particle_coords, t)
         else
```
svchb, line 141:
```suggestion
        else if prescribed_pressure
           inverse_state_equation!(density, state_equation, pressure, particle)
        else
```

svchb (changes requested):
And
else if prescribed_density
          pressure[particle]= state_equation(density)
        else

PR #763, opened 2025-04-10
Improve plane interpolation performance

efaulhaber:
Depends on #755 and https://github.com/trixi-framework/PointNeighbors.jl/pull/105.

main:
```
apply postprocess cb                10     256s   88.5%   25.6s   89.5GiB   99.8%  8.95GiB
  ~apply postprocess cb~            10     256s   88.4%   25.6s   89.5GiB   99.8%  8.95GiB
  update systems                    10    159ms    0.1%  15.9ms   56.2KiB    0.0%  5.62KiB
```
#755:
```
apply postprocess cb                10    64.4s   62.3%   6.44s   89.3GiB   99.7%  8.93GiB
  ~apply postprocess cb~            10    64.2s   62.1%   6.42s   89.3GiB   99.7%  8.93GiB
  update systems                    10    167ms    0.2%  16.7ms   56.2KiB    0.0%  5.62KiB
```
this PR:
```
apply postprocess cb                10    2.22s    3.6%   222ms    305MiB   58.9%  30.5MiB
  write to vtk                      10    1.44s    2.3%   144ms   85.0MiB   16.4%  8.50MiB
  interpolate plane                 10    617ms    1.0%  61.7ms    220MiB   42.5%  22.0MiB
  update systems                    10    162ms    0.3%  16.2ms   56.2KiB    0.0%  5.62KiB
```


test/examples/examples.jl
```diff
@@ -1,89 +1,89 @@
...
+    # @testset verbose=true "Solid" begin
+    #     @trixi_testset "solid/oscillating_beam_2d.jl" begin
+    #         @test_nowarn_mod trixi_include(@__MODULE__,
+    #                                        joinpath(examples_dir(), "solid",
+    #                                                 "oscillating_beam_2d.jl"),
+    #                                        tspan=(0.0, 0.1))
+    #         @test sol.retcode == ReturnCode.Success
+    #         @test count_rhs_allocations(sol, semi) == 0
+    #     end
+    # end
```
LasNikas, line 15: ?

efaulhaber: Oops.

test/general/interpolation.jl
```diff
@@ -1,52 +1,51 @@
 @testset verbose=true "SPH Interpolation" begin
-    function compare_interpolation_result(actual, expected; tolerance=1e-8)
+    function compare_interpolation_result(actual, expected; tolerance=5e-4)
```
svchb, line 2: Is this really the tightest tolerance that works here?

test/general/interpolation.jl
```diff
@@ -1,52 +1,51 @@
...
+        return abs(outside - original_inside)
     end
 
     @testset verbose=true "2D" begin
         function set_values(coord)
             mass = 1.0
             density = 666
             pressure = 2 * coord[2]
-            velocity = [5 + 3 * coord[1], 0.1 * coord[2]^2 + 0.1]
+            velocity = [5+3 * coord[1]; 0.1 * coord[2]^2+0.1;;]
```
svchb, line 48:
```suggestion
            velocity = [5 + 3 * coord[1]; 0.1 * coord[2]^2+0.1;;]
```

svchb (changes requested):

src/general/interpolation.jl
```diff
@@ -436,227 +447,186 @@ results = interpolate_point(points, semi, ref_system, sol)
...
+            m_a = hydrodynamic_mass(neighbor_system, neighbor)
             W_a = kernel(ref_smoothing_kernel, distance, smoothing_length)
 
             if system_id == ref_id
-                interpolated_density += m_a * W_a
-                volume = m_a / current_density(v, system, particle)
-                shepard_coefficient += volume * W_a
+                interpolated_density[point] += m_a * W_a
+                volume = m_a / current_density(v, neighbor_system, neighbor)
+                shepard_coefficient[point] += volume * W_a
```
LasNikas, line 540: As discussed on Slack, this represents the computed density rather than the interpolated density. I would suggest using the interpolation approach described in the attached formula instead.
image

svchb: It should be noted:
1) all the tests have to be changed again
2) you can do this by returning the density in interpolate_system!(). All other values are interpolated in this way.

LasNikas: As a user, I would expect the function to return an interpolated density rather than a computed one, based on the function name and the documentation.

svchb: My arguments are only for doing this in an extra PR.

efaulhaber: https://github.com/trixi-framework/TrixiParticles.jl/issues/796

test/general/interpolation.jl
efaulhaber: I'm afraid I can't change that. Odd that the formatter decides to format like this.

test/general/interpolation.jl
svchb: I know I was just wondering if you actually tested this. If this is the case I think we don't need to store that many decimal places.

svchb (approved):
NEWS.md:

## Version 0.3.1

### Performance

- Improve interpolation performance by using (#763).

LasNikas (approved):

PR #768, opened 2025-04-11
Add shifting for `InFlow`-buffer particles

LasNikas:
I ran a simulation with a step in the transition zone (see image below). This step induced periodic gaps in the buffer zone which didn't decay. Indeed, the opposite is the case. The simulation is becoming unstable.

image

So the workaround is to shift not only the fluid particles but also the buffer particles.
I only shift particles that are in the vicinity of the transition plane. 

Left: No shifting of buffer particle. Right: Shifted buffer particles
image


src/callbacks/particle_shifting.jl
```diff
@@ -123,3 +122,93 @@ function Base.show(io::IO, ::MIME"text/plain",
...
+            particle_coords = current_coords(u, system, particle)
+
+            particle_position = particle_coords - zone_origin
+
+            # Distance to transition plane
+            dist = dot(particle_position, spanning_set[1])
+
+            if dist <= max_dist
+                for neighbor in PointNeighbors.eachneighbor(particle_coords,
+                                                            neighborhood_search)
```
efaulhaber, line 169: 1. This is not public API.
2. This is not available for every NHS implementation.

I see two options:
1. Use `foreach_point_neighbor` and skip all particle-neighbor pairs with particle outside the max distance (benchmark performance and compare with this function here).
2. If this is not fast enough, we can make `foreach_neighbor` public API.

efaulhaber (changes requested):
Can you just cut off the shifting at a point? Is this stable? Or should this be a smooth transition from shifting to no shifting?

src/callbacks/particle_shifting.jl
LasNikas: The first option is not an option, because their exists no NHS for `OpenBoundarySPHSystem`s.
Buffer particles don't need to interact with other particles or with itself. 
I like the second option

LasNikas:
> Can you just cut off the shifting at a point? Is this stable? Or should this be a smooth transition from shifting to no shifting?

Yes, because these buffer particles don't calculate their physical properties physically but extrapolate (Tafuni) them or determine them with a model (Lastiwka). Here, the shifting acts basically as a rearrangement of the position of the particles.

src/callbacks/particle_shifting.jl
efaulhaber: > Buffer particles don't need to interact with other particles or with itself.
You are just implementing an interaction between buffer particles and other particles.

You can't just use another NHS and assume it works for a different set of particles. This is what `PointNeighbors.requires_update` is for.

PR #769, opened 2025-04-14
Implement Tensile Instability Control (TIC)

efaulhaber:
This is an undocumented feature until we figure out when and if it is useful.
So far, it is not useful alone:
grafik

But it might be useful in addition to PST.

Note that, just like PST, this does not work at free surfaces.


efaulhaber:
The point is that we can experiment with it. For example, Niklas can try using PST + TIC for the high-res vortex street he's having problems with.

svchb:
But do we need to merge it than without documentation? He can try this like it is.

src/schemes/fluid/pressure_acceleration.jl
```diff
@@ -58,6 +58,14 @@ end
     return -volume_term * pressure_tilde * W_a
 end
 
+# Same as `pressure_acceleration_continuity_density`, but using the minus formulation
+# when pressures are negative to avoid tensile instability.
+# See Lyu et al. (2021). https://doi.org/10.1016/j.apor.2021.102938
+@inline function tensile_instability_correction(m_a, m_b, rho_a, rho_b, p_a, p_b, W_a)
+    p = ifelse(p_a >= 0, p_a + p_b, p_b - p_a)
+    return -m_b * p / (rho_a * rho_b) * W_a
```
svchb, line 66:
```suggestion
    p_sum_effective = abs(p_a) + abs(p_b)
    return -m_b * p_sum_effective / (rho_a * rho_b) * W_a
```

Your formulation is not correct I think since p_b can be negative as well.

svchb (changes requested):

src/schemes/fluid/pressure_acceleration.jl
```diff
@@ -58,6 +58,14 @@ end
     return -volume_term * pressure_tilde * W_a
 end
 
+# Same as `pressure_acceleration_continuity_density`, but using the minus formulation
+# when pressures are negative to avoid tensile instability.
+# See Lyu et al. (2021). https://doi.org/10.1016/j.apor.2021.102938
```
svchb, line 63:
```suggestion
# See Lyu et al. (2021). https://doi.org/10.1016/j.apor.2021.102938
# See Sun et al. (2018). https://doi.org/10.1016/j.cpc.2017.11.016
# Note: This violates the momentum conservation of a fluid system due to the loss of the symmetry of particle # interactions.
# Note: This should only be applied for the fluid-fluid interaction for non-free surface flows!
```

src/schemes/fluid/pressure_acceleration.jl
```diff
@@ -58,6 +58,14 @@ end
...
+@inline function tensile_instability_correction(m_a, m_b, rho_a, rho_b, p_a, p_b, W_a)
+    p = p_b + abs(p_a)
+    return -m_b * p / (rho_a * rho_b) * W_a
+end
+
```
svchb, line 68:
```suggestion
# Symmetric version of the concept introduced by Sun et al. (2018). https://doi.org/10.1016/j.cpc.2017.11.016 
# Note: This should only be applied for the fluid-fluid interaction for non-free surface flows!
@inline function tensile_instability_correction_sym(m_a, m_b, rho_a, rho_b, p_a, p_b, W_a)
    # Use absolute values of pressure so that negative pressures (tension)
    # also contribute repulsively, preventing unphysical clustering.
    p_abs = abs(p_a) + abs(p_b)

    return -m_b * p_abs / (rho_a * rho_b) * W_a
end
```

svchb (changes requested):

efaulhaber:
> The point is that we can experiment with it. For example, Niklas can try using PST + TIC for the high-res vortex street he's having problems with.

I experimented with this and it is indeed solving the high resolution issues. Top is PST and bottom PST + TIC, left and right are the same setup, just a different resolution.
https://github.com/user-attachments/assets/ffcde256-e7b6-4f24-8031-416bfeb01a3c

I will try @svchb's symmetric version, and then I'll make this public API and add docs and tests.

efaulhaber:
Update: The symmetric version (top panel) is not stable.

grafik

This is shortly before the symmetric version is crashing:
grafik

Here is the code again, so it doesn't get lost in the conversations:
```julia
@inline function tensile_instability_correction_sym(m_a, m_b, rho_a, rho_b, p_a, p_b, W_a)
    # Use absolute values of pressure so that negative pressures (tension)
    # also contribute repulsively, preventing unphysical clustering.
    p_abs = abs(p_a) + abs(p_b)
    return -m_b * p_abs / (rho_a * rho_b) * W_a
end
```

src/schemes/fluid/pressure_acceleration.jl
efaulhaber: Unstable (see comment below).

svchb:
Yes that was the likely result probably if its conservative its not allowing some loss in energy necessary for it to work.

efaulhaber:
> `# Note: This should only be applied for the fluid-fluid interaction for non-free surface flows!`

@svchb where did you get the information from that this is not supposed to be used for fluid-boundary interaction? I don't find any mention of this in any of the papers.

efaulhaber:
I just tried TIC only for fluid-fluid interaction and the regular (p_a + p_b) formulation for fluid-boundary interaction, and I got this. Top is no TIC for fluid-boundary, bottom is TIC for everything.
I would say that is clearly showing that we need TIC for both.
https://github.com/user-attachments/assets/20e494d6-9118-4f5d-9276-c0d21e4b2e8b

svchb:
So TIC should only be applied for dummy boundaries and MK bnd need to be excluded.

efaulhaber:
MK boundaries use a special "pressure acceleration" anyway, so we don't need to change anything.

docs/src/systems/weakly_compressible_sph.md
```diff
@@ -109,3 +109,88 @@ term is very expensive and adds about 40--50% of computational cost.
...
+- ``h`` is the smoothing length,
+- ``R`` and ``n`` are constants, which are set to ``0.2`` and ``4`` respectively,
+- ``W(\Delta x_a)`` is the smoothing kernel of the particle size of particle ``a``,
+  which can be interpreted as the target particle spacing that we want to achieve.
+- ``\nabla W_{ab}`` is the gradient of the smoothing kernel,
+- ``m_b`` is the mass of particle ``b``,
+- ``\rho_a, \rho_b`` is the density of particles ``a`` and ``b``, respectively.
+
+Note that we replaced ``\text{CFL} \cdot \text{Ma}`` by ``\Delta t \cdot v_\text{max} / h``,
+as explained on page 29, right above Equation 9.
```
LasNikas, line 150:
```suggestion
as explained in [Sun2018](@cite Sun2018) on page 29, right above Equation 9.
```

docs/src/systems/weakly_compressible_sph.md
```diff
@@ -109,3 +109,88 @@ term is very expensive and adds about 40--50% of computational cost.
...
+The Particle Shifting Technique (PST) is a technique to correct tensile instability
+in regions of low pressure, as observed in viscous flow around an object.
+Without PST, tensile instability causes non-physical separation of the fluid
+from the object, introducing void regions behind the object.
+
+At lower resolutions, PST alone can be effective to correct a viscous flow
+around a cylinder, as shown in this figure.
+![particle_shifting](https://github.com/user-attachments/assets/70892b0b-af36-4531-b328-f73e63a5e33c)
+At higher resolutions, PST alone is not effective anymore; see the figure in
+[Tensile Instability Control](@ref tic).
```
LasNikas, line 124: I’m not sure I fully understand the reason for this behavior. Is it related to a smaller compact support at higher resolution?
We also observe detachment from the boundary at coarser resolutions, which creates a free surface that PST cannot handle—correct?
Why does the problem actually get worse as the resolution increases?
Is the same void region also present at coarser resolution, but "compensated" for by the larger compact support?

src/schemes/fluid/weakly_compressible_sph/system.jl
```diff
@@ -30,6 +30,11 @@ See [Weakly Compressible SPH](@ref wcsph) for more details on the method.
                                 See [`ArtificialViscosityMonaghan`](@ref) or [`ViscosityAdami`](@ref).
 - `transport_velocity`:         [Transport Velocity Formulation (TVF)](@ref transport_velocity_formulation). Default is no TVF.
 - `density_diffusion`:          Density diffusion terms for this system. See [`DensityDiffusion`](@ref).
+- `pressure_acceleration`:      Pressure acceleration formulation for this system.
```
LasNikas, line 33: Please add also to the constructor above

docs/src/systems/weakly_compressible_sph.md
efaulhaber: The initial detachment is about the same size for both the lower and the higher resolution simulation. But for the lower resolution, the void region is much smaller compared to the particle size, so the shifting can close the hole again, whereas in the higher resolution it is too large to close. 

Another way to think about it is that in the high resolution simulation we have more particles where the velocity is essentially orthogonal to the boundary.

LasNikas (dismissed):

docs/src/systems/weakly_compressible_sph.md
```diff
@@ -109,3 +109,88 @@ term is very expensive and adds about 40--50% of computational cost.
 Modules = [TrixiParticles]
 Pages = [joinpath("schemes", "fluid", "weakly_compressible_sph", "density_diffusion.jl")]
 ```
+
+## [Particle Shifting Technique](@id shifting)
+
+The Particle Shifting Technique (PST) is a technique to correct tensile instability
```
svchb, line 115:
```suggestion
The Particle Shifting Technique (PST) is a method to correct tensile instability
```

docs/src/systems/weakly_compressible_sph.md
```diff
@@ -109,3 +109,88 @@ term is very expensive and adds about 40--50% of computational cost.
...
+Without PST, tensile instability causes non-physical separation of the fluid
+from the object, introducing void regions behind the object.
+
+At lower resolutions, PST alone can be effective to correct a viscous flow
+around a cylinder, as shown in this figure.
+![particle_shifting](https://github.com/user-attachments/assets/70892b0b-af36-4531-b328-f73e63a5e33c)
+At higher resolutions, PST alone is not effective anymore; see the figure in
+[Tensile Instability Control](@ref tic).
+We generally recommend using PST and Tensile Instability Control together
+in such a simulation.
```
svchb, line 126:
```suggestion
We recommend using PST and Tensile Instability Control together
in such simulations.
```

svchb (changes requested):

docs/src/systems/weakly_compressible_sph.md
efaulhaber: I tried to avoid the word method, as this is what we usually use for the solver itself, so it would be the WCSPH method with the PST and TIC techniques.

svchb: Just sounds weird to have technique twice so close repeated.

svchb (approved):

docs/src/systems/weakly_compressible_sph.md
efaulhaber: I asked ChatGPT.

efaulhaber:
```suggestion
The Particle Shifting Technique (PST) is used to correct tensile instability.
```

LasNikas:
\run-gpu-tests

svchb (approved):
/run-gpu-tests

PR #771, opened 2025-04-16
Open boundaries - NHS fix

LasNikas:


svchb (approved):
Erik should have a look.

PR #774, opened 2025-04-17
Add dispatch for Wendland and Shifting

LasNikas:
This is only an ugly workaround until we figure out how we scale the smoothing length correctly with the wendland kernels.


src/callbacks/particle_shifting.jl
```diff
@@ -64,6 +64,9 @@ function particle_shifting!(u, v, system::FluidSystem, v_ode, u_ode, semi,
     Wdx = smoothing_kernel(system, particle_spacing(system, 1), 1)
     h = smoothing_length(system, 1)
 
+    k_factor(::WendlandC2Kernel) = 2
```
svchb, line 67:
```suggestion
    k_factor(::WendlandKernel) = 2
```

svchb (changes requested):

svchb:
superseeded by #775

PR #776, opened 2025-04-17
Hotfix for unit tests

efaulhaber:
The latest PointNeighbors.jl release broke the tests.


svchb:
So i looked at the change I guess there is no version change needed?

PR #778, opened 2025-04-17
Update julia formatter

svchb:


svchb:
> We wanted to remove the option `whitespace_in_kwargs = false` at some point to be consistent with Trixi.jl.
> 
> Do we want to do that now that we have a formatting PR anyway?

I will create a new PR. I just checked there are many changes.

PR #780, opened 2025-04-25
Remove kwarg whitespace option

svchb:


svchb:
@efaulhaber should we make this part of dev branch?

PR #790, opened 2025-05-05
Add full support for parallelization backends

efaulhaber:
Based on https://github.com/trixi-framework/PointNeighbors.jl/pull/109.


src/general/semidiscretization.jl
```diff
@@ -282,16 +282,16 @@ function semidiscretize(semi, tspan; reset_threads=true)
...
 
-    if semi.parallelization_backend isa Bool
-        # Use CPU vectors and the optimized CPU code
-        u0_ode = Vector{ELTYPE}(undef, sum(sizes_u))
-        v0_ode = Vector{ELTYPE}(undef, sum(sizes_v))
-    else
+    if semi.parallelization_backend isa KernelAbstractions.Backend
         # Use the specified backend, e.g., `CUDABackend` or `MetalBackend`
         u0_ode = KernelAbstractions.allocate(semi.parallelization_backend, ELTYPE,
                                              sum(sizes_u))
```
LasNikas, line 288:
```suggestion
        u0_ode = allocate(semi.parallelization_backend, ELTYPE, sum(sizes_u))
```
with 
```julia
function allocate(backend::KernelAbstractions.Backend, ELTYPE, size)
    return KernelAbstractions.allocate(backend, ELTYPE, size)
end

function allocate(backend, ELTYPE, size)
    return Array{ELTYPE, length(size)}(undef, size)
end
```

Is this somehow possible to do the same for the `adapt` function?

efaulhaber: Yes, but this is only used here, so I did not create a new function for it. Once we need this elsewhere, I agree a function would be nice.

LasNikas (approved):

PR #794, opened 2025-05-07
Split fluid and TLSPH time integration to allow for TLSPH substeps

efaulhaber:


src/callbacks/split_integration.jl
```diff
@@ -0,0 +1,342 @@
...
+    kwargs     :: Any
+end
+
+@doc raw"""
+    SplitIntegrationCallback(alg; kwargs...)
+
+Callback to integrate the `TotalLagrangianSPHSystem`s in a `Semidiscretization`
+separately from the other systems.
+After each time step of the main integrator (in which TLSPH systems are ignored),
+the TLSPH systems are integrated for multiple substeps with their own integrator.
```
svchb, line 15:
```suggestion
the TLSPH systems are integrated for multiple smaller timesteps with their own integrator.
```

src/callbacks/split_integration.jl
```diff
@@ -0,0 +1,342 @@
...
+
+This is useful when two conditions are satisfied:
+1. The TLSPH systems require much smaller time steps than the fluid systems,
+   which is usually the case when very stiff materials are simulated.
+2. The amount of TLSPH particles is small compared to the number of fluid particles,
+   so that a fluid time step is much more expensive than a TLSPH substep.
```
svchb, line 21:
```suggestion
This is useful under the condition that the TLSPH systems require much smaller time steps than the fluid systems, which is usually the case when stiff materials are simulated.
```

The 2nd case is irrelevant as long as the fluid particles are not much fewer than the TLSPH particles which is not really a realistic case. Even if the TLSPH particles would be the same number as fluid particles the smaller timestep would dominate.

src/callbacks/split_integration.jl
```diff
@@ -0,0 +1,342 @@
...
+
+For fluid-structure interactions with very stiff materials like steel or carbon fiber
+composites, this can lead to significant speedups of several hundred times if the ratio
+of fluid to solid particles is large enough (e.g. 100:1 or more).
```
svchb, line 25:
```suggestion
For fluid-structure interactions with stiff materials like metals or carbon fiber
composites, this can lead to significant speedups of several hundred times if the ratio
of fluid to solid particles is large enough (e.g. 100:1 or more).
```

src/callbacks/split_integration.jl
```diff
@@ -0,0 +1,342 @@
...
+        v = wrap_v(v_ode, system, semi)
+        u = wrap_u(u_ode, system, semi)
+
+        update_pressure!(system, v, u, v_ode, u_ode, semi, t)
+    end
+
+    # No `update_boundary_interpolation!` for performance reasons, or we will lose
+    # a lot of the speedup that we can gain with split integration.
+    # We assume that the TLSPH particles move so little during the substeps
+    # that the extrapolated pressure/density values can be treated as constant.
```
svchb, line 237: Can you cheaply add an error/warning if the movement becomes too large?

svchb (changes requested):

src/callbacks/split_integration.jl
efaulhaber: Hmmm, not sure. The split integration adds a bit of overhead, so I don't know if it would be noticeably faster if the numbers of fluid and TLSPH particles are the same. Maybe we can rephrase this? It might be useful in this case, but it's the product of the two conditions that determines the speedup factor.

Maybe this?
```suggestion
This is useful if the TLSPH systems require much smaller time steps than the fluid systems,
which is usually the case when stiff materials are simulated.
It is especially useful if additionally the number of TLSPH particles is much smaller than the number of fluid particles.
```

src/callbacks/split_integration.jl
svchb: Yes I know it will probably crash. Just thought it might be interesting to see this information if it does.

svchb (approved):

src/callbacks/split_integration.jl
efaulhaber: But the question is, what is "too large"? And I think the answer is exactly "too large for the time integration".

src/callbacks/split_integration.jl
```diff
@@ -0,0 +1,342 @@
...
+
+This is useful if the TLSPH systems require much smaller time steps than the fluid systems,
+which is usually the case when stiff materials are simulated.
+It is especially useful if additionally the number of TLSPH particles is much smaller
+than the number of fluid particles, so that a fluid time step is much more expensive
+than a TLSPH substep.
+
+For fluid-structure interactions with stiff materials like metal or carbon fiber
+composites, this can lead to significant speedups of several hundred times if the ratio
+of fluid to solid particles is large enough (e.g. 100:1 or more).
```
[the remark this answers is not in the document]
efaulhaber: I would like to add the diagram in a new PR based on #917.

Issue #800, opened 2025-05-13
Reference Files missing for lid driven cavity validation case


svchb:

efaulhaber:
Now I understand what you mean. Still, in the CI nothing is validated at all.

PR #801, opened 2025-05-13
Fix Validation Setups

svchb:


validation/taylor_green_vortex_2d/validation_taylor_green_vortex_2d.jl
```diff
@@ -74,7 +81,7 @@ for density_calculator in density_calculators, perturbation in perturb_coordinat
     output_directory = joinpath("out_tgv", name_density_calculator * name_perturbation,
                                 wcsph ? "wcsph" : "edac",
                                 "validation_run_taylor_green_vortex_2d_nparticles_$(n_particles_xy)x$(n_particles_xy)")
-    saving_callback = SolutionSavingCallback(dt=0.02,
+    saving_callback = SolutionSavingCallback(dt=0.1,
```
LasNikas, line 84:
```suggestion
    saving_callback = SolutionSavingCallback(dt=1//10,
```
and also for the `PostprocessCallback`: `dt=2//100`.
This avoids approximation errors when to `affect` the periodic callback.

LasNikas (approved):

validation/lid_driven_cavity_2d/validation_lid_driven_cavity_2d.jl
```diff
@@ -9,9 +9,9 @@ particle_spacings = [0.02, 0.01, 0.005]
 tspan = (0.0, 25.0)
 reynolds_numbers = [100.0, 1000.0, 10_000.0]
 
-interpolated_velocity(v, u, t, system) = nothing
+interpolated_velocity(system, v, u, semi, t) = nothing
```
efaulhaber, line 12: Why did this work? Is this tested?

validation/taylor_green_vortex_2d/validation_taylor_green_vortex_2d.jl
```diff
@@ -2,7 +2,8 @@ using TrixiParticles
 
 # ==========================================================================================
 # ==== Resolution
-particle_spacings = [0.02, 0.01, 0.005]
+# particle_spacings = [0.02, 0.01, 0.005]
+particle_spacings = [0.02]
```
efaulhaber, line 6:
```suggestion
particle_spacing = 0.02
```
According to the discussion in another PR.

efaulhaber (changes requested):

validation/lid_driven_cavity_2d/validation_lid_driven_cavity_2d.jl
svchb: No it is not tested. I have not included this into the test since the setup has no output which can be directly used to verify the accuracy. For the TGV case I will add this in a follow-up PR.

efaulhaber: I don't understand. You added a test for TGV, which just tests that the simulation finishes. Can you please do the same for LDC? Otherwise, this validation file is completely untested. 

> the setup has no output which can be directly used to verify the accuracy
This is #800.

svchb: This is a very trivial PR to fix the validation setups. 
So I did not adjust anything in the TGV case. I will follow-up in another PR for TGV and add reference files etc.
I will not follow-up on the LDC case so I did not include it even as a basic "run" test.

efaulhaber: But you fixed the file, so there should be at least a test that is running this file to make sure that this doesn't happen again.

svchb: I wanted to save some time for the runs. Anyway I included the LDC setup with that we will get very close to 2h for the CI runs that's why I added #802.

validation/taylor_green_vortex_2d/validation_taylor_green_vortex_2d.jl
```diff
@@ -2,7 +2,8 @@ using TrixiParticles
 
 # ==========================================================================================
 # ==== Resolution
-particle_spacings = [0.02, 0.01, 0.005]
+# particle_spacings = [0.02, 0.01, 0.005]
+particle_spacings = 0.02
```
efaulhaber, line 6:
```suggestion
particle_spacing = 0.02
```
And no loop over `particle_spacings`.

test/validation/validation.jl
```diff
@@ -83,4 +83,27 @@
         # Verify number of plots
         @test length(axs_edac[1].scene.plots) >= 2
     end
+
+    @trixi_testset "TGV_2D" begin
+        @trixi_test_nowarn trixi_include(@__MODULE__,
+                                         joinpath(validation_dir(),
+                                                  "taylor_green_vortex_2d",
+                                                  "validation_taylor_green_vortex_2d.jl")) [
```
efaulhaber, line 91:
```suggestion
                                                  "validation_taylor_green_vortex_2d.jl", tspan=(0.0, 0.01)) [
```
We already test TGV in the examples. This test here is only to verify that there are no bugs in the validation files (for now there is no validation).

validation/lid_driven_cavity_2d/validation_lid_driven_cavity_2d.jl
```diff
@@ -9,9 +9,9 @@ particle_spacings = [0.02, 0.01, 0.005]
 tspan = (0.0, 25.0)
```
efaulhaber, line 9: Please also set
```julia
particle_spacing = 0.02
```
and remove the loop.

test/validation/validation.jl
```diff
@@ -83,4 +83,27 @@
...
+        ]
+        @test sol.retcode == ReturnCode.Success
+        @test count_rhs_allocations(sol, semi) == 0
+    end
+
+    @trixi_testset "LDC_2D" begin
+        @trixi_test_nowarn trixi_include(@__MODULE__,
+                                         joinpath(validation_dir(),
+                                                  "lid_driven_cavity_2d",
+                                                  "validation_lid_driven_cavity_2d.jl"))
```
efaulhaber, line 105:
```suggestion
                                                  "validation_lid_driven_cavity_2d.jl", tspan=(0.0, 0.01)))
```

efaulhaber (changes requested):

validation/lid_driven_cavity_2d/validation_lid_driven_cavity_2d.jl
```diff
@@ -2,17 +2,21 @@ using TrixiParticles
...
+# particle_spacings = [0.02, 0.01, 0.005]
+particle_spacing = 0.02
 
 # ==========================================================================================
 # ==== Experiment Setup
 tspan = (0.0, 25.0)
 reynolds_numbers = [100.0, 1000.0, 10_000.0]
 
+const sensor_capture_time = 24.8
+const capture_started = Ref(false)
```
efaulhaber, line 14:
```suggestion
const SENSOR_CAPTURE_TIME = 24.8
const CAPTURE_STARTED = Ref(false)
```

validation/lid_driven_cavity_2d/validation_lid_driven_cavity_2d.jl
```diff
@@ -2,17 +2,21 @@ using TrixiParticles
 
 # ==========================================================================================
 # ==== Resolution
-particle_spacings = [0.02, 0.01, 0.005]
+# particle_spacings = [0.02, 0.01, 0.005]
```
efaulhaber, line 5:
```suggestion
```

efaulhaber (changes requested):

validation/lid_driven_cavity_2d/validation_lid_driven_cavity_2d.jl
efaulhaber: How about this?
```suggestion
# The paper provides reference data for particle spacings 0.02, 0.01, 0.005
```

validation/lid_driven_cavity_2d/validation_lid_driven_cavity_2d.jl
```diff
@@ -50,14 +54,17 @@ function interpolated_velocity(system::TrixiParticles.FluidSystem, v, u, semi, t
...
     return nothing
 end
 
-for particle_spacing in particle_spacings, reynolds_number in reynolds_numbers,
-    density_calculator in [SummationDensity(), ContinuityDensity()], wcsph in [false, true]
-    n_particles_xy = round(Int, 1.0 / particle_spacing)
+for reynolds_number in reynolds_numbers,
+    density_calculator in [SummationDensity(), ContinuityDensity()], wcsph in [false, true],
+    #particle_spacing in particle_spacings
+
```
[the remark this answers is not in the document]
efaulhaber: I don't see the point of a loop here. I guess it would make more sense to `trixi_include` this file in a loop if one ever wants to run several resolutions.

efaulhaber (approved):

PR #802, opened 2025-05-13
Split unit, examples and validation into separate ci steps

svchb:


svchb:
Yes but with adding more validation tests the examples will run out of time and run forever.

PR #807, opened 2025-05-14
Fix error message when using TVF without UpdateCB

### LINKED ISSUE #767: No error is thrown when using TVF without `UpdateCallback`

svchb:
Fix #767


LasNikas:
we solved this for the open boundaries as follows: https://github.com/trixi-framework/TrixiParticles.jl/blob/3eb4996cedf8044d82d3eecde8950f2fbd6507a3/src/schemes/boundary/open_boundary/system.jl#L199

It would be nice if we also do so here

LasNikas (approved):

src/schemes/fluid/entropically_damped_sph/system.jl
```diff
@@ -306,6 +306,10 @@ function update_final!(system::EntropicallyDampedSPHSystem, v, u, v_ode, u_ode,
     compute_curvature!(system, surface_tension, v, u, v_ode, u_ode, semi, t)
     compute_stress_tensors!(system, surface_tension, v, u, v_ode, u_ode, semi, t)
     update_average_pressure!(system, system.transport_velocity, v_ode, u_ode, semi)
+
+    # checks for callback dependent models
```
efaulhaber, line 310:
```suggestion
    # Check that TVF is only used together with `UpdateCallback`
```

src/schemes/fluid/weakly_compressible_sph/system.jl
```diff
@@ -304,6 +304,10 @@ function update_final!(system::WeaklyCompressibleSPHSystem, v, u, v_ode, u_ode,
     # Surface normal of neighbor and boundary needs to have been calculated already
     compute_curvature!(system, surface_tension, v, u, v_ode, u_ode, semi, t)
     compute_stress_tensors!(system, surface_tension, v, u, v_ode, u_ode, semi, t)
+
+    # checks for callback dependent models
```
efaulhaber, line 308:
```suggestion
    # Check that TVF is only used together with `UpdateCallback`
```

src/schemes/fluid/transport_velocity.jl
```diff
@@ -165,19 +165,13 @@ function update_callback_used!(system, transport_velocity)
...
-                       v, u, v_ode, u_ode, semi, t; update_from_callback=false)
+function check_tvf_configuration(system::FluidSystem, ::Nothing,
+                                 v, u, v_ode, u_ode, semi, t; update_from_callback=false)
     return system
 end
 
-function update_final!(system::FluidSystem, ::TransportVelocityAdami,
-                       v, u, v_ode, u_ode, semi, t; update_from_callback=false)
+function check_tvf_configuration(system::FluidSystem, ::TransportVelocityAdami,
+                                 v, u, v_ode, u_ode, semi, t; update_from_callback=false)
```
efaulhaber, line 174:
```suggestion
                                 v, u, v_ode, u_ode, semi, t; update_from_callback=false)
    # The `UpdateCallback` will set `system.cache.update_callback_used[]` to `true`
    # in the initialization. However, other callbacks might update the system first, hence `update_from_callback`.
```

src/schemes/fluid/entropically_damped_sph/system.jl
```diff
@@ -306,6 +306,10 @@ function update_final!(system::EntropicallyDampedSPHSystem, v, u, v_ode, u_ode,
...
+    check_tvf_configuration(system, system.transport_velocity, v, u, v_ode, u_ode, semi, t;
```
efaulhaber, line 311: Why is this the very last step of `update_final!`? Wouldn't it make more sense to make this the very first step of `update_quantities!`?

efaulhaber (changes requested):

src/schemes/fluid/entropically_damped_sph/system.jl
LasNikas: There is a comment in boundary/system.jl. But I don't remember why
```
# This update depends on the computed quantities of the fluid system and therefore
# has to be in `update_final!` after `update_quantities!`.
```

LasNikas: I don't remember why...

svchb: and now???

efaulhaber: Idk. I think it makes more sense to do that at the beginning to make it fail faster for large problems.

efaulhaber: But I also don't really care.

efaulhaber (approved):

LasNikas (approved):

PR #809, opened 2025-05-15
Fix allocations caused by branching with 1.10

svchb:


src/schemes/boundary/dummy_particles/dummy_particles.jl
```diff
@@ -416,23 +416,13 @@ function compute_pressure!(boundary_model,
         speedup = ceil(Int, Threads.nthreads() / 2)
         parallelize = system_coords isa AbstractGPUArray ||
                       n_boundary_particles < speedup * n_fluid_particles
```
efaulhaber, line 418:
```suggestion
        is_gpu = system_coords isa AbstractGPUArray
        condition_boundary = n_boundary_particles < speedup * n_fluid_particles
        parallelize = is_gpu || condition_boundary || !allow_loop_flipping
```
And then just pass `Val(parallelize)`.

efaulhaber (changes requested):

efaulhaber (approved):

LasNikas (approved):

PR #813, opened 2025-05-21
Make PST GPU-compatible

efaulhaber:
As requested by @LasNikas.

### DIFF
--- a/docs/src/gpu.md
+++ b/docs/src/gpu.md
@@ -13,13 +13,14 @@ Unlike the default cell list, which assumes an unbounded domain,
 this cell list requires a bounding box for the domain.
 For simulations that are bounded by a closed tank, we can simply use the boundary
 of the tank to obtain the bounding box as follows.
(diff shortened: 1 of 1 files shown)


src/schemes/fluid/weakly_compressible_sph/system.jl
```diff
@@ -247,6 +247,23 @@ end
...
+                                  system::WeaklyCompressibleSPHSystem)
+    # When using `SummationDensity`, `v` contains only the velocity
+    return v
+end
+
+@inline function current_velocity(v, ::ContinuityDensity,
+                                  system::WeaklyCompressibleSPHSystem)
+    # When using `ContinuityDensity`, the velocity is stored
+    # in the first `ndims(system)` rows of `v`.
+    return view(v, 1:ndims(system), :)
```
LasNikas, line 264: There could potentially be slicing issues here, no? We’ve encountered a similar problem with `current_pressure`.
https://github.com/JuliaSIMD/StrideArrays.jl/issues/88

svchb (approved):

src/schemes/fluid/weakly_compressible_sph/system.jl
efaulhaber: This might only cause problems when saving or viewing the return value.

LasNikas (approved):

PR #825, opened 2025-06-07
Fix `current_velocity` for EDAC

LasNikas:


LasNikas:
> If this is critical, why didn't any of the tests fail?

idk. Maybe because this is only used by the `ParticleShiftingCallback` and `system_data`

PR #834, opened 2025-06-16
Make PST work for different smoothing length factors

efaulhaber:
On main, PST doesn't work for small smoothing length factors. The following examples are all with the Wendland C2 kernel.
|                         | h = 1.5 * Δx    | h = 2.5 * Δx    |
|:--------------:|:-----------:|:-----------:|
| Δx = 0.02, Δt = 5e-3        |  grafik  |  grafik |

This PR makes it work for a wide range of time steps, smoothing length factors, and resolutions:
|                         | h = 1.5 * Δx    | h = 2.5 * Δx    |
|:--------------:|:-----------:|:-----------:|
| Δx = 0.02, Δt = 1e-3      | grafik |  grafik |
| Δx = 0.02, Δt = 5e-3      | grafik  |  grafik |
| Δx = 0.01, Δt = 5e-3 | grafik | grafik |
| Δx = 0.05, Δt = 1e-2 | grafik | grafik |


efaulhaber:
> Please verify that this is also correct for 3D cases.

This is a slice of the same vortex street simulation in 3D, where the cylinder is now a full sphere.
Top is 1.5Δx, bottom is 2.5Δx. Both with this branch.
anim 0004

On main:
anim 0005

So I guess this is fine in 3D. We will probably only be using consistent shifting in the future anyway.

LasNikas (approved):

svchb (approved):

PR #840, opened 2025-06-18
Write displacement for TLSPH systems

LasNikas:


src/io/write_vtk.jl
```diff
@@ -380,6 +380,10 @@ function write2vtk!(vtk, v, u, t, system::TotalLagrangianSPHSystem; write_meta_d
```
efaulhaber, line 380: While you're at it, could you please fix the velocity?
```julia
    vtk["velocity"] = [current_velocity(v, system, particle)
                       for particle in active_particles(system)]
```
The current version doesn't show up as particle data in ParaView.

src/io/write_vtk.jl
```diff
@@ -380,6 +380,10 @@ function write2vtk!(vtk, v, u, t, system::TotalLagrangianSPHSystem; write_meta_d
 
     vtk["von_mises_stress"] = von_mises_stress(system)
 
+    vtk["displacement"] = [current_coords(system, particle) -
```
efaulhaber, line 383: Why? Is this ever useful information?

efaulhaber (changes requested):

src/io/write_vtk.jl
LasNikas: AFAIK every FEM solver writes out the displacement.
The reason for this is that you can highlight small displacements (e.g. for relatively stiff materials) visually.

src/io/write_vtk.jl
LasNikas: Yes thanks. Same for density btw

efaulhaber (dismissed):

svchb (approved):

efaulhaber (approved):

PR #852, opened 2025-07-01
Fix IC in open boundaries and add `simple_advection_2d.jl`

LasNikas:
The additional example `simple_advection_2d.jl` is to quickly test (new) open boundaries and also to benchmark potential new open boundary implementations.

### DIFF
--- a/examples/fluid/simple_advection_2d.jl
+++ b/examples/fluid/simple_advection_2d.jl
@@ -1,0 +1,146 @@
+using TrixiParticles
+using OrdinaryDiffEq
+
+# ==========================================================================================
+# ==== Resolution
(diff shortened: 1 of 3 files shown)


examples/fluid/simple_advection_2d.jl
```diff
@@ -0,0 +1,144 @@
+# This is a simple advection example with open boundaries. The setup is similar to
+# `periodic_channel_2d.jl`, but uses open boundaries instead of periodic ones.
```
svchb, line 2: Not the same style.

svchb (changes requested):

LasNikas:
superseded by #866

PR #853, opened 2025-07-01
Fix Adami boundary viscosity

efaulhaber:
When using `AdamiPressureExtrapolation`, the velocity extrapolation was passing `system_coords` instead of `v` to `current_velocity`. This argument wasn't used for boundary systems, but for TLSPH, it did not work correctly.

This is on main:
grafik
The boundary layer is clearly smaller at the blade of the fin on the right, which is the integrated TLSPH part. The left are fixed particles, so wall viscosity is working correctly.

This PR:
grafik

### DIFF
--- a/test/schemes/boundary/dummy_particles/dummy_particles.jl
+++ b/test/schemes/boundary/dummy_particles/dummy_particles.jl
@@ -95,7 +95,8 @@
 
         for boundary_system in boundary_systems
             system_name = boundary_system |> typeof |> nameof
-            density_calculator = boundary_system.boundary_model.density_calculator |> typeof |> nameof
+            density_calculator = boundary_system.boundary_model.density_calculator |>
+                                 typeof |> nameof
             @testset "$system_name with $density_calculator" begin
                 boundary_model = boundary_system.boundary_model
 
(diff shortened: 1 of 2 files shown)


test/schemes/boundary/dummy_particles/dummy_particles.jl
```diff
@@ -49,62 +50,81 @@
...
-            AdamiPressureExtrapolation(),
-            BernoulliPressureExtrapolation()
+        boundary_model_adami = BoundaryModelDummyParticles(boundary.density,
+                                                           boundary.mass,
+                                                           state_equation=state_equation,
+                                                           AdamiPressureExtrapolation(),
+                                                           smoothing_kernel,
+                                                           smoothing_length,
+                                                           viscosity=viscosity)
+        boundary_model_bernoulli = BoundaryModelDummyParticles(boundary.density,
```
svchb, line 60:
```suggestion
                                                           viscosity=viscosity)
                                                           
        boundary_model_bernoulli = BoundaryModelDummyParticles(boundary.density,
```

svchb (changes requested):

LasNikas (approved):

PR #863, opened 2025-07-18
Dynamical pressure (open) BC 

LasNikas:
### Literature
https://doi.org/10.1063/5.0254575

based on #866 

# TODO
- [ ] Make it work with PST
- [ ] Make it work with EDAC

# Open questions
- How handle the density in the transition boundary-fluid (`set_particle_density!`)? Is it better to calculate the `inverse_state_eqaution`?
- Impact of density diffusion in boundary zone (see stash: "revise and fix bugs")?
- Investigate the prescribed pressure term (second term eq. 13) 
- Investigate the "smaller smoothing length"-TVF approach (http://dx.doi.org/10.1016/j.jcp.2017.02.016) and the scaled approach (https://www.sciencedirect.com/science/article/pii/S0021999125004863)

based on:
tbd


LasNikas:
superseded by #871

PR #843, opened 2025-06-23
Implement consistent shifting by Sun et al. (2019)

efaulhaber:

PR #864, opened 2025-07-21
Restructure TVF

efaulhaber:
This PR consists of several changes that I did in steps.
## Step 1
The current approach is integrating the transport velocity through stages. So, between time steps, `\tilde{v}` is set to `v`, and then it is integrated for one time step with `d\tilde{v}/dt` set to the term inside the parentheses here:
grafik
Note that this approach is currently flawed, as the `dv/dt` term in the RHS is missing, so only the physical velocity from the last time step is used at every stage.

In the paper, with the different time integration, this transport velocity is only computed once per step:
grafik
While we can interpret this as "transport velocity is integrated within the step starting from v", which leads to the approach explained above, we could also interpret this as "`dr/dt = v + δv`, where `δv` is `dt * v_c` and `v_c` is the second term inside the parentheses above".
This interpretation leads to a `δv = dt * v_c` that is either constant throughout the time step and computed between steps, or `v_c` is recomputed at every stage. The latter is the one I implemented for the following comparisons, but I don't think the difference between computing this only between steps is significant, so we might be able to save some significant computation here.

### Results

anim 0001
As can be seen here in the left column, directly behind the cylinder, the old TVF implementation (top row) is less effective at preventing void regions than the new implementation (top row). The old implementation with p_0 = 100k (which is rho_0 * c^2) is about as effective in preventing void regions as the new version with p_0 = 50k, whereas the new version with p_0 = 100k is about as effective as the old version with p_0 = 200k. This indicates that the new implementation might just be a rescaling of the old implementation.

anim 0000
We can see here, similar to the results before, that the top left and bottom right show similar particle distribution uniformity (mostly in the red area above the wake, but also in the size of the "reordered" region in front of the cylinder), as well as the top right and bottom left.
However, we can now see in the left column that the old TVF implementation changes the solution, producing a slightly wider wake region and lower velocity in the center of the wake. The higher background pressure changes the results even more (bottom right), whereas the top row solution is not affected by the different background pressure.

anim 0002
This is a comparison between TVF in this PR (top), consistent shifting from #843 (middle) and inconsistent shifting (bottom).

## Step 2

The original 2013 paper was using kernel summation to compute the density. When we use this implementation with the continuity equation, we get the same inconsistency as the inconsistent PST, as particles are now moving with the transport velocity, but the physical velocity is used in the continuity equation.
This can lead to an uncontrolled increase of density in a closed simulation.

In "A generalized transport-velocity formulation for smoothed particle hydrodynamics" (Zhang et al. 2017), the density is integrated with the continuity equation and the transport velocity is used here:
grafik


src/schemes/fluid/transport_velocity.jl
```diff
@@ -7,176 +7,117 @@ See [TVF](@ref transport_velocity_formulation) for more details of the method.
...
-function check_tvf_configuration(system::FluidSystem, ::TransportVelocityAdami,
-                                 v, u, v_ode, u_ode, semi, t; update_from_callback=false)
-    # The `UpdateCallback` will set `system.cache.update_callback_used[]` to `true`
-    # in the initialization. However, other callbacks might update the system first, hence `update_from_callback`.
-    if !update_from_callback && !(system.cache.update_callback_used[])
-        throw(ArgumentError("`UpdateCallback` is required when using `TransportVelocityAdami`"))
-    end
+            # Write into the buffer
+            for i in eachindex(delta_v_)
+                @inbounds delta_v[i, particle] += delta_v_[i]
```
efaulhaber, line 117: This needs to contain the time step:
```suggestion
                @inbounds delta_v[i, particle] += time_step * delta_v_[i]
```
I hardcoded this for testing to produce the results in the PR description.

src/general/semidiscretization.jl
```diff
@@ -523,6 +526,14 @@ function update_systems_and_nhs(v_ode, u_ode, semi, t; update_from_callback=fals
         update_pressure!(system, v, u, v_ode, u_ode, semi, t)
     end
 
+    foreach_system(semi) do system
```
LasNikas, line 529: Please add a brief comment, why we need to interpolate boundary values before the final update

Edit: I think you can reuse this comment somehow:
```julia
# This update depends on the computed quantities of the fluid system and therefore
# has to be in `update_boundary_interpolation!` after `update_quantities!`.
```

src/schemes/fluid/entropically_damped_sph/rhs.jl
```diff
@@ -50,10 +50,12 @@ function interact!(dv, v_particle_system, u_particle_system,
                                      sound_speed, m_a, m_b, rho_a, rho_b, grad_kernel)
 
         # Add convection term (only when using `TransportVelocityAdami`)
-        dv_convection = momentum_convection(particle_system, neighbor_system,
-                                            v_particle_system, v_neighbor_system,
-                                            rho_a, rho_b, m_a, m_b,
-                                            particle, neighbor, grad_kernel)
+        dv_convection = dv_transport_velocity(transport_velocity(particle_system),
```
LasNikas, line 53:
```suggestion
        dv_tvf = dv_transport_velocity(transport_velocity(particle_system),
```

src/general/semidiscretization.jl
```diff
@@ -523,6 +526,14 @@ function update_systems_and_nhs(v_ode, u_ode, semi, t; update_from_callback=fals
...
+        v = wrap_v(v_ode, system, semi)
+        u = wrap_u(u_ode, system, semi)
+
+        update_boundary_interpolation!(system, v, u, v_ode, u_ode, semi, t;
+                                       update_from_callback)
```
LasNikas, line 534: You can now remove the `update_from_callback` kwarg for the `update_final!` function and
call the method from `ParticlePackingSystem` also `update_boundary_interpolation!`.

Also, can you please create an Issue that I have to now revise the packing, too.

src/schemes/fluid/weakly_compressible_sph/rhs.jl
```diff
@@ -70,10 +71,12 @@ function interact!(dv, v_particle_system, u_particle_system,
                                                grad_kernel)
 
         # Add convection term (only when using `TransportVelocityAdami`)
-        dv_convection = momentum_convection(particle_system, neighbor_system,
-                                            v_particle_system, v_neighbor_system,
-                                            rho_a, rho_b, m_a, m_b,
-                                            particle, neighbor, grad_kernel)
+        dv_convection = dv_transport_velocity(transport_velocity(particle_system),
```
LasNikas, line 74:
```suggestion
        dv_tvf = dv_transport_velocity(transport_velocity(particle_system),
```

src/schemes/fluid/entropically_damped_sph/system.jl
```diff
@@ -311,10 +313,7 @@ function update_final!(system::EntropicallyDampedSPHSystem, v, u, v_ode, u_ode,
     compute_curvature!(system, surface_tension, v, u, v_ode, u_ode, semi, t)
     compute_stress_tensors!(system, surface_tension, v, u, v_ode, u_ode, semi, t)
     update_average_pressure!(system, system.transport_velocity, v_ode, u_ode, semi)
```
LasNikas, line 315:
```suggestion
    update_average_pressure!(system,  transport_velocity(system), v_ode, u_ode, semi)
```

LasNikas (changes requested):
very nice!

src/schemes/fluid/weakly_compressible_sph/system.jl
```diff
@@ -221,12 +221,14 @@ function Base.show(io::IO, ::MIME"text/plain", system::WeaklyCompressibleSPHSyst
     end
 end
 
-create_cache_tvf(::Val{:wcsph}, initial_condition, ::Nothing) = (;)
+create_cache_tvf_wcsph(initial_condition, ::Nothing) = (;)
 
-function create_cache_tvf(::Val{:wcsph}, initial_condition, ::TransportVelocityAdami)
+function create_cache_tvf_wcsph(initial_condition, ::TransportVelocityAdami)
```
svchb, line 226: Can you move this to transport_velocity.jl? I think moving these to the model implementation makes more sense.

src/schemes/fluid/entropically_damped_sph/system.jl
```diff
@@ -208,14 +208,16 @@ function Base.show(io::IO, ::MIME"text/plain", system::EntropicallyDampedSPHSyst
     end
 end
 
-create_cache_tvf(::Val{:edac}, initial_condition, ::Nothing) = (;)
+create_cache_tvf_edac(initial_condition, ::Nothing) = (;)
 
-function create_cache_tvf(::Val{:edac}, initial_condition, ::TransportVelocityAdami)
+function create_cache_tvf_edac(initial_condition, ::TransportVelocityAdami)
```
svchb, line 213: Can you move this to transport_velocity.jl? I think moving these to the model implementation makes more sense.

src/schemes/fluid/entropically_damped_sph/system.jl
```diff
@@ -208,14 +208,16 @@ function Base.show(io::IO, ::MIME"text/plain", system::EntropicallyDampedSPHSyst
...
+    delta_v = zeros(eltype(initial_condition), ndims(initial_condition),
+                    nparticles(initial_condition))
     pressure_average = copy(initial_condition.pressure)
     neighbor_counter = Vector{Int}(undef, nparticles(initial_condition))
     update_callback_used = Ref(false)
 
-    return (; pressure_average, neighbor_counter, update_callback_used)
+    return (; delta_v, pressure_average, neighbor_counter, update_callback_used)
```
svchb, line 220: update_callback_used is unused?

src/schemes/fluid/weakly_compressible_sph/system.jl
```diff
@@ -221,12 +221,14 @@ function Base.show(io::IO, ::MIME"text/plain", system::WeaklyCompressibleSPHSyst
...
+    delta_v = zeros(eltype(initial_condition), ndims(initial_condition),
+                    nparticles(initial_condition))
     update_callback_used = Ref(false)
 
-    return (; update_callback_used)
+    return (; delta_v, update_callback_used)
```
svchb, line 231: update_callback_used unused?

svchb (changes requested):

src/general/semidiscretization.jl
efaulhaber: Why do you have to revise the packing?

efaulhaber:
Update: I compared the integrated implementation on main (with the bug fixed) with the new implementation. Here is the kinetic energy in the Taylor-Green vortex.
grafik
As expected, the new implementation yields a constant background pressure throughout different time integration methods and step sizes.
Also, the implementation on main with a CFL of 0.25 with the Leapfrog method (the red line shows an adaptive CFL condition, the purple line a fixed time step), which is what the method was designed for in the paper, is close to the gold and grey line, which is the new implementation. The turquoise line is slightly different, probably due to time integration errors.
grafik

src/schemes/fluid/transport_velocity.jl
```diff
@@ -1,179 +1,168 @@
...
+        u_neighbor = wrap_u(u_ode, neighbor_system, semi)
+
+        system_coords = current_coordinates(u, system)
+        neighbor_coords = current_coordinates(u_neighbor, neighbor_system)
+
+        foreach_point_neighbor(system, neighbor_system, system_coords, neighbor_coords,
+                               semi;
+                               points=each_moving_particle(system)) do particle, neighbor,
+                                                                       pos_diff, distance
+            m_a = @inbounds hydrodynamic_mass(neighbor_system, neighbor)
```
svchb, line 111:
```suggestion
            m_a = @inbounds hydrodynamic_mass(system, particle)
```

svchb (changes requested):

src/schemes/fluid/transport_velocity.jl
```diff
@@ -1,179 +1,168 @@
...
-                                 v, u, v_ode, u_ode, semi, t; update_from_callback=false)
-    # The `UpdateCallback` will set `system.cache.update_callback_used[]` to `true`
-    # in the initialization. However, other callbacks might update the system first, hence `update_from_callback`.
-    if !update_from_callback && !(system.cache.update_callback_used[])
-        throw(ArgumentError("`UpdateCallback` is required when using `TransportVelocityAdami`"))
+    sound_speed = system_sound_speed(system)
+
+    set_zero!(delta_v)
+
+    foreach_system(semi) do neighbor_system
```
svchb, line 100:
```suggestion
```

we don't want to shift into another system

svchb (changes requested):

src/schemes/fluid/transport_velocity.jl
svchb: Than you only want to shift into boundary systems. But not actually into other fluid systems.

svchb: image
left is with shifting into another fluid right without... The left has a slightly less sharp interface.

efaulhaber: Do you have any references mentioning this? Or in general, any references using TVF with multi-phase?

svchb: The original Adami paper has a multiphase example but it says nothing about. Its mentioned in the Zhang paper you cite above  but it scales the shifting at the interface.

LasNikas: We definitely need to perform shifting with boundary, open boundary, and TLSPH systems. For other fluid systems, surface detection would be required anyway, where then TVF is disabled for: The interface to "non-included" systems must be computed regardless, and TVG gets disabled in those cases. I would leave it as is

svchb: you don't need surface detection for TVF. The problem with surface detection is that it is so expensive. Still its wrong to shift into other fluid systems.

svchb: I think we need to think about that at a later time :S

svchb:
Can you add the continuity equation based on the advection velocity? At least in my testing this was much better!

efaulhaber:
This is step 2 that I described above. It will be the next PR. This PR doesn't change features, it only refactors the code and fixes a bug.

svchb (approved):

LasNikas (approved):

PR #866, opened 2025-07-25
Revise open boundaries

### LINKED ISSUE #823: Call of `evaluate_characteristics!` is misleading
Either the function name should be changed, or the method should be fully wrapped inside `evaluate_characteristics!`. It is misleading in a sense that the inner call does not actually perform the final evaluation of the characteristics.

LasNikas:
based on #855 

resolves #823


LasNikas:
This is a breaking changes PR

efaulhaber:
This is not ready for review. It looks like it's based on other branches and requires a rebase first.

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -15,184 +12,132 @@ Open boundary system for in- and outflow particles.
...
     smoothing_length = initial_smoothing_length(fluid_system)
 
-    boundary_candidates = fill(false, nparticles(initial_condition))
+    boundary_candidates = fill(false, nparticles(initial_conditions))
     fluid_candidates = fill(false, nparticles(fluid_system))
 
-    return OpenBoundarySPHSystem(boundary_model, initial_condition, fluid_system,
+    boundary_zone_indices = zeros(UInt8, nparticles(initial_conditions))
+
+    boundary_zones_new = map(zone -> BoundaryZone(zone.initial_condition,
```
efaulhaber, line 84: Please add a comment on what is happening here.

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -15,184 +12,132 @@ Open boundary system for in- and outflow particles.
...
-                prescribed_density=prescribed_density,
-                prescribed_velocity=prescribed_velocity)
-    end
+    return system
+end
+
+function create_cache_open_boundary(boundary_model, initial_condition, reference_values)
+    ELTYPE = eltype(initial_condition)
+
+    pressure_references = map(ref -> ref.reference_pressure, reference_values)
```
efaulhaber, line 114:
```suggestion
    # Separate `reference_values` into pressure, density and velocity reference values
    pressure_references = map(ref -> ref.reference_pressure, reference_values)
```
Also, I'm not sure about the name "references". It sounds weird to me. But `density_reference_values` is a bit long. Do you have any other ideas?

src/util.jl
```diff
@@ -11,6 +11,16 @@ end
 
 @inline foreach_noalloc(func, collection::Tuple{}) = nothing
 
+@inline function apply_ith_function(functions, index, args...)
```
efaulhaber, line 14:
```suggestion
# Returns `functions[index](args...)`, but in a type-stable way for a heterogeneous tuple `functions`
@inline function apply_ith_function(functions, index, args...)
```

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -90,21 +108,26 @@ bidirectional_flow = BoundaryZone(; plane=plane_points, plane_normal, particle_s
...
-struct BoundaryZone{BT, IC, S, ZO, ZW, FD, PN}
+struct BoundaryZone{IC, S, ZO, ZW, FD, PN, R}
     initial_condition       :: IC
     spanning_set            :: S
     zone_origin             :: ZO
     zone_width              :: ZW
     flow_direction          :: FD
     plane_normal            :: PN
-    boundary_type           :: BT
+    reference_values        :: R
```
efaulhaber, line 118:
```suggestion
    reference_values        :: R
    # Note that the following can't be static type parameters, as all boundary zones in a system
    # must have the same type, so that we can loop over them in a type-stable way.
```

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -130,9 +135,118 @@ function BoundaryZone(; plane, plane_normal, density, particle_spacing,
...
+    if prescribed_pressure
+        ic.pressure .= pressure_ref.(coordinates_svector, 0)
+    end
+    if prescribed_density
+        ic.density .= density_ref.(coordinates_svector, 0)
+        ic.mass .= ic.density * ic.particle_spacing^NDIMS
+    end
+    if prescribed_velocity
+        ic.velocity .= stack(velocity_ref.(coordinates_svector, 0))
+    end
```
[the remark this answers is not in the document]
efaulhaber: The problem is that all boundary zones in a system must have the same type.

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -130,9 +153,118 @@ function BoundaryZone(; plane, plane_normal, density, particle_spacing,
...
+    else
+        summary_header(io, "BoundaryZone")
+        summary_line(io, "boundary type", boundary_type_name(boundary_zone))
+        summary_line(io, "#particles", nparticles(boundary_zone.initial_condition))
+        summary_line(io, "width", round(boundary_zone.zone_width, digits=3))
+        summary_footer(io)
+    end
 end
 
 function set_up_boundary_zone(plane, plane_normal, flow_direction, density,
```
efaulhaber, line 270: 1. Why does this function have a kwarg? It is only called once and only used internally.
2. Why don't you include `flow_direction` in this function as well? Then you can clean up the constructor a bit.

efaulhaber (changes requested):

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -15,184 +12,132 @@ Open boundary system for in- and outflow particles.
...
     fluid_system_index = Ref(0)
 
     smoothing_length = initial_smoothing_length(fluid_system)
 
-    boundary_candidates = fill(false, nparticles(initial_condition))
+    boundary_candidates = fill(false, nparticles(initial_conditions))
     fluid_candidates = fill(false, nparticles(fluid_system))
 
-    return OpenBoundarySPHSystem(boundary_model, initial_condition, fluid_system,
+    boundary_zone_indices = zeros(UInt8, nparticles(initial_conditions))
```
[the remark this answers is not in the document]
LasNikas: I don't think one wants to simulate more than 255 boundary zones?

src/schemes/boundary/open_boundary/method_of_characteristics.jl
```diff
@@ -1,5 +1,5 @@
 @doc raw"""
-    BoundaryModelLastiwka(; extrapolate_reference_values=nothing)
+    BoundaryModelLastiwkaCharacteristics(; extrapolate_reference_values=nothing)
```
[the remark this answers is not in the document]
LasNikas: But i'm of course fine with `BoundaryModelCharacteristicsLastiwka`

src/schemes/boundary/open_boundary/system.jl
LasNikas: No, I don't. They call it also "reference values" in the literature...

src/schemes/boundary/open_boundary/boundary_zones.jl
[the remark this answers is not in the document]
svchb:
```
function validate_scalar(name, reference_value, ELTYPE)
    if !(reference_value isa Function || reference_value isa Real || isnothing(reference_value))
        throw(ArgumentError("`$name` must be either a function mapping each particle's " *
                            "coordinates and time to its value, or a scalar"))
    end

    if reference_value isa Function
        test_result = reference_value(zeros(1), 0.0)  # only check scalar-ness
        if !(test_result isa Real)
            throw(ArgumentError("`$name` function must be scalar-valued"))
        end
    end

    dummy = convert(ELTYPE, Inf)
    wrapped = wrap_reference_function(reference_value, dummy)
    return wrapped, !isnothing(reference_value)
end

density_ref,    prescribed_density  =
        _validate_and_wrap_scalar("reference_density", reference_density,  ELTYPE)
```

src/schemes/boundary/open_boundary/system.jl
efaulhaber: I'm usually against abbreviations, but here I think I would prefer `pressure_ref_values` over `pressure_references`.
Or even `pressure_reference_values` if the long name is not too tedious.

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -88,10 +87,19 @@ There are three ways to specify the actual shape of the boundary zone:
 plane_points = ([0.0, 0.0], [0.0, 1.0])
 plane_normal=[1.0, 0.0]
 
-# Reference velocity function for the inflow: parabolic velocity profile
+# Constant reference velocity:
```
efaulhaber, line 90: 1. I don't understand. Velocity per particle is now not supported anymore, but pressure per particle still is? Does it even make sense to have values per particle? The particle positions change over time, so why would a particle always have the same reference value unless the reference value is constant over all particles?
2. Please make this a doctest. There should be no untested examples in the docs.
3. I don't like that you define `velocity_in = [1.0, 0.0]` and then overwrite it immediately. Please create two inflows, so that all versions are also tested.

efaulhaber (changes requested):

src/schemes/boundary/open_boundary/boundary_zones.jl
LasNikas: like so? 
image

PR #870, opened 2025-07-30
DummyParticles with ContinuityDensity on GPU is broken

svchb:


svchb:
There is no test for this so this cannot fail. The GPU just crashed.

PR #876, opened 2025-08-01
Use latest version of JuliaFormatter

efaulhaber:
This solves some of the awkward line breaks.
But I would like to wait for https://github.com/domluna/JuliaFormatter.jl/pull/934 in order to fix the rest in the same PR.

### DIFF
--- a/src/callbacks/info.jl
+++ b/src/callbacks/info.jl
@@ -122,9 +122,9 @@ function initialize_info_callback(discrete_callback, u, t, integrator;
 
     # Time integration
     setup = Pair{String, Any}["Start time" => first(integrator.sol.prob.tspan),
-                              "Final time" => last(integrator.sol.prob.tspan),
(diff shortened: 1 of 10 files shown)


src/callbacks/info.jl
```diff
@@ -122,9 +122,9 @@ function initialize_info_callback(discrete_callback, u, t, integrator;
 
     # Time integration
     setup = Pair{String, Any}["Start time" => first(integrator.sol.prob.tspan),
-                              "Final time" => last(integrator.sol.prob.tspan),
-                              "time integrator" => integrator.alg |> typeof |> nameof,
-                              "adaptive" => integrator.opts.adaptive]
+    "Final time" => last(integrator.sol.prob.tspan),
```
efaulhaber, line 125: See https://github.com/domluna/JuliaFormatter.jl/issues/935.

PR #880, opened 2025-08-13
Use fluid system kernel instead of boundary system kernel

svchb:
Non-breaking part of removing the kernel from the boundary.


efaulhaber (changes requested):

test/examples/dam_break_2d_corrections.jl
```diff
@@ -90,21 +93,25 @@
                                          joinpath(examples_dir(), "fluid",
                                                   "dam_break_2d.jl"),
                                          fluid_particle_spacing=particle_spacing,
+                                         boundary_particle_spacing=particle_spacing,
                                          smoothing_length=smoothing_length,
-                                         boundary_density_calculator=SummationDensity(),
+                                         boundary_density_calculator=fluid_density_calculator,
```
efaulhaber, line 98: Why? I thought the point of the test was to have a `SummationDensity` boundary.

test/examples/dam_break_2d_corrections.jl
```diff
@@ -90,21 +93,25 @@
...
                                          fluid_density=fluid_density,
                                          density_diffusion=nothing,
-                                         boundary_layers=5, sol=nothing)
+                                         boundary_layers=4,
+                                         sol=nothing)
 
         # Some correction methods require very small time steps at the beginning of the simulation.
         # An adaptive time integrator makes this easier and faster.
-        sol = solve(ode, RDPK3SpFSAL35(), save_everystep=false, callback=callbacks)
+        sol = solve(ode, RDPK3SpFSAL35(), dt=1e-8, save_everystep=false, callback=callbacks)
```
efaulhaber, line 114: Why this extremely small initial time step?

efaulhaber (changes requested):

src/general/corrections.jl
```diff
@@ -340,24 +359,11 @@ function compute_gradient_correction_matrix!(corr_matrix::AbstractArray, system,
...
-                                         smoothing_kernel, pos_diff, distance,
-                                         smoothing_length_, system, particle)
-                return corrected_kernel_grad(smoothing_kernel, pos_diff, distance,
-                                             smoothing_length_, KernelCorrection(), system,
-                                             particle)
-            end
+            smoothing_length_ = smoothing_length(system, neighbor_system, particle)
 
-            grad_kernel = compute_grad_kernel(correction, smoothing_kernel, pos_diff,
-                                              distance, smoothing_length_, system, particle)
```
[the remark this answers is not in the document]
svchb: This is not needed this is from the PR when Niklas introduced variable smoothing length. This is still functionally the same but I guess this slipped through the review.

test/examples/dam_break_2d_corrections.jl
svchb: Most of the time we use the same density calculators for both the boundary and the fluid.
The test still has summation_density for both as a test.

src/general/corrections.jl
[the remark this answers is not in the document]
svchb: He originally tried to remove 'corrected_kernel_grad' I guess this was left over

PR #885, opened 2025-08-21
Use fluid mass instead of boundary mass in RHS

svchb:
Using the boundary particle mass in the RHS is not correct. If we have for example a different particle spacing for the boundary this will no longer work. It also doesn't work correctly if we have multiple fluids with different density.
I will follow-up with a breaking-change to fix the boundary for multi-phase/fluid.
Main in white, this PR in green example shows a boundary with a different particle spacing to the fluid.
image


efaulhaber:
This does not make any sense. The correct momentum equation contains `m_b`, not `m_a`.
In fact, the force should contain the product `m_a * m_b`, but we want the acceleration on particle `a`, so we have to divide by `m_a`.

On main, we get the same acceleration when `a` is much smaller. This makes sense, as the fluid is supposed to experience the same acceleration no matter the resolution.
Even more importantly, a 100x smaller particle `b` should influence particle `a` 100x less, otherwise a higher resolved boundary would induce a larger force on the fluid.

This is easy to demonstrate by simply increasing the boundary resolution.
Here is the hydrostatic tank with spacing ratio 10 at t=0 and at t=1 on main:
grafik
grafik

This is this PR at t=0.44 (after that it crashes):
grafik

PR #894, opened 2025-09-05
Fix pressure setter

LasNikas:


efaulhaber (commented):
What about `set_particle_density!`?

LasNikas:
we have this in the `fluid.jl` file

efaulhaber:
But for `SummationDensity`, this is just
```julia
@inline set_particle_density!(v, system, ::SummationDensity, particle, density) = v
```
Does this work? Does the summation density computation come after this? Or does it have to be set similar to pressure?

efaulhaber (approved):

PR #895, opened 2025-09-06
[Proof of Concept] Delta boundaries

efaulhaber:
As long as the boundary density is also initialized with a hydrostatic gradient, everything works nicely and yields good results even after a long runtime (with Antuono density diffusion of course):
grafik

However, when initializing both the fluid and the boundary with the rest density, which is closer to a realistic simulation without gravity (like a moving fin where pressure gradients develop unpredictably during the simulation), the fluid slowly drips through the boundary:
grafik
This result is obtained by running the hydrostatic tank example in exactly the configuration that I pushed.

This makes me question the usefulness of this boundary condition. The idea was that it is even more resistant to penetration than pressure extrapolation, while providing volume preservation with consistent shifting and being easy to implement.
Now that it seems to have penetration issues, it's going to be hard to argue its usefulness.


efaulhaber:
I found that in my fin simulation testing branch where I found this to be useful, I did not implement it "correctly".
I applied density diffusion only for boundary-fluid interaction, but not for fluid-boundary interaction.
This means that outlier values (one particle with too high or too low density), which usually cause problems with `ContinuityDensity` boundaries will slowly converge to the density of the neighboring fluid. However, it is no physical "density is moving between fluid and boundary", as this effect is asymmetric and the fluid is not affected.

In this case, the theoretical argument for volume preservation, which is that the sum of the density derivative over all particles (including boundaries) is zero, does not hold anymore. This argument meant that boundary particles act as a reservoir for density, so that the total density is not exactly preserved but bounded.
With the asymmetric formulation, boundary particles can receive density from the fluid without the fluid losing density, so this whole argumentation does not make sense anymore and the total volume error of the fluid can still grow arbitrarily large over time, just as with other boundary conditions.

PR #897, opened 2025-09-09
[Proof of Concept] Integrate energy added through "fixed" TLSPH particles

efaulhaber:
This PR implements two versions of this.
1. Integration of energy with the main time integrator via an `EnergyCalculatorSystem` (introduced in #667).
2. A cheaper version that evaluates the forces on "fixed" particles only in a callback and then integrates the energy with the explicit Euler method. In my small experiment this added a 6% overhead vs 25% of version 1. It can be made even cheaper by applying the callback only every n steps at the cost of larger time integration errors.


efaulhaber:
This feature is now properly implemented as `MechanicalWorkCalculatorCallback`, or, with #1228, `MechanicalWorkCalculator`.

PR #913, opened 2025-09-18
Make `kernel_grad` threshold compatible with single precision and small particle spacing

LasNikas:
This PR adapts the threshold used in kernel_grad to eps(h^2) instead of a fixed value `sqrt(eps(typeof(h)))`. This change ensures robust behavior for both single and double precision, especially when using small smoothing lengths or particle spacings. The comparison `distance^2 < eps(h^2)` is preferred over `distance < sqrt(eps(h^2))` for efficiency, since it avoids an unnecessary square root.

Previously, the threshold could be too strict for `Float32`, causing all gradients to be set to zero for small `h`. By using `eps(h^2)`, the threshold automatically adapts to the precision and scale of the simulation.

Additionally, I attempted to address the ugly formatting issue in the rhs:
image

### DIFF
--- a/src/schemes/boundary/open_boundary/method_of_characteristics.jl
+++ b/src/schemes/boundary/open_boundary/method_of_characteristics.jl
@@ -178,6 +178,7 @@ function evaluate_characteristics!(system, v, u, v_ode, u_ode, semi, t)
             for neighbor in each_integrated_particle(system)
                 # Make sure that only neighbors in the influence of
                 # the fluid particles are used.
+                # Handle numerical precision issues (see also https://github.com/trixi-framework/TrixiParticles.jl/pull/913)
                 if volume[neighbor]^2 > eps(volume[neighbor]^2)
                     avg_J1 += previous_characteristics[1, neighbor]
                     avg_J2 += previous_characteristics[2, neighbor]
--- a/src/schemes/boundary/open_boundary/method_of_characteristics.jl
+++ b/src/schemes/boundary/open_boundary/method_of_characteristics.jl
@@ -178,7 +178,7 @@ function evaluate_characteristics!(system, v, u, v_ode, u_ode, semi, t)
             for neighbor in each_integrated_particle(system)
                 # Make sure that only neighbors in the influence of
                 # the fluid particles are used.
-                if isapprox(volume[neighbor], 0)
+                if volume[neighbor]^2 > eps(volume[neighbor]^2)
                     avg_J1 += previous_characteristics[1, neighbor]
                     avg_J2 += previous_characteristics[2, neighbor]
                     avg_J3 += previous_characteristics[3, neighbor]
--- a/src/schemes/boundary/open_boundary/method_of_characteristics.jl
+++ b/src/schemes/boundary/open_boundary/method_of_characteristics.jl
@@ -178,7 +178,7 @@ function evaluate_characteristics!(system, v, u, v_ode, u_ode, semi, t)
             for neighbor in each_integrated_particle(system)
                 # Make sure that only neighbors in the influence of
                 # the fluid particles are used.
-                if volume[neighbor]^2 > eps(volume[neighbor]^2)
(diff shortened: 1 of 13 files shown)


src/schemes/boundary/open_boundary/method_of_characteristics.jl
```diff
@@ -178,7 +178,8 @@ function evaluate_characteristics!(system, v, u, v_ode, u_ode, semi, t)
             for neighbor in each_integrated_particle(system)
                 # Make sure that only neighbors in the influence of
                 # the fluid particles are used.
-                if volume[neighbor] > sqrt(eps())
+                # Handle numerical precision issues (see also https://github.com/trixi-framework/TrixiParticles.jl/pull/913)
+                if volume[neighbor]^2 > eps(volume[neighbor]^2)
```
svchb, line 182: I am not certain if this is correct also the above isapprox statement is basically also x<sqrt(eps).
So we probably also need to check all the isapprox statements (I have not checked if there are more).
@efaulhaber what is your opinion?

svchb (changes requested):

src/schemes/boundary/open_boundary/method_of_characteristics.jl
efaulhaber: Yes, this doesn't make sense. This is `x > eps(x)`, which is always true except for `x == 0.0`.

The correct solution would be to compare against something of the same unit and scale, so W(0, h). But of course, this would be expensive.

Maybe this makes sense as well, while being much cheaper?
```julia
volume[neighbor] > eps(1 / smoothing_length^ndims)
```
or, equivalent and mathematically nicer:
```julia
volume[neighbor] * smoothing_length^ndims > eps(eltype(smoothing_length))
```
The kernel has the unit [1 / m^3], so the left-hand side should be in the order of 1, no?

src/schemes/boundary/open_boundary/method_of_characteristics.jl
```diff
@@ -165,8 +165,9 @@ function evaluate_characteristics!(system, v, u, v_ode, u_ode, semi, t)
     # of fluid particles by using the average of the values of the previous time step.
     # See eq. 27 in Negi (2020) https://doi.org/10.1016/j.cma.2020.113119
     @threaded semi for particle in each_integrated_particle(system)
-        # Particle is outside of the influence of fluid particles
-        if isapprox(volume[particle], 0)
+        # Particle is outside of the influence of fluid particles.
+        # Handle numerical precision issues (see also https://github.com/trixi-framework/TrixiParticles.jl/pull/913)
+        if volume[particle] * smoothing_length^ndims(system) < eps(eltype(smoothing_length))
```
efaulhaber, line 170:
```suggestion
        # `volume` is in the order of 1 / h^d, so volume * h^d is in the order of 1.
        if volume[particle] * smoothing_length^ndims(system) < eps(eltype(smoothing_length))
```
It is also confusing that `volume` is not actually a volume but an inverse volume. Would it make sense to change the variable name in another PR?

src/schemes/fluid/surface_tension.jl
```diff
@@ -174,8 +174,9 @@ end
                                        neighbor_system::AbstractFluidSystem,
                                        particle, neighbor, pos_diff, distance,
                                        rho_a, rho_b, grad_kernel)
-    # No cohesion with oneself
-    distance < sqrt(eps()) && return zero(pos_diff)
+    # No cohesion with oneself.
+    # Handle numerical precision issues (see also https://github.com/trixi-framework/TrixiParticles.jl/pull/913).
```
efaulhaber, line 178:
```suggestion
    # No cohesion with oneself. See `src/general/smoothing_kernels.jl` for more details.
```
Same for all other changes below.

efaulhaber (changes requested):

svchb (dismissed):

PR #926, opened 2025-09-25
Fix `convert_particle!`

LasNikas:


src/schemes/boundary/open_boundary/system.jl
```diff
@@ -305,7 +305,10 @@ end
 
     # Check if particle is in- or outside the fluid domain.
     # `face_normal` is always pointing into the fluid domain.
-    if signbit(dot(relative_position, boundary_zone.face_normal))
+    dist_to_transition = dot(relative_position, -boundary_zone.face_normal)
+    dist_free_surface = boundary_zone.zone_width - dist_to_transition
+    if dist_free_surface < dist_to_transition
+        # Particle is outside the fluid domain
```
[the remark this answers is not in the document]
LasNikas: This is a variant that is more robust against rounding errors in `is_in_boundary_zone` or rather in `spanning_set`. Assuming that the particle actually enters the fluid domain, but the dot product is barely negative, it is incorrectly assumed that the particle does not enter the fluid domain, but leaves the entire domain.

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -305,7 +305,10 @@ end
 
     # Check if particle is in- or outside the fluid domain.
     # `face_normal` is always pointing into the fluid domain.
```
efaulhaber, line 307:
```suggestion
    # `face_normal` is always pointing into the fluid domain.
    # Since this function is called for a particle that left the boundary zone,
    # it is sufficient to check if it is closer to the free surface or the transition phase
    # to determine if it exited the boundary zone into the fluid or through the free surface.
```

src/schemes/boundary/open_boundary/system.jl
efaulhaber, line 304:
```suggestion
    # Position relative to the origin of the transition face
    relative_position = current_coords(u, system, particle) - boundary_zone.zone_origin
```

efaulhaber (changes requested):
Took me a long time and a long discussion with you to understand what is going on here. I think these two comments clarify my confusion.

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -327,7 +327,7 @@ function set_up_boundary_zone(boundary_face, face_normal, density, particle_spac
     # First vector of `spanning_vectors` is normal to the boundary face.
     dot_face_normal = dot(normalize(spanning_set[:, 1]), face_normal)
 
-    if !isapprox(abs(dot_face_normal), 1, rtol=1e-5)
+    if !isapprox(abs(dot_face_normal), 1)
```
LasNikas, line 330: @efaulhaber I think this change should make this check valid again:
```julia
if signbit(dot(relative_position, boundary_zone.face_normal))
     ...
```
what do you think?
I prefer this check to `dist_free_surface < dist_to_transition` as the outflow particle might have exited the boundary zone through the top or bottom face. In this case, it is better to check the entire plane spanned by the boundary face.

src/schemes/boundary/open_boundary/boundary_zones.jl
```diff
@@ -426,6 +426,12 @@ function update_boundary_zone_indices!(system, u, boundary_zones, semi)
                 system.boundary_zone_indices[particle] = zone_id
             end
         end
+
+        # This typically only occurs if `face_normal` is not exactly normal to the `boundary_face`.
+        # In such cases, particles that are actually outside the simulation domain (outflow particles)
+        # may be incorrectly kept active as inflow particles and therefore cannot be assigned to any boundary zone.
+        # See https://github.com/trixi-framework/TrixiParticles.jl/pull/926 for details.
+        @assert system.boundary_zone_indices[particle] != 0 "No boundary zone found for active buffer particle"
```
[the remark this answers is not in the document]
LasNikas: I think floating-point precision issues could still cause edge cases where particles fall between zones.
It is just a early detection. And also protection for future changes.

src/preprocessing/geometries/geometries.jl
```diff
@@ -102,8 +102,11 @@ function planar_geometry_to_face(planar_geometry::TriangleMesh)
         throw(ArgumentError("geometry is not planar"))
     end
 
+    computed_face_normal = SVector(normalize(cross(edge1, edge2))...)
```
efaulhaber, line 105:
```suggestion
    # The `face_normal` computed above might not be exactly orthogonal to the plane
    # spanned by `edge1` and `edge2`, but this is important for some computations later.
    computed_face_normal = SVector(Tuple(normalize(cross(edge1, edge2))))
```

efaulhaber (changes requested):

efaulhaber (approved):

PR #931, opened 2025-09-30
Automatic ordering of clamped particles

LasNikas:


src/general/initial_condition.jl
```diff
@@ -415,3 +415,16 @@ function find_too_close_particles(coords, min_distance)
 
     return result
 end
+
+function move_particles_to_end!(ic::InitialCondition, particle_ids_to_move)
+    sort_key = [i in particle_ids_to_move ? 1 : 0 for i in eachparticle(ic)]
+    permutation = sortperm(sort_key)
```
svchb, line 421:
```suggestion
    # determine a permutation that sorts 'sort_key' in ascending order
    permutation = sortperm(sort_key)
```

src/schemes/structure/total_lagrangian_sph/system.jl
```diff
@@ -13,6 +15,16 @@ all relevant quantities and operators are measured with respect to the
...
 
+There are two approaches to specify clamped particles in the system:
+
+1. **Manual ordering**: Provide the number of clamped particles via `n_clamped_particles`,
+   ensuring that these particles are the last entries in the `InitialCondition`
+   (see the info box below for details).
+
+2. **Automatic ordering**: Pass the indices of the particles to be clamped via `clamped_particles`.
+   In this case, the specified particles will be internally reordered so that they become
+   the last particles in the `InitialCondition`.
```
svchb, line 26: I would prefer if you removed the first option. If the clamped particles are the last particles providing the indices of the last particles achieves the same  thing.

svchb (changes requested):
A test is missing. 

If you remove the first option it is a breaking change.

src/schemes/structure/total_lagrangian_sph/system.jl
LasNikas: Hm but then we have a breaking change and I have to wait until the next major release.
Why can't we have both?
@efaulhaber  what do you think?

svchb: Anyway I think we can publish 0.4.2 next week. So we could do a smaller major release 0.5 after that.

svchb: @LasNikas can you please prepare an PR to remove this deprecated behavior and add comments here that this option is deprecated?

src/schemes/structure/total_lagrangian_sph/system.jl
```diff
@@ -110,6 +118,27 @@ function TotalLagrangianSPHSystem(initial_condition, smoothing_kernel, smoothing
...
+    # Backwards compatibility: `n_clamped_particles` is deprecated.
+    # Emit a deprecation warning and (if the user didn't supply explicit indices)
+    # convert the old `n_clamped_particles` convention to `clamped_particles`.
+    if n_clamped_particles != 0
+        Base.depwarn("keyword `n_clamped_particles` is deprecated and will be removed in a future release; " *
+                     "pass `clamped_particles` (Vector{Int} of indices) instead.",
+                     :n_clamped_particles)
+        if isempty(clamped_particles)
+            clamped_particles = collect((n_particles - n_clamped_particles + 1):n_particles)
+        end
```
efaulhaber: Throw an error if both are used to avoid confusion.

src/general/initial_condition.jl
```diff
@@ -415,3 +415,35 @@ function find_too_close_particles(coords, min_distance)
...
+    invalid_id = findfirst(i -> i <= 0 || i > nparticles(ic), particle_ids_to_move)
+    isnothing(invalid_id) || throw(BoundsError(ic, invalid_id))
+
+    sort_key = [i in particle_ids_to_move ? 1 : 0 for i in eachparticle(ic)]
+    # determine a permutation that sorts 'sort_key' in ascending order
+    permutation = sortperm(sort_key)
+
+    ic.coordinates .= ic.coordinates[:, permutation]
+    ic.velocity .= ic.velocity[:, permutation]
+    move_particles_to_end!(ic.mass, particle_ids_to_move)
```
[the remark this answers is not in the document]
LasNikas: Does this make a different?

src/schemes/structure/total_lagrangian_sph/system.jl
```diff
@@ -90,12 +97,13 @@ end
 
 function TotalLagrangianSPHSystem(initial_condition, smoothing_kernel, smoothing_length,
                                   young_modulus, poisson_ratio;
-                                  n_clamped_particles=0, clamped_particles_motion=nothing,
-                                  boundary_model=nothing,
+                                  n_clamped_particles=0,
+                                  clamped_particles::Vector{Int}=Int[],
```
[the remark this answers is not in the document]
efaulhaber: Why restrict it to a `Union{Vector{Int}, AbstractUnitRange}`?

examples/structure/oscillating_beam_2d.jl
```diff
@@ -61,7 +61,7 @@ smoothing_kernel = WendlandC2Kernel{2}()
 
 structure_system = TotalLagrangianSPHSystem(structure, smoothing_kernel, smoothing_length,
                                             material.E, material.nu,
-                                            n_clamped_particles=nparticles(clamped_particles),
+                                            clamped_particles=(nparticles(beam) + 1):nparticles(structure),
```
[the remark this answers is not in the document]
LasNikas: It would be easier. However, if you `trixi_include` the setup, the  `InitialCondition`s already have modified indexing. This means that the clamped particles are already the last entries. With `clamped_particles=1:nparticles(clamped_particles)`, you would therefore address the wrong indexing.

src/schemes/structure/total_lagrangian_sph/system.jl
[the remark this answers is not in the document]
efaulhaber: A step range for example. I mean, what do you need from the `clamped_particles`? Only that it supports sorting, so just remove the dispatch and the user will get an error `sort(::MyStupidDatatype) is not defined`.

examples/structure/oscillating_beam_2d.jl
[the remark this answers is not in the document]
LasNikas: I know. But that doesn't work if you run `falling_water_column_2d.jl` which includes the oscillating beam example. 
The initial conditino `structure` has then modified indexing. I mean that can be really confusing.
However, we can copy the initial condition in the constructor so that the indexing of the original IC is not modified.

src/schemes/structure/total_lagrangian_sph/system.jl
```diff
@@ -137,35 +137,40 @@ function TotalLagrangianSPHSystem(initial_condition, smoothing_kernel, smoothing
         @assert allunique(clamped_particles) "`clamped_particles` contains duplicate particle indices"
 
         n_clamped_particles = length(clamped_particles)
-        move_particles_to_end!(initial_condition, collect(clamped_particles))
-        move_particles_to_end!(young_modulus, collect(clamped_particles))
-        move_particles_to_end!(poisson_ratio, collect(clamped_particles))
+        ic_sorted_indices = deepcopy(initial_condition)
+        young_modulus_sorted_indices = copy(young_modulus)
+        poisson_ratio_sorted_indices = copy(poisson_ratio)
+        move_particles_to_end!(ic_sorted_indices, collect(clamped_particles))
```
efaulhaber: Why do you need the `collect`? This should work perfectly fine with just a range without allocating an extra vector.

src/schemes/structure/total_lagrangian_sph/system.jl
```diff
@@ -137,35 +137,40 @@ function TotalLagrangianSPHSystem(initial_condition, smoothing_kernel, smoothing
         @assert allunique(clamped_particles) "`clamped_particles` contains duplicate particle indices"
 
         n_clamped_particles = length(clamped_particles)
-        move_particles_to_end!(initial_condition, collect(clamped_particles))
-        move_particles_to_end!(young_modulus, collect(clamped_particles))
-        move_particles_to_end!(poisson_ratio, collect(clamped_particles))
+        ic_sorted_indices = deepcopy(initial_condition)
```
efaulhaber:
```suggestion
        initial_condition_sorted = deepcopy(initial_condition)
```
I think that's easier to understand in the code below.

svchb (approved):

PR #937, opened 2025-10-03
Reduce `tspan` in tests

LasNikas:


test/examples/examples_fluid.jl
```diff
@@ -394,7 +394,7 @@
     end
 
     @trixi_testset "fluid/pipe_flow_3d.jl" begin
-        @trixi_test_nowarn trixi_include(@__MODULE__,
+        @trixi_test_nowarn trixi_include(@__MODULE__, tspan=(0, 0.1),
```
svchb, line 397: only 23 timesteps are performed are you sure this makes sense?

efaulhaber: For the poiseaulle flow, I guess it's fine because that should be tested by a proper validation test anyway (@LasNikas you are working on this, right?). For the pipe flow, one could argue that a longer tspan makes it more likely to catch an unstable simulation, but I guess the benefit is very low, considering that only a small fraction of unstable simulations actually crash with "instability detected". So I'm fine with the shorter tspan.

efaulhaber (approved):

PR #940, opened 2025-10-06
Add `MechanicalWorkCalculatorCallback`

efaulhaber:
This PR adds a callback that integrates energy added through clamped TLSPH particles. It can also be used to integrate energy added to the fluid through all TLSPH particles, which is equivalent to the work done by the structure acting on the fluid and from which we can compute the propulsion that a moving structure produces.

This PR implements the second version demonstrated in #897:

> A cheaper version that evaluates the forces on "fixed" particles only in a callback and then integrates the energy with the explicit Euler method. In my small experiment this added a 6% overhead vs 25% of version 1. It can be made even cheaper by applying the callback only every n steps at the cost of larger time integration errors.


src/callbacks/energy_calculator.jl
```diff
@@ -0,0 +1,216 @@
...
+                                  only_compute_force_on_fluid=false)
+    ELTYPE = eltype(system)
+    system_index = system_indices(system, semi)
+
+    # Allocate buffer to write accelerations for all particles (including clamped ones)
+    dv = allocate(semi.parallelization_backend, ELTYPE, (ndims(system), nparticles(system)))
+
+    cb = EnergyCalculatorCallback(interval, Ref(zero(ELTYPE)), Ref(zero(ELTYPE)),
+                                  system_index, dv, eachparticle,
+                                  only_compute_force_on_fluid)
```
svchb: initializing t to 0 is not correct if the simulation has been restarted.

svchb (changes requested):

src/callbacks/energy_calculator.jl
```diff
@@ -0,0 +1,216 @@
...
+energy_cb = EnergyCalculatorCallback(system, semi; interval=2)
+
+# After the simulation, retrieve the calculated energy
+total_energy = calculated_energy(energy_cb)
+
+# output
+0.0
+```
+"""
+struct EnergyCalculatorCallback{T, DV, EP}
```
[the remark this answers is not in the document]
svchb: Because its basically free 2nd order. I always use trapezoidal rule instead of euler in explicit cases and euler only for implicit cases in such processes.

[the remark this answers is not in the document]
svchb: Do you have a plot of the power generated in your case?
I guess its varies alot between calls.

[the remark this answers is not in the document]
efaulhaber: Moved to https://github.com/trixi-framework/TrixiParticles.jl/issues/1156 since this PR is in review for half a year now.

src/callbacks/energy_calculator.jl
```diff
@@ -0,0 +1,237 @@
...
+end
+
+function EnergyCalculatorCallback(system::AbstractStructureSystem, semi; interval=1,
+                                  eachparticle=(n_integrated_particles(system) + 1):nparticles(system),
+                                  only_compute_force_on_fluid=false)
+    ELTYPE = eltype(system)
+    system_index = system_indices(system, semi)
+
+    # Allocate buffer to write accelerations for all particles (including clamped ones)
+    dv = allocate(semi.parallelization_backend, ELTYPE, (ndims(system), nparticles(system)))
```
svchb:
```suggestion
    dv = allocate(semi.parallelization_backend, ELTYPE, (v_nvariables(system), nparticles(system)))
```

src/callbacks/energy_calculator.jl
```diff
@@ -0,0 +1,237 @@
...
+
+# Note that `Semidiscretization` might create a deep copy of the system,
+# which means we have to extract the new system from `semi`.
+# When working with GPUs, `semidiscretize` also creates a deep copy of `semi` and another
+# copy of the system, so the clean way to get the correct new system is this:
+semi_new = ode.p
+system_new = semi_new.systems[1]
+
+# Create an energy calculator callback that is called every 2 time steps
+energy_cb = EnergyCalculatorCallback(system_new, semi; interval=2)
```
svchb:
```suggestion
energy_cb = EnergyCalculatorCallback(system_new, semi_new; interval=2)
```

svchb (changes requested):

svchb (dismissed):

svchb (approved):

src/callbacks/energy_calculator.jl
```diff
@@ -0,0 +1,238 @@
+"""
+    EnergyCalculatorCallback(system, semi; interval=1,
+                             eachparticle=(n_integrated_particles(system) + 1):nparticles(system),
+                             only_compute_force_on_fluid=false)
```
LasNikas:
```suggestion
    EnergyCalculatorCallback(system::AbstractStructureSystem, semi; interval=1,
                             eachparticle=(n_integrated_particles(system) + 1):nparticles(system),
                             only_compute_force_on_fluid=false)
```
Is this callback only for structure systems? 
Without having looked at the code yet, shouldn't the callback name suggest that?
Because I was wondering what the point of this is: `eachparticle=(n_integrated_particles(system) + 1):nparticles(system)`?

src/callbacks/energy_calculator.jl
```diff
@@ -0,0 +1,238 @@
...
+
+Callback that accumulates the work done by a set of particles in a
+[`TotalLagrangianSPHSystem`](@ref) by integrating the instantaneous power over time.
+
+With the default arguments it tracks the energy contribution of the clamped particles
+that follow a [`PrescribedMotion`](@ref). By selecting different particles it can also be
```
LasNikas: What do you mean by "different particles"?
Do you mean particles near the fluid domain? Or different types of particles?

src/callbacks/energy_calculator.jl
```diff
@@ -0,0 +1,238 @@
...
+    `(n_integrated_particles(system) + 1):nparticles(system)`.
+- **Fluid load measurement** -- set `eachparticle=eachparticle(system)` together with
+    `only_compute_force_on_fluid=true` to accumulate the work that the entire structure
+    exerts on the surrounding fluid (useful for drag or lift estimates).
+
+Internally the callback integrates the instantaneous power, i.e. the dot product between
+the force exerted by the particle and its prescribed velocity, using an explicit Euler
+time integration scheme.
+
+The accumulated value can be retrieved via [`calculated_energy`](@ref).
```
LasNikas: What value? Do you mean the total energy of the system?

src/callbacks/mechanical_work_calculator.jl
```diff
@@ -0,0 +1,238 @@
+"""
+    EnergyCalculatorCallback(system, semi; interval=1,
+                             eachparticle=(n_integrated_particles(system) + 1):nparticles(system),
+                             only_compute_force_on_fluid=false)
+
+Callback that accumulates the work done by a set of particles in a
+[`TotalLagrangianSPHSystem`](@ref) by integrating the instantaneous power over time.
```
LasNikas, line 7: The implementation seems to accumulate the mechanical work ∫F⋅vdt rather than an energy state directly. Would a name like `WorkCalculatorCallback` or `AccumulatedWorkCallback` better reflect the computed quantity, since the accumulated work is not necessarily identical to the system’s actual energy state?

LasNikas (commented):
Out of curiosity: was there a particular reason to integrate the work via ∫F⋅vdt instead of ∫F⋅ds? I assume the chosen formulation integrates more naturally with the time stepping and avoids explicitly tracking particle displacements between time steps.

src/callbacks/energy_calculator.jl
efaulhaber: It says in the description that it only works for TLSPH. It would make more sense here to use `::TotalLagrangianSPHSystem` because it doesn't work for other structure systems, but this name is not yet available here due to include orders.

src/callbacks/energy_calculator.jl
efaulhaber: Other particles than the clamped particles. How about "by selecting a different particle set"?

src/callbacks/energy_calculator.jl
efaulhaber: Yes, but "total energy" is something else.

src/callbacks/mechanical_work_calculator.jl
efaulhaber: Perhaps that makes more sense. What do you say, @svchb?

Maybe more verbose as `MechanicalWorkCalculatorCallback`?

efaulhaber:
> Out of curiosity: was there a particular reason to integrate the work via ∫F⋅vdt instead of ∫F⋅ds? I assume the chosen formulation integrates more naturally with the time stepping and avoids explicitly tracking particle displacements between time steps.

Instead of computing
```julia
energy[] += dot(-F_particle, velocity) * dt
```
we can compute
```julia
energy[] += dot(-F_particle, ds)
```
but then we would have to store the previous particle positions in the callback to compute `ds`. I don't know if that would make a difference. Perhaps if the time step for energy calculation is large. This is something I can try later when I extend this callback, as this is the first version that is not yet useful for my actual simulation.

svchb:
This version is also not currently working on GPU?

src/callbacks/mechanical_work_calculator.jl
svchb: `MechanicalWorkCalculatorCallback` is fine

efaulhaber:
> This version is also not currently working on GPU?

Not tested, but probably not. Out of scope for this PR.

LasNikas (approved):

PR #941, opened 2025-10-06
Interpolate viscous velocity

LasNikas:


src/general/interpolation.jl
```diff
@@ -716,3 +750,67 @@ function divide_by_shepard_coefficient!(field::AbstractArray{T, 3}, shepard_coef
...
     return field
 end
+
+@inline function interpolate_viscous_velocity(point_coords_, semi, ref_system, v_ode, u_ode;
+                                              smoothing_length=initial_smoothing_length(ref_system),
+                                              neighborhood_searches=process_neighborhood_searches(semi,
+                                                                                                  u_ode,
+                                                                                                  ref_system,
+                                                                                                  smoothing_length,
+                                                                                                  point_coords_))
```
svchb, line 760: Formatting is all broken here

src/general/interpolation.jl
```diff
@@ -716,3 +750,67 @@ function divide_by_shepard_coefficient!(field::AbstractArray{T, 3}, shepard_coef
 
     return field
 end
+
+@inline function interpolate_viscous_velocity(point_coords_, semi, ref_system, v_ode, u_ode;
```
svchb, line 754: Can't you include that in interpolate_system?

svchb (changes requested):

src/general/interpolation.jl
LasNikas: I need two separate `shepard_coefficient`s. How do I decide which one to apply to the field?
image

src/general/interpolation.jl
LasNikas: What do you mean with broeken?

PR #952, opened 2025-10-13
Make interpolation with different smoothing length work on the GPU

### LINKED ISSUE #950: Interpolation with a different smoothing length than the computation is broken

efaulhaber:
fixes #950


svchb (changes requested):
Can you please add a test?

svchb (dismissed):

svchb (dismissed):

svchb (approved):

PR #956, opened 2025-10-15
Allow a different eltype only for the coordinates

efaulhaber:
This PR adds a keyword `coordinates_eltype` to the `RectangularTank`, `RectangularShape` and `SphereShape`, which defaults to `Float64`.
This will make the coordinates `Float64`, while everything else is still `Float32`.
For large numbers of particles, the particle spacing becomes very small relative to the domain size, which create larger errors in the distances between neighboring particles (relative to the particle spacing). With `Float32` coordinates, these errors can easily grow to a point where they cause artifacts in the simulation.

This can easily be demonstrated by running the 2D dam break and passing `min_coordinates = (1000.0, 1000.0)` to the `RectangularTank`. With `Float32`, this yields:
grafik
With this PR and `Float64` coordinates:
grafik

Here is a benchmark of a full dam break simulation and of just the fluid-fluid interaction kernel with 320k particles on an Nvidia RTX A4500.
| Full simulation | Time | Relative |
|---|---|---|
| Float64 | 213 s | 13.6× |
| Float32 | 15.7 s | 1× |
| Float32 + Float64 coordinates | 21.1 s | 1.34× |

| fluid-fluid interact | Time | Relative |
|---|---|---|
| All Float64 | 16.823 ms | 20× |
| All Float32 | 839.356 μs | 1× |
| Float32 + Float64 coordinates | 1.278 ms | 1.5× |


svchb (dismissed):

svchb (approved):

src/io/read_vtk.jl
```diff
@@ -23,11 +23,13 @@ ic = vtk2trixi(joinpath("out", "rectangular.vtu"))
...
-│ InitialCondition{Float64}                                                                        │
-│ ═════════════════════════                                                                        │
+│ InitialCondition                                                                                 │
+│ ════════════════                                                                                 │
 │ #dimensions: ……………………………………………… 2                                                                │
 │ #particles: ………………………………………………… 100                                                              │
 │ particle spacing: ………………………………… 0.1                                                              │
+│ eltype: …………………………………………………………… Float64                                                          │
+│ coordinate eltype: ……………………………… Float64                                                          │
 └──────────────────────────────────────────────────────────────────────────────────────────────────┘
```
LasNikas:
```suggestion
│ eltype: …………………………………………………………… Float64                                                          │
│ coordinate eltype: ……………………………… Float64                                                          │
└──────────────────────────────────────────────────────────────────────────────────────────────────┘
```
```

"```" are missing

File: src/setups/rectangular_shape.jl [hunk 2]
```diff
@@ -3,12 +3,14 @@
                      velocity=zeros(length(n_particles_per_dimension)),
                      mass=nothing, density=nothing, pressure=0.0,
                      acceleration=nothing, state_equation=nothing,
-                     place_on_shell=false, loop_order=nothing)
+                     place_on_shell=false, coordinates_eltype=Float64,
+                     coordinates_perturbation=nothing)
```
LasNikas:
```suggestion
                     place_on_shell=false, coordinates_eltype=Float64,
                     coordinates_perturbation=nothing, loop_order=nothing)
```

LasNikas (changes requested):
I'm, wondering if we want the `ELTYPE` and coordinates eltype info in the show box for `Semidiscretization`, too?

src/preprocessing/geometries/geometries.jl
```diff
@@ -12,7 +12,8 @@ function Base.setdiff(initial_condition::InitialCondition,
         throw(ArgumentError("all passed geometries must have the same dimensionality as the initial condition"))
     end
 
-    coords = reinterpret(reshape, SVector{ndims(geometry), eltype(geometry)},
+    coords = reinterpret(reshape,
+                         SVector{ndims(geometry), eltype(initial_condition.coordinates)},
```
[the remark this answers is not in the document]
efaulhaber: No. We `reinterpret` the array `initial_condition.coordinates`, so in order to get correct results, it is very important that we use the element type of this array. Using a function like `coordinates_eltype` here is dangerous, as it could lead to errors just like the previous `eltype(geometry)` did.

PR #959, opened 2025-10-17
Revise `vtk2trixi` so that it returns a `NamedTuple` instead of an `InitialCondition`

marcelschurer:
This PR changes the return type of the `vtk2trixi` function from `InitialCondition` to a NamedTuple and adds several other important quantities.


svchb:
@LasNikas Didn't you have code somewhere to do a restart via vtk2trixi? If yes it still needs to be updated.

LasNikas:
> @LasNikas Didn't you have code somewhere to do a restart via vtk2trixi? If yes it still needs to be updated.

@svchb No, but it is still a breaking change because the return type has changed.

src/io/read_vtk.jl
```diff
@@ -1,10 +1,15 @@
 """
-	vtk2trixi(file::String)
+	vtk2trixi(file::String; custom_quantities...)
 
-Load VTK file and convert data to an [`InitialCondition`](@ref).
+Load VTK file and convert data to a NamedTuple.
```
LasNikas:
```suggestion
Load VTK file and convert data to a `NamedTuple`.
```

src/io/read_vtk.jl
```diff
@@ -53,35 +60,56 @@ function vtk2trixi(file; element_type=nothing, coordinates_eltype=nothing)
...
+
+    results = NamedTuple(results)
+
+    if create_initial_condition
+        ic = InitialCondition(; coordinates=results.coordinates,
+                              particle_spacing=results.particle_spacing,
+                              velocity=results.velocity, density=results.density,
+                              pressure=results.pressure)
+
+        return merge(results, (initial_condition=ic,))
```
```suggestion
        return (; results..., initial_condition=ic)
```

src/io/read_vtk.jl
```diff
@@ -1,7 +1,10 @@
 """
 	vtk2trixi(file::String; element_type=nothing, coordinates_eltype=nothing)
 
-Load VTK file and convert data to an [`InitialCondition`](@ref).
+Read a VTK file and return a `NamedTuple` with keys
+`:coordinates`, `:velocity`, `:density`, `:pressure`, `:particle_spacing`, `:time`,
+plus any custom quantities.
```
efaulhaber: Please mention `:initial_condition` as well.

PR #967, opened 2025-10-23
Fix `extrude_geometry`

### LINKED ISSUE #965: `extrude_geometry` auto-adjusts the `particle_spacing`
The method splits the segment between the two endpoints into an integer number of equal intervals. That yields a slightly adjusted spacing whenever the desired spacing does not divide the segment exactly.
While `extrude_geometry` logs an info message when the spacing is changed,  the actual spacing deviates from the user-requested value, which is not desired I think.

LasNikas:
resolves #965 

The `particle_spacing` is now kept unchanged.
Instead of adjusting the spacing, we round the geometric length to a multiple of the spacing and move the endpoints if necessary.
This is consistent with the `RectangularTank`.

### DIFF
--- a/src/general/semidiscretization.jl
+++ b/src/general/semidiscretization.jl
@@ -826,6 +826,26 @@ function update_nhs!(neighborhood_search,
 end
 
 function update_nhs!(neighborhood_search,
+                     system::OpenBoundarySystem{<:BoundaryModelDynamicalPressureZhang},
+                     neighbor::TotalLagrangianSPHSystem,
+                     u_system, u_neighbor, semi)
+    update!(neighborhood_search,
+            current_coordinates(u_system, system),
+            current_coordinates(u_neighbor, neighbor),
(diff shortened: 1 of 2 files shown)


src/setups/extrude_geometry.jl
```diff
@@ -166,12 +159,19 @@ function sample_plane(plane_points::NTuple{2}, particle_spacing; place_on_shell=
...
+    p2 = plane_points[2]
+
+    line_dir = p2 - p1
+    line_length = norm(line_dir)
 
-    coords = stack(range(plane_points[1], plane_points[2], length=n_points))
-    particle_spacing_new = norm(coords[:, 1] - coords[:, 2])
+    n_particles,
+    new_length = round_n_particles(line_length, particle_spacing,
+                                   "line length")
```
LasNikas: This is from `RectangulaTank`

efaulhaber (changes requested):

PR #982, opened 2025-11-17
Add IntervalSets to dep and compat sections

### LINKED ISSUE #978: The validation case oscillating_beam_2d is failing
this seems to be related https://github.com/JuliaMath/IntervalSets.jl/pull/205

svchb:
Similar to https://github.com/trixi-framework/Trixi.jl/pull/2649 but since the fix will be breaking for IntervalSets we will allow future versions with the fix as well

fixes #978

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -14,7 +14,6 @@ FastPow = "c0e83750-1142-43a8-81cf-6c956b72b4d1"
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 GPUArraysCore = "46192b85-c4d5-4398-a991-12ede77f4527"
-IntervalSets = "8197267c-284f-5f27-9208-e0e47529a953"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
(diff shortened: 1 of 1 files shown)


Project.toml
```diff
@@ -50,6 +51,7 @@ FastPow = "0.1"
 FileIO = "1"
 ForwardDiff = "1"
 GPUArraysCore = "0.2"
+IntervalSets = "=0.7.11, >=0.8"
```
efaulhaber: Who says it will work with 0.8? I would prefer the same as in Trixi.jl:

```suggestion
IntervalSets = "0.7 - 0.7.11"
```
Once 0.8 is out, we will get a CompatHelper PR anyway.

Project.toml
```diff
@@ -14,6 +14,7 @@ FastPow = "c0e83750-1142-43a8-81cf-6c956b72b4d1"
 FileIO = "5789e2e9-d7fb-5bc7-8068-2c6fae9b9549"
 ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"
 GPUArraysCore = "46192b85-c4d5-4398-a991-12ede77f4527"
+IntervalSets = "8197267c-284f-5f27-9208-e0e47529a953"
```
efaulhaber: Can't we just add it to the `[weakdeps]` since it's not a direct dependency?

efaulhaber (approved):
Let's see if this works with weakdeps 👍

PR #988, opened 2025-11-19
Delta SPH - Boundary Condition by Marrone et al.

RubberLanding:
This boundary condition computes the normal vectors for the dummy particles. These vectors are then used to compute certain interpolation points inside the fluid. At these interpolation points, dummy particle properties (e.g. pressure, etc.) are approximated using a Moving Least-Squares kernel.


RubberLanding:
@efaulhaber Please have a look.

svchb (commented):
For compatibility with the surface tension code the normals of the bnd should be added into the cache of the boundary system.

PR #989, opened 2025-11-20
enhancing falling 3d sphere tutorial to two falling spheres in 3d (Work in Pro…

Jordan0065:
…gress)


svchb:
If you want to design a tutorial the best thing is to start by thinking about what kind of information one wants to communicate to the person reading it. In this case the goal of the tutorial should be to teach a certain new aspect of a TrixiParticles. See the current tutorials for reference. The best thing is probably you come up with an outline for the tutorial first.

PR #991, opened 2025-11-22
Add custom element type support for `vtk2trixi`

LasNikas:


svchb (approved):

src/io/read_vtk.jl
```diff
@@ -31,9 +31,11 @@ ic = vtk2trixi(joinpath("out", "rectangular.vtu"))
 │ eltype: …………………………………………………………… Float64                                                          │
 │ coordinate eltype: ……………………………… Float64                                                          │
 └──────────────────────────────────────────────────────────────────────────────────────────────────┘
-````
+```
 """
-function vtk2trixi(file)
+function vtk2trixi(file; element_type=Float64, coordinates_eltype=Float64)
```
efaulhaber: Why is the default not the type that was stored?

src/io/read_vtk.jl
```diff
@@ -31,18 +36,22 @@ ic = vtk2trixi(joinpath("out", "rectangular.vtu"))
 │ eltype: …………………………………………………………… Float64                                                          │
 │ coordinate eltype: ……………………………… Float64                                                          │
 └──────────────────────────────────────────────────────────────────────────────────────────────────┘
-````
+```
 """
-function vtk2trixi(file)
+function vtk2trixi(file; element_type=:default, coordinates_eltype=Float64)
```
```suggestion
function vtk2trixi(file; element_type=nothing, coordinates_eltype=Float64)
```
Is there a reason why you need to use a "magic value" symbol instead of just `nothing`? We always preferred `nothing` in these cases.

src/io/read_vtk.jl
```diff
@@ -1,11 +1,16 @@
...
 
 Load VTK file and convert data to an [`InitialCondition`](@ref).
 
 # Arguments
 - `file`: Name of the VTK file to be loaded.
 
+# Keywords
+- `element_type`: Element type for particle fields (`:default` keeps the type
+    stored in the VTK file, otherwise converted to the given type).
+- `coordinates_eltype`: Element type for particle coordinates (defaults to `Float64`).
```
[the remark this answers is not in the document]
LasNikas: Didn't we decide that the default type of coordinates should be Float64?

src/io/read_vtk.jl
```diff
@@ -38,16 +41,16 @@ ic = vtk2trixi(joinpath("out", "rectangular.vtu"))
...
 
     # Retrieve data fields (e.g., pressure, velocity, ...)
     point_data = ReadVTK.get_point_data(vtk_file)
     field_data = ReadVTK.get_field_data(vtk_file)
     point_coords = ReadVTK.get_points(vtk_file)
 
-    cELTYPE = coordinates_eltype
-    ELTYPE = element_type === :default ? eltype(point_coords) : element_type
+    cELTYPE = isnothing(coordinates_eltype) ? eltype(point_coords) : coordinates_eltype
+    ELTYPE = isnothing(element_type) ? eltype(point_coords) : element_type
```
efaulhaber: Now both have the stored coordinates type by default. Please also add a test for a different coordinates eltype.

efaulhaber (changes requested):

PR #997, opened 2025-11-26
More robust open boundaries I

LasNikas:
For geometries that expand downstream of the inflow, particle conversion may occasionally fail.
This PR provides a fix for this issue.

In rare cases (likely when using relatively large timesteps), an inflow particle may lie outside the outer edge of the boundary zone at the moment of conversion. This can happen when the fluid domain widens rapidly.
If such a particle is reset upstream during `convert_particle!`, it may be shifted outside the boundary zone.
Because of this, it cannot be assigned to any boundary zone, yet it remains active—leading to an assertion failure.

The screenshots below show one of these cases.
Left: The red region is the fluid domain after the transition face (blue). The yellow particle is the inflow particle that will be incorrectly reset in the next timestep.

Right: The fluid domain (red) and the inflow zone (blue) are shown again. The geometry expands significantly.

A schematic illustration of this situation is shown underneath.

| Last frame before error |  inflow expansion | 
|------------------------|-------------------|
image |  image | 

image

### DIFF
--- a/src/schemes/boundary/open_boundary/system.jl
+++ b/src/schemes/boundary/open_boundary/system.jl
@@ -421,7 +421,7 @@ end
     # Activate a new particle in simulation domain
     transfer_particle!(fluid_system, system, particle, particle_new, v_fluid, u_fluid, v, u)
 
-    # Move the boundary particle to a well-defined position inside the boundary zone.
+    # Reset position of boundary particle back to the beginning of the boundary zone.
     # If we translated it by exactly `zone_width` along `-face_normal`, rounding
     # errors could place it just outside the zone. To avoid this, use a slightly
     # shorter distance (`zone_width - eps(zone_width)`), which guarantees the final


src/schemes/boundary/open_boundary/system.jl
```diff
@@ -421,9 +421,22 @@ end
...
+        u[dim, particle] += reset_vector[dim]
+    end
+
+    # Verify the particle remains inside the boundary zone after the reset; deactivate it if not.
+    particle_coords = current_coords(u, system, particle)
+    if !is_in_boundary_zone(boundary_zone, particle_coords)
+        deactivate_particle!(system, particle, u)
+
+        return system
     end
```
LasNikas: should we add a `@warn "deactivated particle: $particle"` ?

efaulhaber: Will it be activated again? Or did we just lose a particle? If this happens 100 times, do we lose 100 particles from the simulation?

efaulhaber (changes requested):

PR #1000, opened 2025-11-28
Deactivate out of bounds particle

LasNikas:


src/general/neighborhood_search.jl
```diff
@@ -9,3 +9,22 @@ function PointNeighbors.foreach_point_neighbor(f, system, neighbor_system,
...
+
+        if !all(min_corner .<= particle_position .<= max_corner)
+            deactivate_particle!(system, particle, u)
+        end
+    end
+
+    update_system_buffer!(system.buffer, semi)
+
+    return system
+end
```
LasNikas: @efaulhaber Is it possible to add a `update_flag` to the gpu kernel when a particle is deactivated? 
.. and then `update_flag && update_system_buffer!(system.buffer, semi)`
Otherwise, the system buffer will be updated even when no particles are deactivated.

svchb:
PR 1000 🎉🔥

LasNikas:
> I think it should be optional. Otherwise I am fine with this.

I also considered making this optional. However, it would then be necessary to explain that this can only be set if a full grid cell list and a system buffer are used. Users should not expect this to apply to other constellations. This would need to be explained in more detail. Should I do that? @svchb @efaulhaber

efaulhaber (changes requested):

efaulhaber:
Tests are failing.

src/general/neighborhood_search.jl
```diff
@@ -11,6 +11,40 @@ function PointNeighbors.foreach_point_neighbor(f, system, neighbor_system,
...
+                                             cell_list::FullGridCellList, v, u, semi)
+    @trixi_timeit timer() "deactivate out of bounds particle" begin
+        # Remove the padding layer (see comment in PointNeighbors: full_grid.jl line 60)
+        min_corner = cell_list.min_corner .+ 1001 // 1000 .* nhs.cell_size
+        max_corner = cell_list.max_corner .- 1001 // 1000 .* nhs.cell_size
+
+        @threaded semi for particle in each_integrated_particle(system)
+            particle_position = current_coords(u, system, particle)
+
+            if !all(min_corner .<= particle_position .<= max_corner)
```
efaulhaber:
```suggestion
        @threaded semi for particle in each_integrated_particle(system)
            particle_position = current_coords(u, system, particle)
            cell = PointNeighbors.cell_coords(particle_position, nhs)

            if !all(cell[i] in 2:(size(cell_list.linear_indices, i) - 1) for i in eachindex(cell))
```
I think this is the better option. This is the exact check that is used in PointNeighbors.jl when adding particles to the cell list. Then we are not relying on the exact size of the padding that is added.

efaulhaber (changes requested):
Please add more comments explaining why this is done and when. Something like "the full grid cell list requires a bounding box. For this cell list, we deactivate particles that are outside the bounding box. These particles are glitched out of the open boundary and don't contribute anyway, and would crash the simulation if they move outside the bounding box and are not removed.

src/general/neighborhood_search.jl
```diff
@@ -11,6 +11,41 @@ function PointNeighbors.foreach_point_neighbor(f, system, neighbor_system,
...
+# This function deactivates particles that move outside this bounding box to prevent
+# simulation crashes.
+# Note that simply deactivating particles is only possible in combination with a 'SystemBuffer'.
+function deactivate_out_of_bounds_particles!(system, ::SystemBuffer, nhs,
+                                             cell_list::FullGridCellList, v, u, semi)
+    @trixi_timeit timer() "deactivate out of bounds particle" begin
+        @threaded semi for particle in each_integrated_particle(system)
+            particle_position = current_coords(u, system, particle)
+            cell = PointNeighbors.cell_coords(particle_position, nhs)
+            if !all(cell[i] in 2:(size(cell_list.linear_indices, i) - 1)
```
efaulhaber, line 38:
```suggestion
            # This is the same code as is used in PointNeighbors.jl in `check_cell_bounds`.
            # It tests that particles are inside the inner grid (without the padding layer for neighbor query).
            if !all(cell[i] in 2:(size(cell_list.linear_indices, i) - 1)
```

efaulhaber (changes requested):

efaulhaber (dismissed):

src/general/neighborhood_search.jl
```diff
@@ -11,6 +11,44 @@ function PointNeighbors.foreach_point_neighbor(f, system, neighbor_system,
...
+                                             nhs::GridNeighborhoodSearch, v, u, semi)
+    deactivate_out_of_bounds_particles!(system, buffer, nhs, nhs.cell_list, v, u, semi)
+end
+
+function deactivate_out_of_bounds_particles!(system, buffer, nhs, cell_list, v, u, semi)
+    return system
+end
+
+# `GridNeighborhoodSearch` with a `FullGridCellList` requires a bounding box.
+# This function deactivates particles that move outside this bounding box to prevent
```
svchb: the bounding box

src/general/neighborhood_search.jl
```diff
@@ -11,6 +11,44 @@ function PointNeighbors.foreach_point_neighbor(f, system, neighbor_system,
...
+# simulation crashes.
+# Note that simply deactivating particles is only possible in combination with a 'SystemBuffer'.
```
svchb: Unclear what does simply mean? Why does it only work with a systembuffer

svchb (changes requested):

src/general/neighborhood_search.jl
LasNikas: simply means: ` active_particle[particle] = false` which is only possible with a `SystemBuffer`

efaulhaber: Mabye just remove "simply"?

efaulhaber: They cannot be disabled at all without a buffer.

svchb: Yes I meant adjusting the comment.

efaulhaber (approved):

svchb (approved):

PR #998, opened 2025-11-26
More robust open boundaries II

LasNikas:
This PR addresses in some way the opposite of the issue in #997 . 
In this case, the domain narrows in the downstream direction (see the blue fluid domain in the figure below; green indicates the open-boundary particles and yellow marks the particle leaving the zone). 
As soon as the yellow particle exits the domain, the simulation aborts.
(Note that this starts happening in a simulation after 54200 time steps (dt ~4e-5) with 415k particles)

| t = 0.62 | t = 0.64  | t = 0.66 
|----------|-----------|----------
| image  | image | image

Due to the strongly curved inlet geometry, inflow particles are likely being re-initialized at very small steps. In other words, the high curvature at the inlet creates a congestion region where newly populated inflow particles at the free surface region are generated too densely. The particle shifting—which could normally mitigate this—only activates beyond a certain distance from the free surface.

Why does the yellow particle leave the domain at all?
Its density is not abnormal. However, the local particle distribution in that region likely leads to an incorrect extrapolated pressure. This becomes visible in the top view of the clipped boundary (see figures below).We can clearly see where the shifting begins to take effect.

| total view | zoom | coarser resolution 
| ----- | ------| -----
| image | image | image

TVF (or particle shifting) is not aligned with the boundary zone (by design). The particle probably experiences a strong reordering potential (caused by the problem described above). Consequently, the boundary pressure is insufficient to keep this particle inside the zone, since the correction force acting on it is too large.

In summary, the only feasible solution seems to be deactivating such particles.

PR #1011, opened 2025-12-05
Ramp shifting velocity near the free surface

LasNikas:
This is a better solution for the issue described in #998. As we can see in the figure below, particles remain ordered with this PR (right panel), while the particles in the left panel (#998) begin to leave the zone.

Do we need the PR #998 anyway? @efaulhaber @svchb ?

image

Note that this reduces the runtime of a 3 sec simulation by 2.4x because less time steps are required.
For 3 sec sim. time:
PR #998 | This PR
---|----
  #timesteps 89950 | #timesteps 48750
run time 18.18 h | run time 7.5h 

Also note the boundary pressure:
PR #998 | This PR
----------|-------
image | image

### DIFF
--- a/src/schemes/boundary/open_boundary/system.jl
+++ b/src/schemes/boundary/open_boundary/system.jl
@@ -539,7 +539,7 @@ end
 
         if dist_free_surface < compact_support(fluid_system, fluid_system)
             # Ramp shifting velocity near the free surface using a kernel-weighted transition.
-            # According to our experiment, alternative approaches lead to particle disorder:
+            # According to our experiments, the proposed alternative approaches lead to particle disorder:
             # - Sun et al. 2017: only use surface-tangential component
             # - Zhang et al. 2025: disable shifting entirely
             kernel_max = smoothing_kernel(system, 0, particle)


src/schemes/boundary/open_boundary/system.jl
```diff
@@ -538,13 +538,18 @@ end
         dist_free_surface = boundary_zone.zone_width - dist_to_transition
 
         if dist_free_surface < compact_support(fluid_system, fluid_system)
-            # Disable shifting for this particle.
-            # Note that Sun et al. 2017 propose a more sophisticated approach with a transition phase
-            # where only the component orthogonal to the surface normal is kept and the tangential
-            # component is set to zero. However, we assume laminar flow in the boundary zone,
-            # so we simply disable shifting completely.
+            # Ramp shifting velocity near the free surface using a kernel-weighted transition.
+            # According to our experiment, alternative approaches lead to particle disorder:
```
svchb:
```suggestion
            # According to our experiments, the proposed alternative approaches lead to particle disorder:
```

svchb (approved):

PR #1024, opened 2025-12-15
Test timestep selection and add capillary interface condition

svchb:
## Summary

- Add unit coverage for `StepsizeCallback`, per-system time-step selection, and semidiscretization-level reduction.
- Evaluate interface-specific restrictions once when the constant stepsize callback is initialized.
- Use the Brackbill/Morris capillary restriction for interactions between two fluid systems using the same Morris surface-tension model.
- Remove the unsupported interface acoustic and summed-viscosity heuristics.

## Interface condition

For a resolved interface with surface tension, the capillary restriction is

```math
\Delta t_\sigma = \sqrt{\frac{(\rho_a + \rho_b) h^3}{4\pi\sigma}}
                   = \sqrt{\frac{\bar{\rho} h^3}{2\pi\sigma}},
\qquad \bar{\rho}=\frac{\rho_a+\rho_b}{2}.
```

This is the capillary time-step restriction of Brackbill, Kothe, and Zemach (1992), also used by Morris (2000). For systems with different resolutions or coefficients, the implementation conservatively uses the smaller smoothing length and larger surface-tension coefficient.

The pairwise restriction is only evaluated when both systems use the same `SurfaceTensionMorris` or `SurfaceTensionMomentumMorris` model, matching the force dispatch. All other pairs return `Inf` without accessing particle state. The pair traversal itself runs only once because the current stepsize callback is constant.


efaulhaber (changes requested):

src/schemes/fluid/fluid.jl
```diff
@@ -237,6 +237,38 @@ function calculate_dt(v_ode, u_ode, cfl_number, system::AbstractFluidSystem, sem
...
+                                semi)
+    h_interface = harmonic_mean(initial_smoothing_length(system),
+                                initial_smoothing_length(neighbor_system))
+
+    c_system = system_sound_speed(system)
+    c_neighbor_system = system_sound_speed(neighbor_system)
+
+    # Adami et al. (2012) suggest that the stiffest interface waves travel with
+    # an impedance-limited signal speed; using the harmonic mean of both phase
+    # sound speeds penalizes strong contrasts already at t = 0.
```
efaulhaber: Where do they say that? I don't find that in the paper. They don't mention any multiphase CFL conditions. Where does that come from?

src/general/semidiscretization.jl
```diff
@@ -405,13 +405,51 @@ end
...
+end
+
+@inline should_calculate_interface_dt(system, neighbor_system) = true
+
+@inline function should_calculate_interface_dt(system::WeaklyCompressibleSPHSystem,
+                                               neighbor_system::WeaklyCompressibleSPHSystem)
+    rho_system = system.state_equation.reference_density
+    rho_neighbor_system = neighbor_system.state_equation.reference_density
+    return !isapprox(rho_system, rho_neighbor_system)
+end
```
efaulhaber: That is super confusing. So it should calculate for all systems except two WCSPH systems with the same density, but then the calculation is skipped for everything but two fluid systems. Why are there two places in which it is decided if interface dt should be calculated? That is redundant. Why is this not skipped for other two fluid systems with the same density?

efaulhaber (changes requested):
I think there is a lot of information missing here. This PR adds a substantial amount of new code, without stating why we need different CFL conditions for multiphase (example simulations? Benchmarks?) or which paper it is based on.

src/schemes/fluid/fluid.jl
svchb: I think they cite it from a Colagrossi 2003 paper. I have revised that section.

src/general/semidiscretization.jl
svchb: I simplified the condition.

src/general/semidiscretization.jl
```diff
@@ -474,7 +474,29 @@ end
...
+        for i in 1:(nsystems - 1)
+            system = systems[i]
+            for j in (i + 1):nsystems
+                neighbor_system = systems[j]
+                dt_interfaces = min(dt_interfaces,
+                                    calculate_interface_dt(v_ode, u_ode, cfl_number,
+                                                           system, neighbor_system, semi))
+            end
+        end
+    end
```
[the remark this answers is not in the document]
svchb: Agreed. I reworked this in 0c91804b9. The broad acoustic and viscous interface calculations are gone. The remaining pair traversal is used only for the capillary restriction and runs once when the constant `StepsizeCallback` is initialized, not during every time step. Pairs that do not use the same Morris surface-tension model dispatch directly to `Inf` before any particle state is wrapped or accessed. I therefore kept automatic detection instead of adding a callback kwarg for a required stability condition.

src/schemes/fluid/fluid.jl
svchb: You are right: Adami et al. (2012) do not state the interface acoustic condition, and Antuono et al. (2015) do not derive the summed-viscosity condition. I removed both and their attributions in 0c91804b9. The harmonic-mean acoustic expression is a weighted average of the two per-system `h/c` ratios, so it can never be smaller than the existing per-system minimum. The summed-viscosity expression was also inconsistent with the harmonic dynamic-viscosity average used by the pairwise discretization and penalized identical fluids by a factor of two. The only remaining interface-specific condition is the capillary restriction `sqrt((rho_a + rho_b) h^3 / (4 pi sigma))`, now documented and cited to Brackbill et al. (1992) and Morris (2000).

src/general/semidiscretization.jl
svchb: Agreed. I deleted `should_calculate_interface_dt` and the WCSPH-only density special case in 0c91804b9. There is now one dispatch point: the generic interface method returns `Inf`, while two fluid systems using the same Morris surface-tension model evaluate the Brackbill/Morris capillary restriction. Equal reference density is not used as a gate because equal-density immiscible fluids can still have a capillary interface; the active surface-tension interaction is the relevant criterion.

test/schemes/fluid/calculate_dt.jl
```diff
@@ -0,0 +1,96 @@
...
+        viscosity = TestViscosity(1.0e-6)
+        acceleration = (0.0, 9.81)
+        no_surface_tension = TestFluidSystem(0.01, 343.0, viscosity, acceleration, nothing)
+        morris = SurfaceTensionMorris(surface_tension_coefficient=0.072)
+        morris_momentum = SurfaceTensionMomentumMorris(surface_tension_coefficient=0.072)
+        with_morris = TestFluidSystem(0.012, 1482.0, viscosity, acceleration, morris)
+        with_morris_momentum = TestFluidSystem(0.014, 1482.0, viscosity, acceleration,
+                                               morris_momentum)
+
+        # No inter-system surface tension force is evaluated unless both systems use
```
efaulhaber:
```suggestion
        # No inter-system surface tension dt is evaluated unless both systems use
```

test/callbacks/stepsize.jl
```diff
@@ -14,4 +60,55 @@
         └──────────────────────────────────────────────────────────────────────────────────────────────────┘"""
         @test repr("text/plain", callback) == show_box
     end
+
+    @testset verbose=true "condition" begin
+        # Ensure the discrete callback only triggers for non-constant stepsize on
+        # non-adaptive integrators
```
efaulhaber:
```suggestion
        # non-adaptive integrators.
```

test/callbacks/stepsize.jl
```diff
@@ -1,5 +1,51 @@
...
+    factor::Float64
+    offset::Float64
+    update_callback_used::Base.RefValue{Bool}
+    integrate_tlsph::Base.RefValue{Bool}
+end
+
+mutable struct StepsizeTestOpts
+    adaptive::Bool
+    dtmax::Float64
+    callback::Any
```
efaulhaber, line 14: Never used.

src/general/semidiscretization.jl
```diff
@@ -441,13 +441,43 @@ end
...
 
-    return minimum(systems) do system
+    dt_systems = minimum(systems) do system
         if system isa TotalLagrangianSPHSystem && !semi.integrate_tlsph[]
             # Skip TLSPH systems if they are not integrated
             return Inf
         end
         return calculate_dt(v_ode, u_ode, cfl_number, system, semi)
     end
+
```
efaulhaber, line 520:
```suggestion

# Calculate multiphase dt. This is only relevant in multiphase simulations.
```

src/general/semidiscretization.jl
```diff
@@ -441,13 +441,43 @@ end
...
+    dt_interfaces = Inf
+    nsystems = length(systems)
+
+    # This is evaluated only once when the constant stepsize callback is initialized.
+    # Pairs without an interface-specific restriction return `Inf` through dispatch.
```
efaulhaber, line 527:
```suggestion
    # Pairs without an interface-specific restriction return `Inf` through dispatch.
    # TODO avoid recomputation when implementing adaptive stepsize callback
```

efaulhaber (changes requested):
Please add an entry to NEWS.md.

test/callbacks/stepsize.jl
svchb: used by StepsizeTestIntegrator

src/general/semidiscretization.jl
svchb: This is now a dt which currently only influences dt step size when surface tension is used.

efaulhaber (dismissed):

efaulhaber (approved):

PR #1036, opened 2026-01-06
Compute Normal Vectors for Boundary Particles

RubberLanding:
Compute Normal Vectors for Boundary Particles. Used in the Boundary Condition by Marrone et al.


src/setups/sphere_shape.jl
```diff
@@ -431,3 +434,25 @@ function round_sphere(sphere, particle_spacing, radius, center::SVector{3})
...
+    for i in 1:n_points
+        point = view(coordinates, :, i)
+        diff = point - center_position
+
+        dist = norm(diff)
+        if dist > threshold
+            normals[:, i] .= center_position + radius * (diff / dist)
+        else
+            normals[:, i] .= zero(T)
+        end
```
efaulhaber: 1. Use `dist > eps(radius)`. `eps(T)` is problematic because it is independent of the scale that is used. We could simulate something in the micrometer range, and then `eps(Float32)` would be large compared to the sphere. `eps(radius)` is relative to the radius, so it takes the scale of the problem into account.
2. You can skip the `else`, it's already initialized as zero.

efaulhaber (changes requested):
I haven't looked at the rectangular tank stuff yet.

src/general/initial_condition.jl
```diff
@@ -98,21 +98,30 @@ struct InitialCondition{ELTYPE, C, MATRIX, VECTOR}
...
+                                   pressure, particle_spacing, normals)
 end
 
-# Function barrier to make `NDIMS` static and therefore SVectors type-stable
+# Wrapper function for outer constructor without `normals`-keyword
 function InitialCondition{NDIMS}(coordinates, velocity, mass, density,
                                  pressure, particle_spacing) where {NDIMS}
+    return InitialCondition{NDIMS}(coordinates, velocity, mass, density,
+                                   pressure, particle_spacing, nothing)
+end
```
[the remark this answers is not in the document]
RubberLanding: I wrote the wrapper because else, some initializations of initial conditions fail, because they call the inner constructor.

src/general/initial_condition.jl
```diff
@@ -102,21 +102,23 @@ struct InitialCondition{ELTYPE, C, MATRIX, VECTOR}
     mass             :: VECTOR # Array{ELTYPE, 1}
     density          :: VECTOR # Array{ELTYPE, 1}
     pressure         :: VECTOR # Array{ELTYPE, 1}
+    normals          :: Union{Nothing, MATRIX}
```
efaulhaber: This is not type-stable. Change it to `::N` and make `N` a type parameter.

src/general/initial_condition.jl
```diff
@@ -201,8 +204,15 @@ function InitialCondition{NDIMS}(coordinates, velocity, mass, density,
         masses = mass_fun.(coordinates_svector)
     end
 
+    if normals isa AbstractMatrix
```
efaulhaber:
```suggestion
    if !isnothing(normals)
```
And this check has to be before the unchecked `normals` are used above in the 0 particles case.

src/setups/sphere_shape.jl
```diff
@@ -431,3 +434,23 @@ function round_sphere(sphere, particle_spacing, radius, center::SVector{3})
...
+    normals = zeros(T, NDIMS, n_points)
+
+    threshold = eps(radius)
+    for i in axes(coordinates, 2)
+        point = extract_svector(coordinates, Val(NDIMS), i)
+        diff = point - center_position
+        dist = norm(diff)
+
+        if dist > threshold
+            normals[:, i] .= center_position + radius * (diff / dist)
```
efaulhaber: I don't understand how the center position belongs here. 
The unit normal is `diff / dist`, and we want to scale it by the distance to the surface of the sphere, which is `radius - dist`, so shouldn't the whole thing be

```suggestion
            normals[:, i] .= diff / dist * (radius - dist)
```

src/setups/rectangular_tank.jl
```diff
@@ -154,10 +156,14 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
         fluid_size_ = check_tank_overlap(fluid_size_, tank_size_,
                                          particle_spacing, n_particles_per_dim)
 
+        normals = zeros(ELTYPE, size(boundary_coordinates))
+        calculate_normals!(normals, boundary_coordinates, boundary_spacing,
+                           face_indices, boundary_indices, faces)
```
efaulhaber:
```suggestion
                           face_indices, boundary_indices, faces, Val(NDIMS))
```
Then you don't need the wrapper function.

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,215 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+                        boundary_indices,
+                        faces,
+                        Val(size(boundary_coordinates, 1)))
+end
+
+# 2D
+function _calculate_normals!(normals, boundary_coordinates, boundary_spacing,
+                             face_indices, boundary_indices, faces, ::Val{2})
+    corner_indices, = boundary_indices
+    face_indices = Tuple(vec(x) for x in face_indices)
```
efaulhaber: Why do you need this line? Doesn't everything work with the `face_indices` as passed to this function?
Also, this is not type-stable, as it changes the type of `face_indices`.

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,215 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+    # any particles associated with it. 
+    function face_has_particles(i)
+        return faces[i] && !isempty(face_indices[i])
+    end
+
+    #### Left boundary
+    if face_has_particles(1)
+        left_boundary = maximum(boundary_coordinates[1, face_indices[1]]) + offset
+        for idx in face_indices[1]
+            normals[1, idx] = -abs(boundary_coordinates[1, idx] - left_boundary)
```
efaulhaber:
```suggestion
            normals[1, idx] = boundary_coordinates[1, idx] - left_boundary
```
Same for the other boundaries below. The boundary particles are always left of the left boundary.

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,215 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+        for idx in face_indices[4]
+            normals[2, idx] = abs(boundary_coordinates[2, idx] - top_boundary)
+        end
+    end
+
+    # Bottom left corner
+    if face_has_particles(1) && face_has_particles(3)
+        boundary_corner_point = [maximum(boundary_coordinates[1, corner_indices[1]])
+                                 maximum(boundary_coordinates[2, corner_indices[1]])]
+        corner_point = boundary_corner_point + [offset; offset]
```
efaulhaber:
```suggestion
        boundary_corner_point = SVector(maximum(boundary_coordinates[1, corner_indices[1]]),
                                 maximum(boundary_coordinates[2, corner_indices[1]]))
        corner_point = boundary_corner_point + SVector(offset, offset)
```

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,215 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+    coords = (xs, ys, zs)
+
+    # Returns the coordinate of the "virtual wall" interface
+    function get_wall_pos(dim_coords, idxs, side)
+        if side == 1 # Negative side (Left, Bottom, Front)
+            return maximum(view(dim_coords, idxs)) + offset
+        else # Positive side (Right, Top, Back)
+            return minimum(view(dim_coords, idxs)) - offset
+        end
+    end
```
efaulhaber, line 306: There must be a better way to get the exact positions of the wall. Can't you just pass the size of the tank to this function? Same in 2D.

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,215 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+    # are any particles associated with it. 
+    function face_has_particles(idxs, i)
+        return faces[i] && !isempty(idxs[i])
+    end
+
+    # Compute normals for the faces
+    # Faces: 1=Left(-x), 2=Right(+x), 3=Bottom(-y), 4=Top(+y), 5=Front(-z), 6=Back(+z)
+    LEFT, RIGHT = 1, 2
+    BOTTOM, TOP = 3, 4
+    FRONT, BACK = 5, 6
```
efaulhaber: Use lowercase variable names.

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,215 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+                 (FRONT, 3, 1, -1), (BACK, 3, 2, 1))
+
+    for (f_idx, dim, side_type, sign) in face_defs
+        if face_has_particles(face_indices, f_idx)
+            idxs = face_indices[f_idx]
+            wall_pos = get_wall_pos(coords[dim], idxs, side_type)
+            # Compute normals: sign * (abs(coord - wall) + offset)
+            @views normals[dim,
+                           idxs] .= sign .*
+                                    abs.(boundary_coordinates[dim, idxs] .- wall_pos)
```
efaulhaber: Use a loop to avoid allocations.

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,215 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+            d1, s1 = face_defs[f1][2], face_defs[f1][3]
+            d2, s2 = face_defs[f2][2], face_defs[f2][3]
+
+            w1 = get_wall_pos(coords[d1], idxs, s1)
+            w2 = get_wall_pos(coords[d2], idxs, s2)
+
+            # Compute normals
+            # The third dimension of the normal remains 0
+            @views normals[d1, idxs] .= coords[d1][idxs] .- w1
+            @views normals[d2, idxs] .= coords[d2][idxs] .- w2
```
efaulhaber: Use a loop.

src/setups/rectangular_tank.jl
```diff
@@ -415,6 +634,13 @@ function initialize_boundaries(particle_spacing, tank_size::NTuple{2},
                                                       (n_layers, n_layers),
                                                       (layer_offset, layer_offset))
         boundary_coordinates = hcat(boundary_coordinates, bottom_left_corner)
+
+        # store the indices of each particle
```
efaulhaber:
```suggestion
        # Store the indices of each particle
```
Same below.

src/setups/rectangular_tank.jl
```diff
@@ -439,24 +679,62 @@ function initialize_boundaries(particle_spacing, tank_size::NTuple{2},
...
 
-    # Store each particle index
-    face_indices_1 = Array{Int, 2}(undef, n_layers, n_particles_y * n_particles_z)
-    face_indices_2 = Array{Int, 2}(undef, n_layers, n_particles_y * n_particles_z)
-    face_indices_3 = Array{Int, 2}(undef, n_layers, n_particles_x * n_particles_z)
-    face_indices_4 = Array{Int, 2}(undef, n_layers, n_particles_x * n_particles_z)
-    face_indices_5 = Array{Int, 2}(undef, n_layers, n_particles_x * n_particles_y)
-    face_indices_6 = Array{Int, 2}(undef, n_layers, n_particles_x * n_particles_y)
+    # Faces
+    face_indices_LEFT = Array{Int, 2}(undef, n_layers, n_particles_y * n_particles_z)
```
efaulhaber: Please use lowercase variable names.

efaulhaber (changes requested):

src/setups/rectangular_tank.jl
RubberLanding: What exactly is bothering you? Do you have specific changes in mind?

src/general/initial_condition.jl
```diff
@@ -201,8 +204,15 @@ function InitialCondition{NDIMS}(coordinates, velocity, mass, density,
         masses = mass_fun.(coordinates_svector)
     end
 
+    if !isnothing(normals)
+        if size(coordinates) != size(normals)
```
efaulhaber:
```suggestion
    if !isnothing(normals) && size(coordinates) != size(normals)
```

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,210 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+    coords = (xs, ys, zs)
+
+    # Returns the coordinate of the "virtual wall" interface
+    function get_wall_pos(dim_coords, idxs, side)
+        if side == 1 # Negative side (Left, Bottom, Front)
+            return maximum(view(dim_coords, idxs)) + offset
+        else # Positive side (Right, Top, Back)
+            return minimum(view(dim_coords, idxs)) - offset
+        end
+    end
```
efaulhaber, line 306:
```suggestion
    # Return the geometric position of the wall for a given face and the indices
    # of the particles associated with this face.
    function wall_position(face_id, dimension, indices)
        if iseven(face_id) # Positive coordinate direction: right, top, or back
            return minimum(view(boundary_coordinates, dimension, indices)) - offset
        else # Negative coordinate direction: left, bottom, or front
            return maximum(view(boundary_coordinates, dimension, indices)) + offset
        end
    end
```

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,210 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+    for (f_idx, dim, side_type, sign) in face_defs
+        if face_has_particles(face_indices, f_idx)
+            idxs = face_indices[f_idx]
+            wall_pos = get_wall_pos(coords[dim], idxs, side_type)
+            # Compute normals: sign * (abs(coord - wall) + offset)
+            for idx in idxs
+                normals[dim, idx] = sign * abs(boundary_coordinates[dim, idx] - wall_pos)
+            end
+        end
+    end
```
efaulhaber, line 334:
```suggestion
    function set_intersection_normals!(indices, face_ids)
        # Do nothing if no particles belong to these faces
        # or if the intersection doesn't exist (not all faces of it activated).
        isempty(indices) && return
        all(faces[face_ids]) || return

        for face_id in face_ids
            dimension = div(face_id, 2 RoundUp())
            wall = wall_position(face_id, dimension, indices)
            for index in indices
                normals[dimension, index] = boundary_coordinates[dimension, index] - wall
            end
        end
    end
    
    for (face_id, indices) in enumerate(face_indices)
        set_intersection_normals!(indices, (face_id,))
    end
```
This function can then be reused for edges and corners as well (see below).

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,210 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+            w2 = get_wall_pos(coords[d2], idxs, s2)
+
+            # Compute normals
+            # The third dimension of the normal remains 0
+            for idx in idxs
+                normals[d1, idx] = coords[d1][idx] - w1
+                normals[d2, idx] = coords[d2][idx] - w2
+            end
+        end
+    end
```
efaulhaber, line 368:
```suggestion
    for (indices, face_ids) in zip(edge_indices, edge_faces)
        set_intersection_normals!(indices, face_ids)
    end
```

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,210 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+                normals[dim, idx] = sign * abs(boundary_coordinates[dim, idx] - wall_pos)
+            end
+        end
+    end
+
+    # Compute normals for the edges
+    # Each edge is the intersection of 2 faces.
+    # Iterate 1..12 assuming edge_indices are ordered: 1-4: Z-aligned, 5-8: Y-aligned, 9-12: X-aligned
+
+    edge_defs = (
```
efaulhaber, line 340:
```suggestion
    # The order matches `edge_indices` from `initialize_boundaries` (important for `zip`).
    edge_faces = (
```

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,210 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+            for idx in idxs
+                normals[d1, idx] = coords[d1][idx] - w1
+                normals[d2, idx] = coords[d2][idx] - w2
+            end
+        end
+    end
+
+    # Compute normals for the corners
+    # Each corner is the intersection of 3 faces
+    corner_defs = ((left, bottom, front), (left, bottom, back),
```
efaulhaber, line 372:
```suggestion
    # The order matches `corner_indices` from `initialize_boundaries` (important for `zip`).
    corner_faces = ((left, bottom, front), (left, bottom, back),
```

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,210 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+
+            # Compute normals
+            for idx in idxs
+                normals[d1, idx] = coords[d1][idx] - p1
+                normals[d2, idx] = coords[d2][idx] - p2
+                normals[d3, idx] = coords[d3][idx] - p3
+            end
+        end
+    end
+end
```
efaulhaber, line 404:
```suggestion
    for (indices, face_ids) in zip(corner_indices, corner_faces)
        set_intersection_normals!(indices, face_ids)
    end
```

src/setups/rectangular_tank.jl
```diff
@@ -439,24 +674,62 @@ function initialize_boundaries(particle_spacing, tank_size::NTuple{2},
...
-    face_indices_5 = Array{Int, 2}(undef, n_layers, n_particles_x * n_particles_y)
-    face_indices_6 = Array{Int, 2}(undef, n_layers, n_particles_x * n_particles_y)
+    # Faces
+    face_indices_left = Array{Int, 2}(undef, n_layers, n_particles_y * n_particles_z)
+    face_indices_right = Array{Int, 2}(undef, n_layers, n_particles_y * n_particles_z)
+    face_indices_bottom = Array{Int, 2}(undef, n_layers, n_particles_x * n_particles_z)
+    face_indices_top = Array{Int, 2}(undef, n_layers, n_particles_x * n_particles_z)
+    face_indices_front = Array{Int, 2}(undef, n_layers, n_particles_x * n_particles_y)
+    face_indices_back = Array{Int, 2}(undef, n_layers, n_particles_x * n_particles_y)
+
```
efaulhaber, line 707: While left, right, top, bottom is fine in 2D, we found that it was confusing in 3D. Please use the numbers instead to make it more mathematically abstract (because we know that 1,2 is x, 3/4 is y and 5/6 is z, so we don't have to think about what is supposed to be front or back).

For the comments we used "+x", "-x" etc. So please also use "-y/+z edge" instead of "front-bottom edge" to avoid confusion.

src/setups/rectangular_tank.jl
```diff
@@ -193,6 +199,210 @@ struct RectangularTank{NDIMS, NDIMSt2, ELTYPE <: Real, F, B}
...
+    # are any particles associated with it. 
+    function face_has_particles(idxs, i)
+        return faces[i] && !isempty(idxs[i])
+    end
+
+    # Compute normals for the faces
+    # Faces: 1=Left(-x), 2=Right(+x), 3=Bottom(-y), 4=Top(+y), 5=Front(-z), 6=Back(+z)
+    left, right = 1, 2
+    bottom, top = 3, 4
+    front, back = 5, 6
```
efaulhaber, line 318: Same here. Please use the numbers directly and only mention "+y" instead of "top" in the comments.

src/setups/sphere_shape.jl
```diff
@@ -431,3 +434,23 @@ function round_sphere(sphere, particle_spacing, radius, center::SVector{3})
...
+    normals = zeros(T, NDIMS, n_points)
+
+    threshold = eps(radius)
+    for i in axes(coordinates, 2)
+        point = extract_svector(coordinates, Val(NDIMS), i)
+        diff = point - center_position
+        dist = norm(diff)
+
+        if dist > threshold
+            normals[:, i] .= diff / dist * (radius - dist)
```
efaulhaber, line 451:
```suggestion
            # The unit normal is `diff / dist`, and we scale it by the distance to the surface of the sphere.
            normals[:, i] .= diff / dist * (radius - dist)
```

efaulhaber (changes requested):

efaulhaber (dismissed):

src/general/initial_condition.jl
```diff
@@ -201,8 +204,13 @@ function InitialCondition{NDIMS}(coordinates, velocity, mass, density,
         masses = mass_fun.(coordinates_svector)
     end
 
+    if !isnothing(normals) && size(coordinates) != size(normals)
```
efaulhaber, line 211: **High:** When callers pass `normals` with a different number of columns than `coordinates`, this branch constructs an `ArgumentError` but never throws it, allowing invalid `InitialCondition`s to be created. Throw the error so malformed normals are rejected immediately.

src/general/initial_condition.jl
```diff
@@ -283,9 +291,11 @@ function Base.union(initial_condition::InitialCondition, initial_conditions...)
     mass = vcat(initial_condition.mass, ic.mass[valid_particles])
     density = vcat(initial_condition.density, ic.density[valid_particles])
     pressure = vcat(initial_condition.pressure, ic.pressure[valid_particles])
+    normals = isnothing(initial_condition.normals) ? nothing :
```
efaulhaber: **High:** When unioning two initial conditions with normals, the appended normals for `ic` are taken from `ic.coordinates` instead of `ic.normals`. Use `ic.normals` so downstream boundary-normal logic receives normal vectors rather than particle positions.

src/general/initial_condition.jl
```diff
@@ -318,9 +328,10 @@ function Base.setdiff(initial_condition::InitialCondition, initial_conditions...
     mass = initial_condition.mass[valid_particles]
     density = initial_condition.density[valid_particles]
     pressure = initial_condition.pressure[valid_particles]
+    normals = isnothing(initial_condition.normals) ? nothing : initial_condition.normals
```
efaulhaber: **High:** `setdiff` filters particle arrays by `valid_particles` but leaves `normals` at the original size, so normals no longer align with the remaining particles. Slice `initial_condition.normals[:, valid_particles]` when normals are present.

src/preprocessing/geometries/geometries.jl
```diff
@@ -50,10 +53,12 @@ function Base.intersect(initial_condition::InitialCondition,
     mass = initial_condition.mass[keep_indices]
     density = initial_condition.density[keep_indices]
     pressure = initial_condition.pressure[keep_indices]
+    normals = isnothing(initial_condition.normals) ? nothing : initial_condition.normals
```
efaulhaber: **High:** `intersect(initial_condition, geometry)` filters coordinates and particle fields by `keep_indices` but passes through the full original normals array. Slice normals with the same `keep_indices` so removed particles do not leave stale normals attached to the result.

test/setups/normals.jl
```diff
@@ -0,0 +1,70 @@
+@testset verbose=true "Dummy Particles with `MarronePressureExtrapolation`" begin
```
efaulhaber: **Medium:** The new Marrone test file is not included from the unit-test runner, so CI skips all assertions added here. Include this file from the existing schemes test entry point so regressions in normal-vector behavior are actually tested.

src/general/initial_condition.jl
```diff
@@ -95,28 +95,30 @@ initial_condition = InitialCondition(; coordinates, velocity=x -> 2x, mass=1.0,
...
     velocity         :: MATRIX # Array{ELTYPE, 2}
     mass             :: VECTOR # Array{ELTYPE, 1}
     density          :: VECTOR # Array{ELTYPE, 1}
     pressure         :: VECTOR # Array{ELTYPE, 1}
+    normals          :: N
 end
 
 # The default constructor needs to be accessible for Adapt.jl to work with this struct.
 # See the comments in general/gpu.jl for more details.
 function InitialCondition(; coordinates, density, velocity=zeros(size(coordinates, 1)),
```
efaulhaber, line 114: **Low:** The public `InitialCondition` constructor now accepts `normals`, but the docstring signature and keywords omit it. Document the keyword, expected shape, default, and semantics so callers can construct boundary initial conditions correctly.

src/setups/rectangular_tank.jl
```diff
@@ -361,10 +453,10 @@ function initialize_boundaries(particle_spacing, tank_size::NTuple{2},
                                                   (tank_size[1], 0.0),
                                                   loop_order=:x_first)
 
-        # store coordinates of left boundary
+        # Store coordinates of left boundary
```
efaulhaber: **Low:** This comment labels the right-boundary coordinates as left-boundary coordinates, and similar repeated wording appears for other boundary blocks. Update the comments to match the boundary being appended so the indexing and normal-ordering logic remains verifiable.

efaulhaber (commented):
## Codex Review

The patch introduces normals but several construction and set-operation paths either fail to validate them or keep/apply the wrong normal columns. These are correctness issues for new production behavior around boundary normals.

src/general/initial_condition.jl
efaulhaber: Oh, that's true. @RubberLanding you forgot a `throw`.

src/general/initial_condition.jl
efaulhaber: Indeed. @RubberLanding This is a typo and should be `ic.normals`. Please also add a small test for this (or just add normals to the existing union tests).

src/general/initial_condition.jl
efaulhaber: @RubberLanding also true.

src/preprocessing/geometries/geometries.jl
efaulhaber: @RubberLanding same here

test/setups/normals.jl
efaulhaber: @RubberLanding 1. This is true, this test is not included.
2. Please rename the file and test to `test/setups/normals.jl`, since these will not only be used by the Marrone wall BC.

src/general/initial_condition.jl
efaulhaber: @RubberLanding also true, please update the docstring.

src/setups/rectangular_tank.jl
efaulhaber: Yeah idk, can be renamed to `# Append coordinates of left boundary`, but that's very low priority.

src/general/initial_condition.jl
svchb, line 471:
```suggestion
if !isnothing(ic.normals)
    ic.normals .= ic.normals[:, permutation]
end
```

src/general/initial_condition.jl
```diff
@@ -129,7 +131,8 @@ function InitialCondition{NDIMS}(coordinates, velocity, mass, density,
 
     if n_particles == 0
         return InitialCondition(particle_spacing, coordinates, zeros(ELTYPE, NDIMS, 0),
-                                zeros(ELTYPE, 0), zeros(ELTYPE, 0), zeros(ELTYPE, 0))
+                                zeros(ELTYPE, 0), zeros(ELTYPE, 0), zeros(ELTYPE, 0),
+                                normals)
```
svchb:
```suggestion
                                zeros(ELTYPE, NDIMS, 0))
```

src/general/initial_condition.jl
svchb: Another issue is that currently behavior of unioning ics with and without normals depends on the order so there needs to be an error check for this as well.

src/setups/sphere_shape.jl
```diff
@@ -125,8 +125,11 @@ function SphereShape(particle_spacing, radius, center_position, density;
     particles_not_in_cutout = map(!in_cutout, axes(coordinates, 2))
     coordinates = coordinates[:, particles_not_in_cutout]
 
+    normals = calculate_sphere_normals(coordinates, collect(center_position), radius,
+                                       Val(NDIMS))
```
svchb:
```suggestion
normals = calculate_sphere_normals(coordinates, SVector{NDIMS}(center_position),
                                   radius, Val(NDIMS))
```

svchb (changes requested):

src/general/initial_condition.jl
RubberLanding: @efaulhaber If `initial_condition` has normals but `ic` does not, I feel like we should drop the normals altogether for union. For a valid initial condition, the normals should have the same length as the other property arrays, or be `nothing`. I think throwing an error here would be a bad idea, since we calculate the normals automatically for all `RectangularTank`'s and `SphereShape`'s. If we throw an error like
```
if isnothing(initial_condition.normals) != isnothing(ic.normals)
    throw(ArgumentError("Cannot union InitialConditions with mixed normal definitions..."))
end
``` 
we can not create any unions of the above two types and any other shape that doesn't calculate the normals.
Should it throw a warning instead?

efaulhaber: Yes, good point! I agree.

Both have normals -> union should have normals
None have normals -> union should have no normals
only one has normals (no matter if first or second) -> union should not have normals and add an `@warn` to notify the user that the result doesn't have normals.

src/general/initial_condition.jl
```diff
@@ -284,6 +286,10 @@ function Base.union(initial_condition::InitialCondition, initial_conditions...)
         throw(ArgumentError("all passed initial conditions must have the same particle spacing"))
     end
 
+    if isnothing(initial_condition.normals) != isnothing(ic.normals)
+        @warn "Unioning InitialConditions with mixed normal definitions: one geometry has surface normals while the other does not. Surface normals will be discarded (`nothing`) in the combined result."
```
efaulhaber: Please break this long line.

efaulhaber (changes requested):

efaulhaber (dismissed):

src/general/initial_condition.jl
```diff
@@ -274,6 +286,12 @@ function Base.union(initial_condition::InitialCondition, initial_conditions...)
         throw(ArgumentError("all passed initial conditions must have the same particle spacing"))
     end
 
+    if isnothing(initial_condition.normals) != isnothing(ic.normals)
+        @warn "Unioning InitialConditions with mixed normal definitions: one geometry has " *
+              "surface normals while the other does not. Surface normals will be " *
+              "discarded (`nothing`) in the combined result."
+    end
+
```
efaulhaber: Hmmm, maybe we don't want a warning? The tests are currently failing because the open boundary system creates unions in the constructor. We definitely don't want to show warnings there, but we could just disable them in this case.

The real question is: do we even want warnings here in general? In most situations, we don't need normals. And then a user is getting a bunch of irrelevant warnings every time they set up nontrivial initial conditions. Perhaps it's better to throw a descriptive error like "initial condition does not contain normals. Make sure all initial conditions in a `union` have normals" in the future in the places where normals are actually used.

efaulhaber: If @svchb agrees, I suggest just removing this warning here.

svchb: yes its fine

svchb (dismissed):

svchb:
A NEWS.md entry is missing.

efaulhaber (approved):

PR #1037, opened 2026-01-06
Restart from VTU

LasNikas:
based on

- #959


sloede:
Happy to see that restarting is receiving some TLC ♥️

svchb (changes requested):
- Add also an example in which you restart TLSPH
- All systems should be covered by tests

src/general/semidiscretization.jl
```diff
@@ -193,6 +193,10 @@ Create an `ODEProblem` from the semidiscretization with the specified `tspan`.
 - `tspan`: The time span over which the simulation will be run.
 
 # Keywords
+- `restart_with`: Can be used to restart the simulation from VTK solution files (see [`SolutionSavingCallback`](@ref)).
+  This has to be a tuple of filenames, one for each system in the [`Semidiscretization`](@ref).
+  The order of the filenames has to match the order of the systems in the [`Semidiscretization`](@ref).
```
svchb: Unclear meaning of this sentences.

src/general/restart.jl
```diff
@@ -0,0 +1,119 @@
...
+function time_span(tspan, restart_with::Tuple{Vararg{String}})
+    restart_data = vtk2trixi(first(restart_with))
+    t_restart = convert(eltype(tspan), restart_data.time)
+
+    if !isapprox(tspan[1], t_restart)
+        @info "Adjusting initial time from $(tspan[1]) to restart time $t_restart"
+    end
+
+    return (t_restart, tspan[2])
+end
```
svchb, line 53: This should check if all the files start from the same time.

svchb (changes requested):
why don't you read in the index for the buffered particles by modifying vtk2trixi?
As far as I understand it the particles in the buffer don't have to be continuous?

LasNikas:
> why don't you read in the index for the buffered particles by modifying vtk2trixi?
As far as I understand it the particles in the buffer don't have to be continuous?

I don't quite understand the question. Could you elaborate @svchb ? The vtk files only contain active particles. So only active particles are written and loaded anyway.

NEWS.md
```diff
@@ -4,6 +4,17 @@ TrixiParticles.jl follows the interpretation of
 [semantic versioning (semver)](https://julialang.github.io/Pkg.jl/dev/compatibility/#Version-specifier-format-1)
 used in the Julia ecosystem. Notable changes will be documented in this file for human readability.
 
+## Version 0.5.1
+
+### Features
+- Added the ability to restart simulations from VTK solution files generated by `SolutionSavingCallback`.
+  Users can now pass the `restart_with` keyword argument to `semidiscretize` (#1190).
```
svchb: not correct anymore

svchb (changes requested):

src/schemes/structure/total_lagrangian_sph/system.jl
```diff
@@ -602,6 +602,16 @@ function restart_with!(system::TotalLagrangianSPHSystem, v, u)
...
 
+function restart_u(system::AbstractStructureSystem, data)
+    # The integration array requires only the particles that need to be integrated
+    return data.coordinates[:, each_integrated_particle(system)]
+end
+
+function restart_v(system::AbstractStructureSystem, data)
+    # The integration array requires only the particles that need to be integrated
+    return data.velocity[:, each_integrated_particle(system)]
+end
```
svchb: I don't think this is correct since this doesn't support RigidBodySystem which is also a AbstractStructureSystem

src/general/restart.jl
```diff
@@ -0,0 +1,125 @@
...
+
+function write_density_and_pressure!(v_restart, system::EntropicallyDampedSPHSystem,
+                                     density_calculator::ContinuityDensity,
+                                     pressure, density)
+    v_restart[size(v_restart, 1), :] = density
+    v_restart[size(v_restart, 1) - 1, :] = pressure
+
+    return v_restart
+end
```
svchb, line 83: Version for summationDensity is missing

test/examples/gpu.jl
```diff
@@ -896,4 +896,58 @@ end
...
+
+        # Since this is an open boundary simulation, the number of active particles may
+        # differ. The results must be interpolated to enable comparison with the restart
+        # simulation. The fluid domain starts at `x = 10 * particle_spacing`.
+        n_interpolation_points = 10
+        start_point = [0.0f0 + 10 * particle_spacing, wall_distance / 2]
+        end_point = [flow_length - 10 * particle_spacing, wall_distance / 2]
+        result_full = interpolate_line(start_point, end_point, n_interpolation_points,
+                                       sol.prob.p, sol.prob.p.systems[1], sol,
+                                       cut_off_bnd=false)
```
svchb, line 921:
```suggestion
result_full = interpolate_line(start_point, end_point, n_interpolation_points,
                               sol.prob.p.semi, sol.prob.p.semi.systems[1], sol;
                               cut_off_bnd=false)
```

test/examples/gpu.jl
```diff
@@ -896,4 +896,58 @@ end
...
+        trixi_include_changeprecision(Float32, @__MODULE__,
+                                      joinpath(examples_dir(), "fluid",
+                                               "poiseuille_flow_2d.jl"),
+                                      tspan=(0.0f0, 0.3f0), sound_speed_factor=10,
+                                      particle_spacing=4.0f-5,
+                                      coordinates_eltype=Float32,
+                                      parallelization_backend=Main.parallelization_backend)
+
+        # Transfer `semi` back to CPU
+        semi_new = TrixiParticles.Adapt.adapt(Array, sol.prob.p)
```
svchb:
```suggestion
     semi_cpu = TrixiParticles.Adapt.adapt(Array, sol.prob.p.semi)
```

svchb (changes requested):
Also doesn't work with BoundaryModelMonaghanKajtar.

svchb (dismissed):

src/schemes/fluid/fluid.jl
```diff
@@ -253,6 +253,47 @@ end
...
+    return coords_total
+end
+
+function restart_v(system::AbstractFluidSystem, data)
+    velocity_total = zeros(eltype(system), v_nvariables(system),
+                           n_integrated_particles(system))
+    velocity_active = zeros(eltype(system), v_nvariables(system), size(data.velocity, 2))
+
+    velocity_active[1:ndims(system), :] = data.velocity
+    write_density_and_pressure!(velocity_active, system, density_calculator(system),
```
efaulhaber, line 286: ERROR: Restarting a fluid system calls `density_calculator(system)`, but fluid systems store this as `system.density_calculator` and do not define that accessor. Use the field or add the missing accessor so `semidiscretize(...; restart_with=...)` does not throw a `MethodError`.

test/examples/gpu.jl
```diff
@@ -896,4 +896,58 @@ end
...
+        ode_restart = semidiscretize(semi_cpu, (0.3f0, 0.6f0);
+                                     restart_with=(fluid_restart, open_boundary_restart,
+                                                   boundary_restart))
+
+        sol_restart = solve(ode_restart, RDPK3SpFSAL35(), abstol=1.0f-5,
+                            reltol=1.0f-3, dtmax=1.0f-2, save_everystep=false,
+                            callback=UpdateCallback())
+
+        result_restart = interpolate_line(start_point, end_point,
+                                          n_interpolation_points, sol_restart.prob.p,
```
efaulhaber: ERROR: The GPU restart test passes `sol_restart.prob.p` where the semidiscretization is expected, but the semidiscretization is stored in `sol_restart.prob.p.semi`. Pass `sol_restart.prob.p.semi` and its first system to avoid a runtime failure before the restart comparisons run.

src/schemes/fluid/fluid.jl
```diff
@@ -253,6 +253,47 @@ end
...
+
+    coords_active = data.coordinates
+
+    for particle in axes(coords_active, 2)
+        for dim in 1:ndims(system)
+            coords_total[dim, particle] = coords_active[dim, particle]
+        end
+    end
+
+    if !isnothing(buffer(system))
```
efaulhaber: ERROR: For buffered fluid systems with inactive gaps, restart ignores the saved VTK particle `index` and marks slots `1:N` active, which can attach restored coordinates and velocities to the wrong per-particle state. Restore particles into their saved indices instead of compacting them.

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -705,6 +710,66 @@ function interpolate_velocity!(system::OpenBoundarySystem, boundary_zone,
...
+    coords_total .= coordinates_eltype(system)(1e16)
+
+    coords_active = data.coordinates
+    for particle in axes(coords_active, 2)
+        for dim in 1:ndims(system)
+            coords_total[dim, particle] = coords_active[dim, particle]
+        end
+    end
+
+    system.buffer.active_particle .= false
```
efaulhaber, line 728: ERROR: For buffered open boundaries, restart compacts active coordinates into slots `1:N` instead of using the saved VTK particle `index`, so particles can be associated with the wrong zone/cache data after holes existed in the active set. Restore open-boundary particles to their saved indices.

test/general/restart.jl
```diff
@@ -0,0 +1,109 @@
...
+        open_boundary_restart = joinpath("out", "open_boundary_1_$iter.vtu")
+        boundary_restart = joinpath("out", "boundary_1_$iter.vtu")
+
+        ode_restart = semidiscretize(semi, (0.3, 0.6);
+                                     restart_with=(fluid_restart, open_boundary_restart,
+                                                   boundary_restart))
+        restart_pressure = ode_restart.p.semi.systems[2].cache.pressure_reference_values[2].pressure[]
+        restart_flow_rate = ode_restart.p.semi.systems[2].cache.pressure_reference_values[2].flow_rate[]
+
+        @test isapprox(restart_pressure,
```
efaulhaber, line 80: WARNING: These assertions compare against `open_boundary`, which aliases the restarted semidiscretization on the CPU path, so they can pass even if VTK pressure or flow-rate fields are not restored. Capture independent expected values before restart or restart from a fresh semidiscretization with different cache values.

src/io/read_vtk.jl
```diff
@@ -51,8 +51,19 @@ function vtk2trixi(file; element_type=nothing, coordinates_eltype=nothing,
...
+    if !isnothing(element_type)
+        ELTYPE = element_type
+    else
+        # Try to get element type from pressure ore density (whichever exists)
+        ELTYPE = if "pressure" in keys(point_data)
+            eltype(first(ReadVTK.get_data(point_data["pressure"])))
+        elseif "material_density" in keys(point_data)
+            eltype(first(ReadVTK.get_data(point_data["material_density"])))
+        else
+            error("Neither 'pressure' nor 'material_density' field found in VTK file")
```
efaulhaber: INFO: This VTK-content validation uses `error(...)`, but project conventions require explicit `throw(ArgumentError(...))` for user input validation; also format the message lowercase without a trailing period.

src/general/restart.jl
```diff
@@ -0,0 +1,133 @@
+function set_initial_conditions!(v0_ode, u0_ode, semi, restart_with::Tuple{Vararg{String}})
+    # Check number of systems
+    if length(semi.systems) != length(restart_with)
+        throw(ArgumentError("Number of systems in `semi` does not match number of restart files provided " *
```
efaulhaber: INFO: The new restart validation messages start uppercase and some end with periods, which conflicts with the repository error-message style. Normalize these new messages to start lowercase and omit terminal periods for single-sentence errors.

src/general/semidiscretization.jl
```diff
@@ -193,6 +193,10 @@ Create an `ODEProblem` from the semidiscretization with the specified `tspan`.
 - `tspan`: The time span over which the simulation will be run.
 
 # Keywords
+- `restart_with`: Can be used to restart the simulation from VTK solution files (see [`SolutionSavingCallback`](@ref)).
```
efaulhaber, line 196: INFO: The `restart_with` keyword documentation describes the file arguments but not that `semidiscretize` replaces the initial time with the timestamp read from the VTK files. Document that behavior, including mismatch adjustment/logging, so users know `tspan[1]` may not be honored.

examples/postprocessing/restart_oscillating_beam_2d.jl
```diff
@@ -0,0 +1,27 @@
+# ==========================================================================================
+# Restart Example: Oscillting Beam 2D
```
efaulhaber: INFO: The example title has a user-facing typo: `Oscillting` should be `Oscillating`. Correct the title so the restart example header reads cleanly.

src/io/read_vtk.jl
```diff
@@ -51,8 +51,19 @@ function vtk2trixi(file; element_type=nothing, coordinates_eltype=nothing,
     point_coords = ReadVTK.get_points(vtk_file)
 
     cELTYPE = isnothing(coordinates_eltype) ? eltype(point_coords) : coordinates_eltype
-    ELTYPE = isnothing(element_type) ?
-             eltype(first(ReadVTK.get_data(point_data["pressure"]))) : element_type
+
+    if !isnothing(element_type)
+        ELTYPE = element_type
+    else
+        # Try to get element type from pressure ore density (whichever exists)
```
efaulhaber: INFO: The added comment says `pressure ore density`; change it to `pressure or density` so the implementation note is clear.

efaulhaber (commented):
## Codex Review

The PR intends to add restarting simulations from VTK solution files via `semidiscretize(...; restart_with=...)`, including restart helpers for fluid, boundary, and structure systems.

src/schemes/fluid/fluid.jl
efaulhaber: Correct. But not entirely. `density_calculator(system)` is defined for all systems, but will error for some with a `FieldError` (not a `MethodError`).

Found by both bots.

test/examples/gpu.jl
efaulhaber: Sounds correct. This was changed in #1049. It looks like GPU tests would fail. @LasNikas can you please check if this is correct?

Missed by Copilot.

src/schemes/fluid/fluid.jl
efaulhaber: I don't know what that is supposed to mean. @LasNikas `data` only contains active data, right?

src/schemes/boundary/open_boundary/system.jl
efaulhaber: I think this is the same as the previous comment. @LasNikas can this be a problem or will inactive particles be handled correctly?

test/general/restart.jl
efaulhaber: This is confusing and sounds incorrect.

src/io/read_vtk.jl
efaulhaber: Okay that's on me for using a misleading custom instruction for the review bot.

src/general/restart.jl
efaulhaber: Correct.
```suggestion
        throw(ArgumentError("number of systems in `semi` does not match number of restart files provided " *
```
Not caught by Copilot because that's a custom instruction.

src/general/semidiscretization.jl
efaulhaber: Correct. This should be mentioned in the docs.

Missed by Copilot.

examples/postprocessing/restart_oscillating_beam_2d.jl
efaulhaber: Correct and found by both bots.

src/io/read_vtk.jl
efaulhaber: Correct and found by both bots.

src/schemes/fluid/fluid.jl
```diff
@@ -253,6 +253,47 @@ end
...
+        system.buffer.active_particle .= false
+        system.buffer.active_particle[1:size(coords_active, 2)] .= true
+    end
+
+    update_system_buffer!(system.buffer)
+
```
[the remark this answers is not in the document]
efaulhaber: Correct and missed by my review bot.

src/schemes/fluid/fluid.jl
```diff
@@ -253,6 +253,47 @@ end
...
+end
+
+function restart_v(system::AbstractFluidSystem, data)
+    velocity_total = zeros(eltype(system), v_nvariables(system),
+                           n_integrated_particles(system))
+    velocity_active = zeros(eltype(system), v_nvariables(system), size(data.velocity, 2))
+
+    velocity_active[1:ndims(system), :] = data.velocity
+    write_density_and_pressure!(velocity_active, system, density_calculator(system),
+                                data.pressure, data.density)
```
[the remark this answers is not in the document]
efaulhaber: Correct and found by both bots.

src/general/semidiscretization.jl
```diff
@@ -219,9 +223,16 @@ timespan: (0.0, 1.0)
...
-function semidiscretize(semi, tspan; reset_threads=true)
+function semidiscretize(semi, tspan; reset_threads=true, restart_with=nothing)
     (; systems) = semi
 
+    if restart_with isa String
+        restart_with = (restart_with,)
+    elseif !isnothing(restart_with) && !(restart_with isa Tuple)
+        throw(ArgumentError("`restart_with` must be `nothing`, a string, or a tuple of strings, " *
+                            "got $(typeof(restart_with))"))
+    end
```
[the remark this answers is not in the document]
efaulhaber: Correct, and missed by my bot. But the suggestion is weird. Should be
```suggestion
    if restart_with isa String
        restart_with = (restart_with,)
    elseif !isnothing(restart_with) && !(restart_with isa NTuple{<:Any, String})
        throw(ArgumentError("`restart_with` must be `nothing`, a string, or a tuple of strings, " *
                            "got $(typeof(restart_with))"))
    end
```

test/general/restart.jl
```diff
@@ -0,0 +1,109 @@
...
+        result_full = interpolate_line(start_point, end_point, n_interpolation_points,
+                                       semi, fluid_system, sol, cut_off_bnd=false)
+
+        # Run half simulation
+        trixi_include(@__MODULE__,
+                      joinpath(examples_dir(), "fluid", "poiseuille_flow_2d.jl"),
+                      tspan=(0.0, 0.3), sound_speed_factor=10, particle_spacing=4e-5,
+                      info_callback=nothing)
+
+        iter = round(Int, 0.3 / 0.02)
```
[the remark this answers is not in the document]
efaulhaber: Correct. Not important, but okay.

test/general/restart.jl
```diff
@@ -0,0 +1,109 @@
...
+        pressure_model = RCRWindkesselModel(; peripheral_resistance=R2,
+                                            compliance=C,
+                                            characteristic_resistance=R1)
+        # Run half simulation
+        trixi_include(@__MODULE__,
+                      joinpath(examples_dir(), "fluid", "poiseuille_flow_2d.jl"),
+                      outlet_reference_pressure=pressure_model, info_callback=nothing,
+                      tspan=(0.0, 0.3), sound_speed_factor=10, particle_spacing=4e-5)
+
+        iter = round(Int, 0.3 / 0.02)
```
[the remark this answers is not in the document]
efaulhaber: Same as above.

src/io/read_vtk.jl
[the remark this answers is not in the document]
efaulhaber: Correct and found by both bots.

examples/postprocessing/restart_oscillating_beam_2d.jl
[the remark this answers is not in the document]
efaulhaber: Correct and found by both bots.

test/examples/gpu.jl
```diff
@@ -896,4 +896,58 @@ end
...
+                                               "poiseuille_flow_2d.jl"),
+                                      tspan=(0.0f0, 0.3f0), sound_speed_factor=10,
+                                      particle_spacing=4.0f-5,
+                                      coordinates_eltype=Float32,
+                                      parallelization_backend=Main.parallelization_backend)
+
+        # Transfer `semi` back to CPU
+        semi_cpu = TrixiParticles.Adapt.adapt(Array, sol.prob.p.semi)
+
+        iter = round(Int, 0.3 / 0.02)
```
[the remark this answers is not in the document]
efaulhaber: Correct and missed by my bot.

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -705,6 +710,66 @@ function interpolate_velocity!(system::OpenBoundarySystem, boundary_zone,
...
+
+function restart_v(system::OpenBoundarySystem, data)
+    v_total = zeros(eltype(system), v_nvariables(system),
+                    n_integrated_particles(system))
+
+    v_active = zeros(eltype(system), v_nvariables(system), size(data.velocity, 2))
+
+    v_active[1:ndims(system), :] = data.velocity
+    write_density_and_pressure!(v_active, system.fluid_system,
+                                density_calculator(system), data.pressure, data.density)
```
efaulhaber, line 747: `density_calculator(system)` is always `nothing` for open boundaries, so the density with the dynamical pressure model is not correctly applied, no?

test/examples/gpu.jl
```diff
@@ -896,4 +896,58 @@ end
...
+                                      particle_spacing=4.0f-5,
+                                      coordinates_eltype=Float32,
+                                      parallelization_backend=Main.parallelization_backend)
+
+        # Since this is an open boundary simulation, the number of active particles may
+        # differ. The results must be interpolated to enable comparison with the restart
+        # simulation. The fluid domain starts at `x = 10 * particle_spacing`.
+        n_interpolation_points = 10
+        start_point = [0.0f0 + 10 * particle_spacing, wall_distance / 2]
+        end_point = [flow_length - 10 * particle_spacing, wall_distance / 2]
```
efaulhaber: `flow_length` and `wall_distance` are not defined in the example file.

svchb:
@codex review

src/schemes/fluid/fluid.jl
```diff
@@ -282,8 +282,11 @@ function restart_v(system::AbstractFluidSystem, data)
     velocity_active = zeros(eltype(system), v_nvariables(system), size(data.velocity, 2))
 
     velocity_active[1:ndims(system), :] = data.velocity
-    write_density_and_pressure!(velocity_active, system, density_calculator(system),
-                                data.pressure, data.density)
+
+    density_calc = hasproperty(system, :density_calculator) ? density_calculator(system) :
+                   nothing
```
efaulhaber: This should be handled by dispatch by defining `density_calculator(system)` for all systems.

src/schemes/boundary/open_boundary/system.jl
efaulhaber: @LasNikas what happens here?

examples/postprocessing/restart_oscillating_beam_2d.jl
```diff
@@ -0,0 +1,27 @@
+# ==========================================================================================
```
efaulhaber: Do we need a second restart example? This is trivial if one looked at the other file before, no?

examples/postprocessing/restart_poiseuille_flow_2d.jl
```diff
@@ -0,0 +1,31 @@
+# ==========================================================================================
```
efaulhaber, line 1: Does it make sense to have this as an example file? I think a Literate tutorial with a handful of extra explanations (like that the order of the files has to match the order of systems in the semidiscretization) might be better. Can be a new PR though to accelerate review on this one.

src/general/restart.jl
```diff
@@ -0,0 +1,133 @@
+function set_initial_conditions!(v0_ode, u0_ode, semi, restart_with::Tuple{Vararg{String}})
```
efaulhaber, line 2:
```suggestion
# Update `v0_ode` and `u0_ode` with the data from the restart files.
function set_initial_conditions!(v0_ode, u0_ode, semi, restart_with::Tuple{Vararg{String}})
```

src/general/restart.jl
```diff
@@ -0,0 +1,133 @@
...
+        u_restart = restart_u(system, restart_data)
+
+        v0_system .= Adapt.adapt(semi.parallelization_backend, v_restart)
+        u0_system .= Adapt.adapt(semi.parallelization_backend, u_restart)
+
+        restore_previous_state!(system, restart_file)
+    end
+end
+
+function time_span(tspan, restart_with::Tuple{Vararg{String}})
```
efaulhaber, line 38:
```suggestion
# Compute a new tspan based on the restart files.
function time_span(tspan, restart_with::Tuple{Vararg{String}})
```

src/general/restart.jl
```diff
@@ -0,0 +1,133 @@
...
+    # Read restart times from all files
+    restart_times = [vtk2trixi(file).time for file in restart_with]
+    t_restart = convert(eltype(tspan), first(restart_times))
+
+    # Check if all restart files have the same time
+    if !all(isapprox(t, t_restart) for t in restart_times)
+        throw(ArgumentError("All restart files must start from the same time."))
```
efaulhaber:
```suggestion
        throw(ArgumentError("all restart files must start from the same time"))
```

src/general/semidiscretization.jl
```diff
@@ -193,6 +193,14 @@ Create an `ODEProblem` from the semidiscretization with the specified `tspan`.
 - `tspan`: The time span over which the simulation will be run.
 
 # Keywords
+- `restart_with`: Can be used to restart the simulation from VTK solution files (see [`SolutionSavingCallback`](@ref)).
+  This can be either `nothing` (default, no restart), a single filename as a `String`,
```
efaulhaber: Mention that it can only be a single filename when the simulation contains only a single system. Or maybe don't mention this at all and just say "tuple of file names, one for each system". I think this is clear enough for one system as well, and if it also works when forgetting the comma and therefore passing the file without a tuple, that's good.

src/io/read_vtk.jl
```diff
@@ -51,8 +51,19 @@ function vtk2trixi(file; element_type=nothing, coordinates_eltype=nothing,
...
+    if !isnothing(element_type)
+        ELTYPE = element_type
+    else
+        # Try to get element type from pressure or density (whichever exists)
+        ELTYPE = if "pressure" in keys(point_data)
+            eltype(first(ReadVTK.get_data(point_data["pressure"])))
+        elseif "material_density" in keys(point_data)
+            eltype(first(ReadVTK.get_data(point_data["material_density"])))
+        else
+            error("Neither 'pressure' nor 'material_density' field found in VTK file")
```
```suggestion
            error("neither 'pressure' nor 'material_density' field found in VTK file")
```

src/io/read_vtk.jl
```diff
@@ -87,7 +98,8 @@ function vtk2trixi(file; element_type=nothing, coordinates_eltype=nothing,
                                  results[:particle_spacing]
     results[:coordinates] = coordinates
     results[:time] = "time" in keys(field_data) ?
-                     first(ReadVTK.get_data(field_data["time"])) : zero(ELTYPE)
+                     convert.(ELTYPE, first(ReadVTK.get_data(field_data["time"]))) :
```
```suggestion
                     convert(ELTYPE, first(ReadVTK.get_data(field_data["time"]))) :
```
Not an array?

src/schemes/fluid/fluid.jl
```diff
@@ -253,6 +253,50 @@ end
     return nothing
 end
 
+function restart_u(system::AbstractFluidSystem, data)
+    coords_total = zeros(coordinates_eltype(system), u_nvariables(system),
+                         n_integrated_particles(system))
+    coords_total .= coordinates_eltype(system)(1e16)
```
efaulhaber:
```suggestion
    inactive_coords = convert(coordinates_eltype(system), 1e16)
    coords_total = fill(inactive_coords, u_nvariables(system),
                         n_integrated_particles(system))
```

test/examples/gpu.jl
```diff
@@ -896,4 +896,55 @@ end
...
+                            callback=UpdateCallback())
+
+        result_restart = interpolate_line(start_point, end_point,
+                                          n_interpolation_points, sol_restart.prob.p.semi,
+                                          sol_restart.prob.p.semi.systems[1],
+                                          sol_restart, cut_off_bnd=false)
+
+        @test isapprox(result_full.velocity, result_restart.velocity, rtol=8.0f-3)
+        @test isapprox(result_full.density, result_restart.density, rtol=8.0f-4)
+        @test isapprox(result_full.pressure, result_restart.pressure, rtol=8.0f-2)
```
efaulhaber: Why such a large tolerance? Shouldn't this be exact to machine precision?

test/general/restart.jl
```diff
@@ -0,0 +1,109 @@
...
+
+        result_restart = interpolate_line(start_point, end_point,
+                                          n_interpolation_points, sol_restart.prob.p.semi,
+                                          sol_restart.prob.p.semi.systems[1],
+                                          sol_restart, cut_off_bnd=false)
+
+        @test isapprox(result_full.velocity, result_restart.velocity, rtol=1e-2)
+        @test isapprox(result_full.density, result_restart.density, rtol=8e-4)
+        @test isapprox(result_full.pressure, result_restart.pressure, rtol=8e-2)
+    end
```
efaulhaber: Same question here.

test/general/restart.jl
```diff
@@ -0,0 +1,109 @@
...
+                      tspan=(0.0, 0.5), n_particles_y=5)
+
+        # Get latest iteration and create restart
+        iter = saving_callback.affect!.affect!.latest_saved_iter
+        restart_file = joinpath("out", "structure_1_$iter.vtu")
+
+        ode_restart = semidiscretize(semi, (0.5, 1.0);
+                                     restart_with=restart_file)
+
+        # Create a new saving callback for the restart
```
efaulhaber: Why? This is not used anyway in this test.

efaulhaber (changes requested):
Now a human review. Only small things because AI found all other issues.

examples/postprocessing/restart_oscillating_beam_2d.jl
LasNikas: This was @svchb idea

examples/postprocessing/restart_poiseuille_flow_2d.jl
LasNikas: In that case, it would be better to discuss this in another PR or in an issue.

examples/postprocessing/restart_oscillating_beam_2d.jl
efaulhaber: I think the other example file is simple enough that we don't need a simpler one as well.

examples/postprocessing/restart_poiseuille_flow_2d.jl
LasNikas: see #1236

src/schemes/boundary/open_boundary/system.jl
LasNikas: Here is the dispatch for the DPBC:

https://github.com/trixi-framework/TrixiParticles.jl/blob/2b76eafbc0967f247eb92f71b9799bbd7e53faf1/src/schemes/boundary/open_boundary/dynamical_pressure.jl#L56-L58

src/schemes/fluid/fluid.jl
LasNikas: That's how I had implemented it. But your bot suggested something else, and you confirmed it.

https://github.com/trixi-framework/TrixiParticles.jl/pull/1037#discussion_r3347583629

test/examples/gpu.jl
LasNikas: Hmm idk. Maybe restarting the simulation splits error accumulation across two integration phases. The errors from the first half propagate into the second, resulting in larger total deviations than a continuous run would produce.

src/schemes/fluid/fluid.jl
efaulhaber: I see. That's a misunderstanding. I agreed to
> restart_v assumes every AbstractFluidSystem supports density_calculator(system), but some systems (e.g. ImplicitIncompressibleSPHSystem) do not define a density_calculator field/method and will error during restart. 

The suggestion is nonsense. The clean solution would be to just define `density_calculator` for all systems.

test/examples/gpu.jl
efaulhaber: It just continues the integration. If the full simulation also has a tstop at 0.3 (which it should have from the saving callback), they should be identical.

But I just noticed the example is using adaptive time integration. Of course, the scheme doesn't know what the last stable time step was, so it starts with a different time step. Can you please use a non-adaptive scheme in this test?

LasNikas: Okay best I can do is 
```
        @test isapprox(result_full.velocity, result_restart.velocity, rtol=7e-5)
        @test isapprox(result_full.density, result_restart.density, rtol=8e-5)
        @test isapprox(result_full.pressure, result_restart.pressure, rtol=9e-4)
```
however,  it requires quite a few changes to make it compatible with the non-adaptive scheme

test/examples/gpu.jl
```diff
@@ -934,17 +942,18 @@ end
                                      restart_with=(fluid_restart, open_boundary_restart,
                                                    boundary_restart))
 
-        sol_restart = solve(ode_restart, RDPK3SpFSAL35(), abstol=1.0f-5,
-                            reltol=1.0f-3, dtmax=1.0f-2, save_everystep=false,
-                            callback=UpdateCallback())
+        sol_restart = solve(ode_restart, CarpenterKennedy2N54(williamson_condition=false),
+                            dt=1.0f0, save_everystep=false,
+                            callback=CallbackSet(UpdateCallback(),
+                                                 StepsizeCallback(cfl=0.9)))
```
efaulhaber: This is not the same CFL as above. Can you get even lower tolerances with cfl=1.5 here as well?

LasNikas: No, that was just a copying error. I tested everything with 1.5.

efaulhaber (dismissed):

test/examples/gpu.jl
sloede: > But I just noticed the example is using adaptive time integration. Of course, the scheme doesn't know what the last stable time step was, so it starts with a different time step.

That's a good guess! Note that for that reason we save additional info for adaptive time stepping schemes in Trixi.jl:
https://github.com/trixi-framework/Trixi.jl/blob/4f183aab8c1431fdf5e179744df4af353a2ca09c/src/callbacks_step/save_restart_dg.jl#L338-L354

efaulhaber (approved):

test/examples/gpu.jl
efaulhaber: Thanks, that's cool! @LasNikas do you mind adding this and a small test in a small follow-up PR?

LasNikas: #1239

PR #1041, opened 2026-01-11
Allow overwriting solutions

LasNikas:
At first glance, this keyword does not really make sense. 
Since one would achieve the same result with `save_final_solution=true` and `save_times=[tspan[end]]`, right? 
However, one would then not have regular backups of the results. 
As I mention in the docs:

>  Useful for memory efficiency in large simulations where only
> the final result matters. The difference to simply setting
> `save_final_solution=true` is that this provides regular checkpoint
> backups at each save interval.


src/callbacks/solution_saving.jl
```diff
@@ -23,6 +23,12 @@ To ignore a custom quantity for a specific system, return `nothing`.
 - `save_times=[]`               List of times at which to save a solution.
 - `save_initial_solution=true`: Save the initial solution.
 - `save_final_solution=true`:   Save the final solution.
+- `overwrite=false`:            If `true`, overwrite the previous VTK file with the latest solution.
```
svchb:
```suggestion
- `overwrite=false`:            If `true`, overwrite the previous set of VTK files instead of creating a new set of files each interval.
```

src/io/write_vtk.jl
```diff
@@ -103,9 +104,9 @@ function trixi2vtk(system_, dvdu_ode_, vu_ode_, semi_, t, periodic_box;
     v = wrap_v(v_ode, system, semi)
     u = wrap_u(u_ode, system, semi)
 
-    file = joinpath(output_directory,
-                    add_underscore_to_optional_prefix(prefix) * "$system_name"
-                    * add_underscore_to_optional_postfix(iter))
+    file_ = joinpath(output_directory,
+                     add_underscore_to_optional_prefix(prefix) * "$system_name")
+    file = iter < 0 ? file_ : file_ * add_underscore_to_optional_postfix(iter)
```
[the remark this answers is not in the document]
LasNikas: Hmm but then it is not consistent with `pvd = paraview_collection(collection_file; append=iter > 0)`.
What's wrong with `-1`?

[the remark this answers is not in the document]
svchb: No its not wrong but you could skip writing the collection file since with one timestep in it it doesn't really make sense.

src/callbacks/solution_saving.jl
```diff
@@ -23,6 +23,14 @@ To ignore a custom quantity for a specific system, return `nothing`.
...
 - `save_initial_solution=true`: Save the initial solution.
 - `save_final_solution=true`:   Save the final solution.
+- `overwrite=false`:            If `true`, overwrite the previous set of VTK files instead of
+                                creating a new set of files each interval.
+                                Useful for memory efficiency in large simulations where only
+                                the final result matters. The difference to simply setting
+                                `save_final_solution=true` is that this provides regular checkpoint
+                                backups at each save interval.
+                                If `false` (default), files are not overwritten and an iteration
+                                postfix is appended for each interval.
```
svchb:
```suggestion
- `overwrite=false`:            If `true`, overwrite the previous set of VTK files instead of
                                creating a new set of files each interval.
                                Useful for memory efficiency in large simulations where only
                                the final results matter. The difference to simply setting
                                `save_final_solution=true` is that this provides regular checkpoint
                                backups at each save interval.
                                If `false` (default), files are not overwritten and an iteration
                                postfix is appended for each interval.
```

src/io/write_vtk.jl
```diff
@@ -351,17 +361,17 @@ function write2vtk!(vtk, v, u, t, system::AbstractFluidSystem)
...
-                                                                system,
-                                                                system,
-                                                                particle,
-                                                                neighbor,
-                                                                pos_diff,
-                                                                distance,
-                                                                rho_a,
-                                                                rho_b,
-                                                                grad_kernel)
+            particle] .+= surface_tension_force(surface_tension_a,
```
svchb: the formatting is not correct here

src/callbacks/solution_saving.jl
```diff
@@ -183,8 +193,8 @@ function (solution_callback::SolutionSavingCallback)(integrator)
         end
 
         trixi2vtk(dvdu_ode, vu_ode, semi, integrator.t;
-                  iter, output_directory, prefix, git_hash=git_hash[],
-                  max_coordinates, custom_quantities...)
+                  iter=(overwrite ? -1 : iter), output_directory, prefix,
```
svchb:
```suggestion
                  iter=(overwrite ? nothing : iter), output_directory, prefix,
```

src/io/write_vtk.jl
```diff
@@ -9,7 +9,7 @@ function system_names(systems)
 end
 
 """
-    trixi2vtk(vu_ode, semi, t; iter=nothing, output_directory="out", prefix="",
+    trixi2vtk(vu_ode, semi, t; iter=-1, output_directory="out", prefix="",
```
svchb:
```suggestion
    trixi2vtk(vu_ode, semi, t; iter=nothing, output_directory="out", prefix="",
```

src/io/write_vtk.jl
```diff
@@ -21,8 +21,9 @@ Convert Trixi simulation data to VTK format.
 - `t`:      Current time of the simulation.
 
 # Keywords
-- `iter=nothing`:           Iteration number when multiple iterations are to be stored in
+- `iter=-1`:                Iteration number when multiple iterations are to be stored in
```
svchb:
```suggestion
- `iter=nothing`:                Iteration number when multiple iterations are to be stored in
```

src/io/write_vtk.jl
```diff
@@ -21,8 +21,9 @@ Convert Trixi simulation data to VTK format.
...
                             separate files. This number is just appended to the filename.
+                            If `-1`, no iteration number is appended (default).
```
svchb:
```suggestion
                            If `nothing`, no iteration number is appended (default).
```

src/io/write_vtk.jl
```diff
@@ -47,7 +48,7 @@ trixi2vtk(sol.u[end], semi, 0.0, iter=1, my_custom_quantity=kinetic_energy)
 
 ```
 """
-function trixi2vtk(vu_ode, semi, t; iter=nothing, output_directory="out",
+function trixi2vtk(vu_ode, semi, t; iter=-1, output_directory="out",
```
svchb:
```suggestion
function trixi2vtk(vu_ode, semi, t; iter=nothing, output_directory="out",
```

src/io/write_vtk.jl
```diff
@@ -58,7 +59,7 @@ function trixi2vtk(vu_ode, semi, t; iter=nothing, output_directory="out",
                      prefix, git_hash, max_coordinates, custom_quantities...)
 end
 
-function trixi2vtk(dvdu_ode, vu_ode, semi, t; iter=nothing, output_directory="out",
+function trixi2vtk(dvdu_ode, vu_ode, semi, t; iter=-1, output_directory="out",
```
svchb:
```suggestion
function trixi2vtk(dvdu_ode, vu_ode, semi, t; iter=nothing, output_directory="out",
```

src/io/write_vtk.jl
```diff
@@ -85,7 +86,7 @@ end
 
 # Convert data for a single TrixiParticle system to VTK format
 function trixi2vtk(system_, dvdu_ode_, vu_ode_, semi_, t, periodic_box;
-                   output_directory="out", prefix="", iter=nothing,
+                   output_directory="out", prefix="", iter=-1,
```
svchb:
```suggestion
                   output_directory="out", prefix="", iter=nothing,
```

src/io/write_vtk.jl
```diff
@@ -103,15 +104,21 @@ function trixi2vtk(system_, dvdu_ode_, vu_ode_, semi_, t, periodic_box;
...
-                    * add_underscore_to_optional_postfix(iter))
+    overwrite = iter == -1 ? true : false
+
+    file_ = joinpath(output_directory,
+                     add_underscore_to_optional_prefix(prefix) * "$system_name")
+    if overwrite
+        file = file_ * "_current"
+    else
+        file = file_ * add_underscore_to_optional_postfix(iter)
+    end
```
svchb:
```suggestion

```

src/io/write_vtk.jl
```diff
@@ -103,15 +104,21 @@ function trixi2vtk(system_, dvdu_ode_, vu_ode_, semi_, t, periodic_box;
     v = wrap_v(v_ode, system, semi)
     u = wrap_u(u_ode, system, semi)
 
-    file = joinpath(output_directory,
-                    add_underscore_to_optional_prefix(prefix) * "$system_name"
-                    * add_underscore_to_optional_postfix(iter))
+    overwrite = iter == -1 ? true : false
```
svchb:
```suggestion
```

src/io/write_vtk.jl
```diff
@@ -153,9 +160,12 @@ function trixi2vtk(system_, dvdu_ode_, vu_ode_, semi_, t, periodic_box;
         end
 
         # Add to collection
-        pvd[t] = vtk
+        overwrite || pvd[t] = vtk
```
svchb:
```suggestion
      pvd[t] = vtk
```

src/io/write_vtk.jl
```diff
@@ -153,9 +160,12 @@ function trixi2vtk(system_, dvdu_ode_, vu_ode_, semi_, t, periodic_box;
...
     end
-    vtk_save(pvd)
+
+    overwrite || vtk_save(pvd)
```
svchb, line 168:
```suggestion
    vtk_save(pvd)
```

src/io/write_vtk.jl
```diff
@@ -103,15 +104,21 @@ function trixi2vtk(system_, dvdu_ode_, vu_ode_, semi_, t, periodic_box;
     v = wrap_v(v_ode, system, semi)
     u = wrap_u(u_ode, system, semi)
 
-    file = joinpath(output_directory,
-                    add_underscore_to_optional_prefix(prefix) * "$system_name"
-                    * add_underscore_to_optional_postfix(iter))
+    overwrite = iter == -1 ? true : false
+
+    file_ = joinpath(output_directory,
+                     add_underscore_to_optional_prefix(prefix) * "$system_name")
```
svchb, line 110:
```suggestion
file = joinpath(output_directory,
                    add_underscore_to_optional_prefix(prefix) * "$system_name"
                    * add_underscore_to_optional_postfix(iter))
```

svchb (changes requested):

LasNikas:
I'm confused now. I thought the discussion made it clear that we definitely don't want to create a pvd collection when we overwrite? But  @svchb  's proposed changes seem to suggest otherwise.
@efaulhaber  do you have any thoughts before I change everything again?

src/io/write_vtk.jl
```diff
@@ -103,15 +104,21 @@ function trixi2vtk(system_, dvdu_ode_, vu_ode_, semi_, t, periodic_box;
     v = wrap_v(v_ode, system, semi)
     u = wrap_u(u_ode, system, semi)
 
-    file = joinpath(output_directory,
-                    add_underscore_to_optional_prefix(prefix) * "$system_name"
-                    * add_underscore_to_optional_postfix(iter))
+    overwrite = iter == -1
```
efaulhaber:
```suggestion
    overwrite = isnothing(iter)
```

src/io/write_vtk.jl
efaulhaber: Plus you need to define
```juli
add_underscore_to_optional_postfix(::Nothing) = ""
```

src/io/write_vtk.jl
efaulhaber: This was correct before. We don't want a PVD when overwriting.
1. It's useless if there is only every going to be one file.
2. When overwriting, the file name will not have a suffix, so the file doesn't conflict with another non-overwriting callback. The PVD however would conflict.

efaulhaber (changes requested):

src/io/write_vtk.jl
LasNikas: I don't need this, as I check if this is `overwrite` before

src/callbacks/solution_saving.jl
```diff
@@ -23,6 +23,14 @@ To ignore a custom quantity for a specific system, return `nothing`.
...
 - `save_initial_solution=true`: Save the initial solution.
 - `save_final_solution=true`:   Save the final solution.
+- `overwrite=false`:            If `true`, overwrite the previous set of VTK files instead of
+                                creating a new set of files each interval.
+                                Useful for memory efficiency in large simulations where only
+                                the final results matters. The difference to simply setting
+                                `save_final_solution=true` is that this provides regular checkpoint
+                                backups at each save interval.
+                                If `false` (default), files are not overwritten and an iteration
+                                postfix is appended for each interval.
```
efaulhaber, line 33: Mention that if `true`, the postfix `_current` is appended.

efaulhaber (changes requested):

efaulhaber (approved):

svchb:
> I'm confused now. I thought the discussion made it clear that we definitely don't want to create a pvd collection when we overwrite? But @svchb 's proposed changes seem to suggest otherwise. @efaulhaber do you have any thoughts before I change everything again?

I forgot to submit the review over a month ago. When I came across this again I did not check again.

svchb (approved):

PR #1043, opened 2026-01-13
Make density diffusion in open boundaries settable

### LINKED ISSUE #1072: Zhang open boundaries don't implement density diffusion correctly
https://github.com/trixi-framework/TrixiParticles.jl/blob/5ca7140cd724145a8d3e5a7357d8f15f0f10fdd5/src/schemes/boundary/open_boundary/rhs.jl#L65-L68
This is called for both openboundary-openboundary interaction and openboundary-fluid interaction, but the density diffusion is only implemented for a single system. Here, the density diffusion is called for the open boundary system as `particle_system`, but `neighbor` is an index from a different system, which leads to out-of-bounds errors for Antuono in the renormalized density gradient.

LasNikas:
On main, the density diffusion is taken from the fluid system and cannot be set.
With this PR, one can take density diffusion independent of the fluid.
In addition, the same method of `continuity_equation!` is now called as for the `AbstractFluidSystem`
(I only had to copy the method because of the include order).
It also fixes a bug with the `v_max` calculation for `ParticleShiftingTechnique`.

Closes #1072.


src/schemes/boundary/open_boundary/rhs.jl
```diff
@@ -104,3 +102,31 @@ function pressure_evolution!(dv, particle_system, neighbor_system, v_diff, grad_
...
+    # Artificial density diffusion should only be applied to systems representing a fluid
+    # with the same physical properties i.e. density and viscosity.
+    # TODO: shouldn't be applied to particles on the interface (depends on PR #539)
+    if particle_system === neighbor_system
+        density_diffusion!(dv, density_diffusion(particle_system),
+                           v_particle_system, particle, neighbor,
+                           pos_diff, distance, m_b, rho_a, rho_b, particle_system,
+                           grad_kernel)
+    end
+end
```
svchb: Instead of copying move the conti equation to its own file please.

svchb (changes requested):

src/schemes/boundary/open_boundary/rhs.jl
LasNikas: We don't have a continuity euquation file?

svchb: than create one?

LasNikas: Should we create an extra file for this, or put it in `density_calculators.jl`? I think we had a discussion about this at the very beginning? @efaulhaber

efaulhaber: If you can't fix the include order otherwise, I'd prefer a new file in the fluid directory. `density_calculators.jl` is too abstract IMO.

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -34,6 +34,7 @@ struct OpenBoundarySystem{BM, ELTYPE, NDIMS, IC, FS, FSI, K, ARRAY1D, BC, FC, BZ
     initial_condition                 :: IC
     fluid_system                      :: FS
     fluid_system_index                :: FSI
+    density_calculator                :: ContinuityDensity
```
efaulhaber: This is redundant.

efaulhaber (changes requested):

efaulhaber (changes requested):
Please don't force-push on a PR that is in review. This breaks the "changes since last review" feature (among other things).

svchb (dismissed):

efaulhaber (changes requested):

src/schemes/boundary/open_boundary/rhs.jl
```diff
@@ -59,13 +59,11 @@ function interact!(dv, v_particle_system, u_particle_system,
...
 
-        # Continuity equation
-        @inbounds dv[end, particle] += rho_a / rho_b * m_b * dot(v_diff, grad_kernel)
-
-        density_diffusion!(dv, density_diffusion(particle_system),
-                           v_particle_system, particle, neighbor,
-                           pos_diff, distance, m_b, rho_a, rho_b,
-                           particle_system, grad_kernel)
+        # Propagate `@inbounds` to the continuity equation, which accesses particle data
+        @inbounds continuity_equation!(dv, particle_system, neighbor_system,
```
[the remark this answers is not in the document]
LasNikas: Doesn't this cause heap allocations? Or is this optimized away by the compiler?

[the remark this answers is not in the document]
efaulhaber: An immutable struct with immutable fields lives on the stack. Just like a `Tuple`, which is a struct to bundle multiple scalar values. This struct doesn't even have any fields, so it doesn't require any memory and is just a tool for dispatch, which will be completely optimized away by the compiler.

efaulhaber (approved):

Issue #1048, opened 2026-01-16
Revise fluid CFL condition


efaulhaber:
Currently, we have implemented:
1. Multiple different time step conditions, but only one of them is scaled by the CFL number. When using very stable time integration methods, this leads to situations where a larger time step would still be stable, but the `StepsizeCallback` doesn't make it larger when the CFL number is increased. IMO, the CFL number should linearly scale **all** time step conditions.
2. Only static time step conditions. Nothing based on the current solution. We should investigate if that makes sense, i.e. if a simulation like a Taylor-Green vortex where the velocity tends to zero can increase the time step throughout the simulation.

PR #1049, opened 2026-01-16
Implement stage-level coupling for split integration

efaulhaber:
This PR
- implements stage-level coupling as opposed to the previously used step-level coupling.
- fixes a memory "leak". The sub-integrator was storing the split solution for every single sub-integration call (each fluid time step), which caused massive VRAM allocations for long-running GPU simulations.

In step-level coupling, the fluid is advanced a full step before the structure is advanced a full step. This reduces the stability of the main time integration, reducing the maximum stable time step by a factor of 2 (in my tests with Carpenter-Kennedy).
Stage-level coupling calls the sub-integration in every fluid RK stage. The advantage is that the stability properties of the time integrator are preserved. In the current version, this does not have a significant performance impact and is therefore the default (edit: it does not work for some RK schemes with non-monotonic stage times and is therefore not the default). Only for small ratios (like 2-5x smaller time step for the structure), the step-level coupling might be more efficient. For details on how this is implemented, check out my comment below. The implemented version is the one without "restart", and "predict" is enabled by default but can be disabled via kwarg.

| Δt | Step-level coupling | Stage-level coupling |
| -- | -------------------- | -------------------- |
| 9e-4 | grafik | grafik |
| 1e-3 | grafik | grafik |
| 1.8e-3 |  | grafik |

For 1.9e-3, I get "instability detected" with stage-level coupling. The maximum stable time step of 1.8e-3 with stage-level coupling is the same as when making the ball a moving solid wall boundary (non-elastic).
Note that we cannot do this with a CFL because the `StepsizeCallback` has an upper limit of 1.2e-3 due to #1048.

Note that something is still not right, as the larger time step makes the ball fall deeper with stage-level coupling. (Edit: This is fixed with "predict" below.)


svchb (changes requested):

svchb:
Since you are changing the extension can you please address this issue https://github.com/trixi-framework/TrixiParticles.jl/pull/1047

LasNikas (changes requested):

src/callbacks/split_integration.jl
```diff
@@ -150,118 +182,264 @@ end
...
+    # Update split solution stored in `integrator.p`
+    data.vu_ode_split .= data.integrator.u
+    data.t_ref[] = new_t
+
+    if data.stage_coupling
+        # Tell OrdinaryDiffEq that `u` has NOT been modified.
+        # Theoretically, the TLSPH part has been modified, but in the FSAL case,
+        # the time at the last stage is the same as the step time, so the split integration
+        # above is skipped and `u` is not modified at all.
+        # Therefore, the derivative at the last stage can be reused for the next step.
```
svchb:
```
# Tell OrdinaryDiffEq that `u` was not modified.
# In the FSAL case, the last stage occurs at the step time,
# so the split integration is skipped and `u` remains unchanged.
# Therefore, the derivative at the last stage can be reused for the next step.
```

src/callbacks/split_integration.jl
```diff
@@ -150,118 +182,264 @@ end
...
+    dv_ode_split = split_integrator.p.dv_ode_split
+    semi_large = split_integrator.p.semi_large
 
     @trixi_timeit timer() "split integration" begin
-        # Update quantities that are stored in the systems. These quantities (e.g. pressure)
-        # still have the values from the last stage of the previous step if not updated here.
+        # Copy the solutions back to the original integrator.
+        # We modify `v_ode` and `u_ode`, which is technically not allowed during stages,
+        # hence there are no guarantees about the structure part of `v_ode` and `u_ode`.
+        # By copying the current split integration values, we make sure that it's correct.
```
svchb: unclear

src/callbacks/split_integration.jl
```diff
@@ -150,118 +182,264 @@ end
...
-        u = wrap_u(u_ode, system, semi)
+    # The `TotalLagrangianSPHSystem` doesn't have an `update_final!` method
+end
+
+function self_interaction_split!(dv_ode_split, v_ode_split, u_ode_split, semi_split, semi)
+    # Only loop over (TLSPH) systems in the split integrator
+    foreach_system(semi_split) do system
+        # Construct string for the interactions timer.
+        # Avoid allocations from string construction when no timers are used.
+        # TODO do we need to disable timers in split integration?
```
svchb: todo?

svchb (changes requested):

src/callbacks/split_integration.jl
```diff
@@ -150,132 +183,278 @@ end
...
+        # step. This means that the split integration above is skipped and `u` is not
+        # modified.
+        # Therefore, the RHS from the last stage can be reused for the next step, which is
+        # what `u_modified!` is used for.
+        # TODO is `u_modified` ever relevant for non-FSAL methods?
+        u_modified!(integrator, false)
+    else
+        # Tell OrdinaryDiffEq that `u` has been modified.
+        u_modified!(integrator, true)
+    end
```
svchb: Not a proper way to do this. Have you added an issue?

src/callbacks/split_integration.jl
```diff
@@ -150,132 +183,278 @@ end
...
+        # stage time). In both cases, the scheme is not suitable for stage-level coupling
+        # in the form implemented here.
+        msg = "stage-level coupling with `SplitIntegrationCallback` requires that stage " *
+              "times are monotonically increasing and that no stage time is smaller " *
+              "than the time of the last full step. This time integration scheme can " *
+              "only be used with `stage_coupling=false` when creating the " *
+              "`SplitIntegrationCallback`. It is recommended to use a different time " *
+              "integration scheme with monotonic stage times."
+        throw(ArgumentError(msg))
+    end
```
svchb, line 262: No test

svchb (changes requested):

src/callbacks/split_integration.jl
efaulhaber: Why not?

svchb: Well there is a todo item. And its unclear if this has an impact or not from your statement.

efaulhaber: I see. According to the docs, this should be fine:

> A true will result trigger extra calculations, such as the re-calculation of the derivative the beginning of the next step or Jacobians, which is not necessary if the algorithm is FSAL and no discontinuous change is hit.

This sounds like this is indeed only relevant for FSAL.

src/callbacks/split_integration.jl
efaulhaber: This is tested in the example tests:
```julia
            msg = "stage-level coupling with `SplitIntegrationCallback` requires that"
            @test_throws msg solve(ode, RDPK3SpFSAL49(), maxiters=400,
                                   save_everystep=false, callback=callbacks)
```

src/callbacks/split_integration.jl
svchb: But shouldn't it than be just true?

efaulhaber: I updated the comment and logic. How is this now?

svchb (approved):

LasNikas (approved):

PR #1055, opened 2026-01-22
Introduce foreach_system_wrapped

svchb:
Based on #1053


src/general/semidiscretization.jl
```diff
@@ -175,6 +175,24 @@ end
...
+
+@inline function foreach_system_wrapped(f, semi::Union{NamedTuple, Semidiscretization},
+                                        dv_ode, v_ode, u_ode)
+    return foreach_system(semi) do system
+        f(system,
+          wrap_v(dv_ode, system, semi),
+          wrap_v(v_ode, system, semi),
+          wrap_u(u_ode, system, semi))
+    end
+end
```
efaulhaber: Does it make sense to have an extra function here if we only need it once?

src/general/semidiscretization.jl
```diff
@@ -175,6 +175,24 @@ end
 
 @inline foreach_system(f, systems) = foreach_noalloc(f, systems)
 
+# This is just for readability to loop over all systems with wrapped arrays.
+@inline function foreach_system_wrapped(f, semi::Union{NamedTuple, Semidiscretization},
+                                        v_ode, u_ode)
+    return foreach_system(semi) do system
```
efaulhaber:
```suggestion
    foreach_system(semi) do system
```

src/general/semidiscretization.jl
```diff
@@ -175,6 +175,24 @@ end
...
+        f(system, wrap_v(v_ode, system, semi), wrap_u(u_ode, system, semi))
```
efaulhaber:
```suggestion
        @inline f(system, wrap_v(v_ode, system, semi), wrap_u(u_ode, system, semi))
```

efaulhaber (changes requested):

src/general/semidiscretization.jl
svchb: I know but for consistency it might still be better. Maybe in the future there will be more uses for this?

efaulhaber: I'd prefer adding it once we actually need it.

efaulhaber: What about this now?

svchb: I will remove it tonight

efaulhaber (approved):

PR #1057, opened 2026-01-22
Enable custom interactions between system pairs

svchb:
This PR enables two functionalities by allowing to pass an interaction matrix with either bool values or with a custom interpolation function.


src/general/semidiscretization.jl
```diff
@@ -734,12 +796,23 @@ end
...
+    has_system_interaction(semi, system_index, neighbor_index) || return dv_ode
+
+    dv = wrap_v(dv_ode, system, semi, system_index)
+    v_system = wrap_v(v_ode, system, semi, system_index)
+    u_system = wrap_u(u_ode, system, semi, system_index)
 
-    v_neighbor = wrap_v(v_ode, neighbor, semi)
-    u_neighbor = wrap_u(u_ode, neighbor, semi)
+    v_neighbor = wrap_v(v_ode, neighbor, semi, neighbor_index)
+    u_neighbor = wrap_u(u_ode, neighbor, semi, neighbor_index)
```
efaulhaber: Why did you make this so much more complicated and also changed the wrap functions? This could be a simple change in `system_interaction!`:
```julia
        foreach_system(semi) do neighbor
            has_system_interaction(system, neighbor, semi) || return dv_ode
```

examples/fluid/dam_break_2phase_2d.jl
```diff
@@ -82,11 +82,54 @@ air_system_system = WeaklyCompressibleSPHSystem(air_system;
...
+
+# Semidiscretization order:
+# 1: water fluid, 2: air fluid, 3: water wall, 4: air wall
+function matched_phase_wall_interaction(system_index, neighbor_index)
+    !((system_index == 1 && neighbor_index == 4) ||
+      (system_index == 4 && neighbor_index == 1) ||
+      (system_index == 2 && neighbor_index == 3) ||
+      (system_index == 3 && neighbor_index == 2))
+end
+
```
efaulhaber: This interface seems oddly inconvenient to me. Why not simply make it a matrix?

```suggestion
interaction_matrix = trues(4, 4)
interaction_matrix[1, 4] = interaction_matrix[4, 1] = false
interaction_matrix[2, 3] = interaction_matrix[3, 2] = false
```

efaulhaber (changes requested):
I feel like this could be a change of 10 lines. Why did you change so many internal functions?

src/general/semidiscretization.jl
svchb: I also wanted to enable the capability to define a custom interaction for a system pair.

svchb: So the skipping the whole interaction is just the simplest example of a "custom" interaction.

test/examples/examples_fluid.jl
```diff
@@ -335,6 +335,17 @@
...
 
+    @trixi_testset "fluid/custom_interphase_drag_2d.jl" begin
+        @trixi_test_nowarn trixi_include(@__MODULE__,
+                                         joinpath(examples_dir(), "fluid",
+                                                  "custom_interphase_drag_2d.jl"),
+                                         tspan=(0.0, 0.02))
+        @test sol.retcode == ReturnCode.Success
+        @test semi.interaction_matrix[1, 2] isa InterfacialTangentialDrag
+        @test semi.interaction_matrix[2, 1] isa InterfacialTangentialDrag
+        @test count_rhs_allocations(sol) == 0
```
efaulhaber, line 355: I'm confused. Why does this not produce allocations when accessing a type-unstable matrix?

src/callbacks/split_integration.jl
```diff
@@ -408,16 +407,24 @@ function update_systems_split!(semi_split, v_ode_split, u_ode_split, t)
...
+                                                          v_neighbor,
+                                                          u_neighbor
+            # Construct string for the interactions timer.
+            system_index = system_indices(system, semi)
+            neighbor_index = system_indices(neighbor, semi)
+            timer_str = "$(timer_name(system))$system_index-$(timer_name(neighbor))$neighbor_index"
+
+            @trixi_timeit timer() timer_str begin
+                apply_system_interaction!(dv, v, u, v_neighbor, u_neighbor,
+                                          system, neighbor, semi; integrate_tlsph=true)
```
efaulhaber: This is not correct. We don't want tlsph-* interaction here, only tlsph-tlsph self-interaction. You added a loop where no loop was before.

I'm surprised the tests don't fail. We should have tests that detect such errors.

examples/fluid/custom_interphase_drag_2d.jl
```diff
@@ -0,0 +1,134 @@
+# ==========================================================================================
+# 2D Interphase Drag Example
+#
+# This example demonstrates a custom ordered-pair interaction in the `interaction_matrix`.
```
efaulhaber: What is an "ordered-pair interaction"? Please make this more understandable.

examples/fluid/custom_interphase_drag_2d.jl
```diff
@@ -0,0 +1,134 @@
...
+    # phases, then add the custom drag term below.
+    TrixiParticles.interact!(dv, v_system, u_system, v_neighbor, u_neighbor,
+                             system, neighbor, semi; eachparticle, kwargs...)
+
+    system_coords = TrixiParticles.current_coordinates(u_system, system)
+    neighbor_coords = TrixiParticles.current_coordinates(u_neighbor, neighbor)
+
+    TrixiParticles.foreach_point_neighbor(system, neighbor, system_coords,
+                                          neighbor_coords,
+                                          semi;
```
efaulhaber:
```suggestion
                                          neighbor_coords, semi;
```

examples/fluid/custom_interphase_drag_2d.jl
```diff
@@ -0,0 +1,134 @@
...
+                                          points=eachparticle) do particle,
+                                                                  neighbor_particle,
+                                                                  pos_diff,
+                                                                  distance
+        v_a = TrixiParticles.current_velocity(v_system, system, particle)
+        v_b = TrixiParticles.current_velocity(v_neighbor, neighbor, neighbor_particle)
+        velocity_difference = v_b - v_a
+        normal_velocity_difference = sum(velocity_difference[i] *
+                                         drag.interface_normal[i]
+                                         for i in 1:ndims(system))
```
efaulhaber: That is just a dot product?

```suggestion
        normal_velocity_difference = velocity_difference' * drag.interface_normal
```

examples/fluid/custom_interphase_drag_2d.jl
```diff
@@ -0,0 +1,134 @@
...
+                              normal_velocity_difference * drag.interface_normal[i]
+            dv[i, particle] += acceleration_factor * tangential_slip
+        end
+    end
+
+    return dv
+end
+
+# `true` keeps the default interaction, while a callable handles the interaction for that
+# ordered pair. This callable calls the default interaction before adding drag. Since the
```
efaulhaber:
```suggestion
# `true` keeps the default interaction, while a method handles the interaction for that
# ordered pair. This method calls the default interaction before adding drag. Since the
```

examples/fluid/custom_interphase_drag_2d.jl
```diff
@@ -0,0 +1,134 @@
...
+# matrix is ordered, reciprocal coupling needs both entries.
+interfacial_drag = InterfacialTangentialDrag(1.0e3, (0.0, 1.0))
+interaction_matrix = Matrix{Union{Bool, typeof(interfacial_drag)}}(trues(2, 2))
+interaction_matrix[1, 2] = interfacial_drag
+interaction_matrix[2, 1] = interfacial_drag
```
efaulhaber:
```suggestion
interaction_matrix = [true interface_drag; interface_drag true]
```
As discussed, union types should not be handled by the user.

examples/fluid/dam_break_2phase_2d.jl
```diff
@@ -82,11 +83,59 @@ air_system_system = WeaklyCompressibleSPHSystem(air_system;
...
+# Use two copies of the wall geometry with hydrodynamic properties matched to the adjacent
+# fluid phase. The interaction matrix below disables the cross-phase wall interactions.
+water_boundary_model = BoundaryModelDummyParticles(tank.boundary.density,
+                                                   tank.boundary.mass,
+                                                   boundary_density_calculator,
+                                                   smoothing_kernel, smoothing_length;
+                                                   state_equation=state_equation,
+                                                   correction=nothing,
+                                                   reference_particle_spacing=0,
```
efaulhaber: Are these overwritten somewhere in the tests? If not, please remove them.

src/general/semidiscretization.jl
```diff
@@ -168,21 +286,66 @@ end
...
+
+# `semi` owns the interaction matrix; `semi_wrap` only selects the array layout for wrapping.
+# Split integration intentionally passes different values here.
+@inline function foreach_interacting_system_wrapped(f, system, semi, v_ode, u_ode,
+                                                    semi_wrap)
+    foreach_interacting_system(system, semi_wrap, semi) do neighbor_system
+        @inline f(neighbor_system, wrap_v(v_ode, neighbor_system, semi_wrap),
+                  wrap_u(u_ode, neighbor_system, semi_wrap))
+    end
+end
```
efaulhaber: I don't think this improves readability. The opposite even. I think that's yet another function where the user stops and has to check what it even does.
```julia
foreach_system_wrapped(semi, v_ode, u_ode) do neighbor_system, v, u
    has_system_interaction(system, neighbor_system, semi) || return
    ...
end
```
is more readable IMO than
```julia
foreach_interacting_system_wrapped(f, system, semi, v_ode, u_ode) do neighbor_system, v, u
    ...
end
```
because it explicitly says "skip systems that don't have interaction".

src/callbacks/split_integration.jl
```diff
@@ -430,10 +437,13 @@ function other_interaction_split!(dv_ode_split, semi, v_ode, u_ode, semi_split)
...
-        foreach_system_wrapped(semi, v_ode, u_ode) do neighbor, v_neighbor, u_neighbor
-            if system === neighbor
-                # Only compute interaction with other systems
+        # Loop over all interacting neighbors in the big integrator
+        foreach_interacting_system_wrapped(system, semi, v_ode,
+                                           u_ode) do neighbor,
+                                                     v_neighbor,
+                                                     u_neighbor
+            if neighbor isa TotalLagrangianSPHSystem
+                # TLSPH self-interactions are integrated with the split state.
```
efaulhaber: This is also not the same as `system === neighbor`. We only want self-interaction here.

src/general/corrections.jl
```diff
@@ -139,11 +139,13 @@ function compute_shepard_coeff!(system, system_coords, v_ode, u_ode, semi,
                                 kernel_correction_coefficient)
     set_zero!(kernel_correction_coefficient)
 
-    # Use all other systems for the density summation
-    @trixi_timeit timer() "compute correction value" foreach_system(semi) do neighbor_system
-        u_neighbor_system = wrap_u(u_ode, neighbor_system, semi)
-        v_neighbor_system = wrap_v(v_ode, neighbor_system, semi)
-
+    # Use enabled neighbor systems for the correction value.
+    @trixi_timeit timer() "compute correction value" foreach_interacting_system_wrapped(system,
```
efaulhaber:
```suggestion
    @trixi_timeit timer() "compute correction value" begin
        foreach_interacting_system_wrapped(system,
```
Also use `begin` for the other two changes below for readability.

src/general/semidiscretization.jl
```diff
@@ -15,9 +15,17 @@ The semidiscretization couples the passed systems to one simulation.
                             and the examples below for more details.
                             To use a periodic domain, pass a [`PeriodicBox`](@ref) to the
                             neighborhood search.
-- `threaded_nhs_update=true`:   Can be used to deactivate thread parallelization in the neighborhood search update.
-                                This can be one of the largest sources of variations between simulations
-                                with different thread numbers due to particle ordering changes.
+- `parallelization_backend=PolyesterBackend()`: Backend used for thread-parallel loops,
+                            including the generic neighborhood-search update paths.
+                            Pass `SerialBackend()` to disable thread parallelization.
```
efaulhaber:
```suggestion
- `parallelization_backend=PolyesterBackend()`: Backend used for thread-parallel loops.
                            Pass `SerialBackend()` to disable thread parallelization.
                            See [the docs](@ref gpu_support) on how to use GPU backends.
```
What even is a "generic neighborhood-search update path"?

src/general/semidiscretization.jl
```diff
@@ -15,9 +15,17 @@ The semidiscretization couples the passed systems to one simulation.
...
+- `interaction_matrix=trues(n_systems, n_systems)`: Matrix controlling ordered system-pair
+                            interactions after filtering out `nothing` systems. Rows refer
+                            to the system being updated and columns to the neighbor system.
+                            Entries are `true` for the default interaction, `false` to
+                            disable the ordered pair, or a callable custom interaction with
+                            the same arguments as `interact!(...; kwargs...)`. Disabled
+                            pairs are skipped in RHS and neighbor-derived state updates, but
+                            neighborhood searches are still built.
```
efaulhaber:
```suggestion
- `interaction_matrix=trues(n_systems, n_systems)`: Matrix controlling interactions of each
                            system-neighbor pair after filtering out `nothing` systems.
                            `A[i, j] == true` uses the default interaction for computing forces
                            on system `i` by particles of system `j`. `A[i, j] == false` disables
                            the interaction. Set a matrix entry to a method with
                            the same arguments as `interact!(...; kwargs...)` to use a custom interaction function for this forces computation. Disabled
                            pairs are skipped in RHS and neighbor-derived state updates, but
                            neighborhood searches are still built.
```
What do you mean by "neighbor-derived state updates"? It is not clear to me what that means when reading this docstring.

Why are the neighborhood searches still built? Are they ever used? If so, why? The point is that two systems don't see each other, no?

src/general/semidiscretization.jl
```diff
@@ -36,6 +44,12 @@ semi = Semidiscretization(fluid_system, boundary_system,
 semi = Semidiscretization(fluid_system, boundary_system,
                           neighborhood_search=nothing)
 
+interaction_matrix = trues(2, 2)
+interaction_matrix[1, 2] = false # fluid_system skips system-pair interactions with boundary_system
```
efaulhaber:
```suggestion
interaction_matrix[1, 2] = false # `fluid_system` skips interactions with `boundary_system`
```

src/general/semidiscretization.jl
```diff
@@ -46,14 +60,17 @@ semi = Semidiscretization(fluid_system, boundary_system,
 │ total #particles: ………………………………… 636                                                              │
 │ eltype: …………………………………………………………… Float64                                                          │
 │ coordinates eltype: …………………………… Float64                                                          │
+│ interaction matrix: …………………………… 1 disabled, 0 custom                                             │
+│ disabled pairs: ……………………………………… 1 -> 2                                                           │
```
efaulhaber:
```suggestion
│ disabled interactions: ……………………………………… 1 -> 2                                                           │
```

src/general/semidiscretization.jl
```diff
@@ -62,20 +79,103 @@ struct Semidiscretization{BACKEND, S, RU, RV, NS, UCU, IT}
...
 end
 
+function default_interaction_matrix(systems::Tuple)
+    n_systems = length(systems)
+    return trues(n_systems, n_systems)
+end
+
+function create_interaction_matrix(::Nothing, systems::Tuple)
+    return default_interaction_matrix(systems)
+end
```
efaulhaber:
```suggestion
function create_interaction_matrix(::Nothing, systems)
    n_systems = length(systems)
    return trues(n_systems, n_systems)
end
```
Unnecessary function and unnecessary dispatch.

src/general/semidiscretization.jl
```diff
@@ -62,20 +79,103 @@ struct Semidiscretization{BACKEND, S, RU, RV, NS, UCU, IT}
...
+
+function create_interaction_matrix(interaction_matrix, systems::Tuple)
```
efaulhaber:
```suggestion
function create_interaction_matrix(interaction_matrix, systems)
```

src/general/semidiscretization.jl
```diff
@@ -62,20 +79,103 @@ struct Semidiscretization{BACKEND, S, RU, RV, NS, UCU, IT}
...
+                            "$(size(interaction_matrix))"))
+    end
+
+    # Validate values before looking at the declared element type. This lets users pass
+    # abstract containers such as `Matrix{Any}` while still rejecting invalid entries.
+    for entry in interaction_matrix
+        if !is_interaction_entry(entry)
+            throw(ArgumentError("`interaction_matrix` entries must be `true`, `false`, " *
+                                "or callable custom interactions, but found " *
+                                "`$(typeof(entry))`"))
```
efaulhaber:
```suggestion
                                "or methods, but found `$(typeof(entry))`"))
```

src/general/semidiscretization.jl
```diff
@@ -62,20 +79,103 @@ struct Semidiscretization{BACKEND, S, RU, RV, NS, UCU, IT}
...
+        end
+    end
+
+    # Materialize to a one-based `Matrix`. Rebuild abstractly typed matrices from the
+    # concrete entry types to avoid dynamic dispatch and allocations in pairwise loops.
```
efaulhaber:
```suggestion
    # Rebuild abstractly typed matrices from the concrete entry types
    # to avoid dynamic dispatch and allocations in pairwise loops.
```
Matrices are always one-based?

src/general/semidiscretization.jl
```diff
@@ -62,20 +79,103 @@ struct Semidiscretization{BACKEND, S, RU, RV, NS, UCU, IT}
...
+    pairs = String[]
+    for system_index in axes(semi.interaction_matrix, 1),
+        neighbor_index in axes(semi.interaction_matrix, 2)
+        interaction = semi.interaction_matrix[system_index, neighbor_index]
+        interaction isa Bool && continue
+        push!(pairs, "$system_index -> $neighbor_index ($(nameof(typeof(interaction))))")
+    end
+
+    return isempty(pairs) ? nothing : join(pairs, ", ")
+end
```
efaulhaber: Can you please move those to the bottom where the `show` function is?

Edit: I just realized the `show` function is not at the bottom of this file. Can you please move the `show` functions to the bottom as well for better readability of the stuff that's actually important? Can be a new PR.

src/general/semidiscretization.jl
```diff
@@ -168,21 +286,66 @@ end
     return index
 end
 
+@inline has_system_interaction(system, neighbor_system, semi) = true
+
+@inline system_interaction(system, neighbor_system, semi) = true
+
```
efaulhaber: What is the purpose of these functions? Why is `semi` not a `Semidiscretization`?

src/general/semidiscretization.jl
```diff
@@ -62,20 +79,103 @@ struct Semidiscretization{BACKEND, S, RU, RV, NS, UCU, IT}
...
+end
+
+@inline is_interaction_entry(entry::Bool) = true
+
+function is_interaction_entry(entry)
+    return !isempty(methods(entry))
+end
+
+@inline is_enabled_interaction(entry::Bool) = entry
+@inline is_enabled_interaction(entry) = true
```
efaulhaber: Please move these functions down to where they are used.

src/general/semidiscretization.jl
```diff
@@ -362,8 +525,8 @@ end
     range = ranges_v[system_indices(system, semi)]
 
     @boundscheck begin
-        expected = v_nvariables(system) * n_integrated_particles(system)
         if length(range) != v_nvariables(system) * n_integrated_particles(system)
+            expected = v_nvariables(system) * n_integrated_particles(system)
```
efaulhaber:
```suggestion
        expected = v_nvariables(system) * n_integrated_particles(system)
        if length(range) != expected
```

src/schemes/boundary/wall_boundary/dummy_particles.jl
```diff
@@ -496,11 +496,15 @@ function compute_pressure!(boundary_model,
...
 
-    # Use all other systems for the pressure extrapolation
-    @trixi_timeit timer() "compute boundary pressure" foreach_system(semi) do neighbor_system
-        v_neighbor_system = wrap_v(v_ode, neighbor_system, semi)
-        u_neighbor_system = wrap_u(u_ode, neighbor_system, semi)
-
+    # Use enabled neighbor systems for the pressure extrapolation. This lets phase-specific
+    # walls isolate their auxiliary pressure state with the same ordered interaction matrix
+    # used for pairwise RHS dispatch.
+    @trixi_timeit timer() "compute boundary pressure" foreach_interacting_system_wrapped(system,
```
efaulhaber: Use `begin`.

src/schemes/fluid/entropically_damped_sph/system.jl
```diff
@@ -334,11 +334,13 @@ function update_average_pressure!(system, ::Val{true}, v_ode, u_ode, semi)
 
     u = wrap_u(u_ode, system, semi)
 
-    # Use all other systems for the average pressure
-    @trixi_timeit timer() "compute average pressure" foreach_system(semi) do neighbor_system
-        u_neighbor_system = wrap_u(u_ode, neighbor_system, semi)
-        v_neighbor_system = wrap_v(v_ode, neighbor_system, semi)
-
+    # Use enabled neighbor systems for the average pressure.
+    @trixi_timeit timer() "compute average pressure" foreach_interacting_system_wrapped(system,
```
efaulhaber: Use `begin`.

src/schemes/fluid/surface_normal_sph.jl
```diff
@@ -202,10 +202,12 @@ function compute_surface_normal!(system::AbstractFluidSystem,
     set_zero!(cache.neighbor_count)
 
     # TODO: if color values are set only different systems need to be called
-    @trixi_timeit timer() "compute surface normal" foreach_system(semi) do neighbor_system
-        u_neighbor_system = wrap_u(u_ode, neighbor_system, semi)
-        v_neighbor_system = wrap_v(v_ode, neighbor_system, semi)
-
+    @trixi_timeit timer() "compute surface normal" foreach_interacting_system_wrapped(system,
```
efaulhaber: Use `begin`.

src/schemes/fluid/surface_normal_sph.jl
```diff
@@ -275,10 +277,12 @@ function compute_curvature!(system::AbstractFluidSystem,
     # Reset surface curvature
     set_zero!(cache.curvature)
 
-    @trixi_timeit timer() "compute surface curvature" foreach_system(semi) do neighbor_system
-        u_neighbor_system = wrap_u(u_ode, neighbor_system, semi)
-        v_neighbor_system = wrap_v(v_ode, neighbor_system, semi)
-
+    @trixi_timeit timer() "compute surface curvature" foreach_interacting_system_wrapped(system,
```
efaulhaber: Use `begin`.

src/general/semidiscretization.jl
```diff
@@ -757,8 +916,46 @@ end
...
+    return apply_interaction!(interaction, dv, v_system, u_system, v_neighbor,
+                              u_neighbor, system, neighbor, semi; kwargs...)
+end
+
+@inline function apply_system_interaction!(dv, v_system, u_system, v_neighbor,
+                                           u_neighbor, system::TotalLagrangianSPHSystem,
+                                           neighbor, semi;
+                                           integrate_tlsph=semi.integrate_tlsph[],
+                                           kwargs...)
+    integrate_tlsph || return dv
```
efaulhaber, line 875: What is the point of checking here only to pass `integrate_tlsph` to the next function? Either remove this method and leave the check in `interact!` or remove the check and the kwarg from `interact!`.

test/general/semidiscretization.jl
```diff
@@ -21,12 +21,52 @@
     TrixiParticles.compact_support(::System1, neighbor) = 0.2
     TrixiParticles.compact_support(::System2, neighbor) = 0.2
 
+    function TrixiParticles.interact!(dv, v_system, u_system, v_neighbor, u_neighbor,
```
efaulhaber, line 24: Please move all of these into the interaction matrix testset where they are actually used.

test/general/semidiscretization.jl
```diff
@@ -143,6 +238,218 @@
...
+            u = TrixiParticles.wrap_u(u_ode, system, semi)
+
+            TrixiParticles.compute_correction_values!(system,
+                                                      TrixiParticles.system_correction(system),
+                                                      u, v_ode, u_ode, semi)
+
+            return copy(system.cache.kernel_correction_coefficient), semi
+        end
+
+        @testset "disabled pairs skip ordered RHS dispatch" begin
```
efaulhaber:
```suggestion
        @testset "disabled interactions are skipped" begin
```

test/general/semidiscretization.jl
```diff
@@ -143,6 +238,218 @@
...
+
+            @test !TrixiParticles.has_system_interaction(semi.systems[1],
+                                                         semi.systems[2], semi)
+            @test TrixiParticles.has_system_interaction(semi.systems[2],
+                                                        semi.systems[1], semi)
+            @test system_dv(dv_ode, semi, 1)[1, 1] == 1
+            @test system_dv(dv_ode, semi, 2)[1, 1] == 1100
+        end
+
+        @testset "callable entries replace ordered RHS dispatch" begin
```
efaulhaber: This sounds extremely cryptic.

```suggestion
        @testset "custom interaction function" begin
```

test/general/semidiscretization.jl
```diff
@@ -143,6 +238,218 @@
...
+
+            semi.integrate_tlsph[] = false
+            TrixiParticles.self_interaction_split!(dv_ode_split, v_ode_split,
+                                                   u_ode_split, semi_split, semi)
+
+            @test interaction.calls[] == 1
+            @test interaction.integrate_tlsph_seen[]
+            @test system_dv(dv_ode_split, semi_split, 1)[1, 1] == 50
+            @test system_dv(dv_ode_split, semi_split, 2)[1, 1] == 0
+        end
```
efaulhaber: There should only be self-interaction.

efaulhaber (changes requested):

examples/fluid/custom_interphase_drag_2d.jl
svchb: I think this is harder to understand.

src/general/corrections.jl
svchb: I guess you started this with an older version??? Because I don't understand?

src/schemes/boundary/wall_boundary/dummy_particles.jl
svchb: ?

src/schemes/fluid/entropically_damped_sph/system.jl
svchb: ?

src/schemes/fluid/surface_normal_sph.jl
svchb: ?

src/schemes/fluid/surface_normal_sph.jl
svchb: ?

examples/fluid/custom_interphase_drag_2d.jl
efaulhaber: Then at least make it `Matrix{Any}` to prevent users from thinking that the type is relevant.

src/callbacks/mechanical_work_calculator.jl
```diff
@@ -179,19 +179,20 @@ function update_mechanical_work_calculator!(work, system, eachparticle,
...
+                                                      neighbor_system isa
+                                                      AbstractFluidSystem),
+                                  semi, v_ode,
+                                  u_ode) do neighbor_system,
+                                            v_neighbor,
+                                            u_neighbor
+            apply_system_interaction!(dv, v, u, v_neighbor, u_neighbor,
+                                      system, neighbor_system, semi,
+                                      integrate_tlsph=true, # Required when using split integration
+                                      eachparticle=eachparticle)
```
efaulhaber:
```suggestion
        foreach_system_wrapped(semi, v_ode,
                               u_ode) do neighbor_system, v_neighbor, u_neighbor
            if only_compute_force_on_fluid && !(neighbor_system isa AbstractFluidSystem)
                # Not a fluid system, ignore this system.
                return dv
            end
            
            if !has_system_interaction(system, neighbor_system, semi)
                # No interaction between these systems.
                return dv
            end
                                            
            apply_system_interaction!(dv, v, u, v_neighbor, u_neighbor,
                                      system, neighbor_system, semi,
                                      integrate_tlsph=true, # Required when using split integration
                                      eachparticle=eachparticle)
```
I think this is much easier to read. This is something I can easily understand by skimming over the code. The `foreach_system_wrapped_if` is another unnecessary obfuscation that requires users to look up what this function does. Even worse than the `foreach_interacting_system_wrapped` because it doesn't even make the code look more readable if you know what this function does.

src/general/semidiscretization.jl
```diff
@@ -15,9 +15,23 @@ The semidiscretization couples the passed systems to one simulation.
...
+                            See [the docs](@ref gpu_support) on how to use GPU backends.
+- `interaction_matrix=trues(n_systems, n_systems)`: Matrix controlling ordered system-pair
+                            interactions after filtering out `nothing` systems. Rows refer
+                            to the system being updated and columns to the neighbor system.
+                            `A[i, j] == true` uses the default interaction for computing forces
+                            on system `i` by particles of system `j`. `A[i, j] == false` disables
+                            the interaction. Set a matrix entry to a method with
+                            the same arguments as `interact!(...; kwargs...)` to use a custom
+                            interaction function for this force computation. Disabled
+                            pairs are skipped in the RHS and in auxiliary neighbor loops such as
```
efaulhaber:
```suggestion
                            pairs are skipped in the forces computation and in auxiliary neighbor loops such as
```

src/general/semidiscretization.jl
efaulhaber: Not resolved.

src/general/semidiscretization.jl
```diff
@@ -713,9 +863,11 @@ end
 function system_interaction!(dv_ode, v_ode, u_ode, semi)
     reset_interaction_caches!(semi)
 
-    # Call `interact!` for each pair of systems
+    # Call `interact!` for each ordered pair of systems.
     foreach_system(semi) do system
         foreach_system(semi) do neighbor
+            has_system_interaction(system, neighbor, semi) || return
```
efaulhaber:
```suggestion
            has_system_interaction(system, neighbor, semi) || return dv_ode
```
For type stability.

src/schemes/fluid/surface_normal_sph.jl
efaulhaber: I guess you pushed while I was reviewing.

test/general/semidiscretization.jl
```diff
@@ -143,6 +167,323 @@
...
+        TestInteraction() = TestInteraction(Ref(0))
+
+        function (interaction::TestInteraction)(dv, v_system, u_system, v_neighbor,
+                                                u_neighbor, system, neighbor, semi;
+                                                kwargs...)
+            interaction.calls[] += 1
+            dv[1, 1] += 50
+            return dv
+        end
```
efaulhaber:
```suggestion
        function test_interaction(dv, v_system, u_system, v_neighbor,
                                                u_neighbor, system, neighbor, semi;
                                                kwargs...)
            dv[1, 1] += 10_000
            return dv
        end
```
I don't understand why you need a second way to count calls if you already count by checking the result of `dv`.

test/general/semidiscretization.jl
```diff
@@ -143,6 +167,323 @@
...
+                                                     semi, v_short, u_short) do system, v, u
+                push!(visited, system)
+                @test size(v) == (TrixiParticles.v_nvariables(system),
+                       TrixiParticles.n_integrated_particles(system))
+                @test size(u) == (TrixiParticles.u_nvariables(system),
+                       TrixiParticles.n_integrated_particles(system))
+            end
+
+            @test visited == Any[semi.systems[1]]
+        end
```
efaulhaber: Wrapping is basically free, so this is really not necessary.

efaulhaber (changes requested):

src/general/semidiscretization.jl
```diff
@@ -168,19 +201,59 @@ end
...
+@inline function foreach_system(f, semi::Union{NamedTuple, Semidiscretization},
+                                arg1, arg2, args...)
+    return foreach_noalloc(f, semi.systems, arg1, arg2, args...)
+end
+
 @inline foreach_system(f, systems) = foreach_noalloc(f, systems)
+@inline foreach_system(f, systems, arg1, arg2,
+                       args...) = foreach_noalloc(f, systems,
+                                                  arg1, arg2,
+                                                  args...)
```
efaulhaber: Why two versions? Why do you have one with zero additional args and one with 2+, what about 1 additional arg?

```suggestion
@inline function foreach_system(f, semi::Union{NamedTuple, Semidiscretization}, args...)
    return foreach_noalloc(f, semi.systems, args...)
end

@inline foreach_system(f, systems, args...) = foreach_noalloc(f, systems, args...)
```

src/general/semidiscretization.jl
```diff
@@ -168,19 +201,59 @@ end
...
+# This avoids wrapping `foreach_system` in another closure while still keeping every
+# heterogeneous system type visible to inference at the call site.
+@inline function _foreach_system_wrapped(system, f, semi_wrap, v_ode, u_ode)
+    v = wrap_v(v_ode, system, semi_wrap)
+    u = wrap_u(u_ode, system, semi_wrap)
+
+    # This mirrors the explicit inline callback invocation used by `foreach_noalloc`.
+    # Without it, large do-block environments can allocate even though the tuple traversal
+    # itself is allocation-free.
+    @inline f(system, v, u)
```
efaulhaber: I don't understand these comments. Are they AI-generated?

```suggestion
# Define an explicit function to be passed to `foreach_system` instead of using a closure
# because nested closures can cause allocations.
@inline function _foreach_system_wrapped(system, f, semi_wrap, v_ode, u_ode)
    v = wrap_v(v_ode, system, semi_wrap)
    u = wrap_u(u_ode, system, semi_wrap)

    @inline f(system, v, u)
```

src/util.jl
```diff
@@ -17,6 +17,21 @@ end
...
+    element = first(collection)
+    remaining_collection = Base.tail(collection)
+
+    @inline func(element, arg1, arg2, args...)
+
+    # Process remaining collection
+    foreach_noalloc(func, remaining_collection, arg1, arg2, args...)
+end
+
+@inline foreach_noalloc(func, collection::Tuple{}, arg1, arg2, args...) = nothing
```
efaulhaber: 1. I also don't understand this comment.
2. Why two versions again?

```suggestion
# Same as `foreach`, but it is unrolled by the compiler for small input tuples.
# Additional arguments are just passed through to `func` unchanged.
@inline function foreach_noalloc(func, collection, args...)
    element = first(collection)
    remaining_collection = Base.tail(collection)

    @inline func(element, args...)

    # Process remaining collection
    foreach_noalloc(func, remaining_collection, args...)
end

@inline foreach_noalloc(func, collection::Tuple{}, args...) = nothing
```

efaulhaber (changes requested):

test/general/semidiscretization.jl
efaulhaber: Not resolved?

test/general/semidiscretization.jl
```diff
@@ -63,6 +85,83 @@
         @test semi.neighborhood_searches == nhs
     end
 
+    @testset verbose=true "foreach_system" begin
```
efaulhaber, line 116: Why is this testset so massively overengineered? That's over 100 lines for a super simple function. And why is a bunch of stuff for this testset defined above before another random testset?

This is sufficient:
```julia
systems = (1, 2)
visited = [0, 0]
TrixiParticles.foreach_system(systems) do system
    visited[system] += 1
end
@test visited == [1, 1]

# Additional arguments should be passed through unchanged.
TrixiParticles.foreach_system(systems, 42) do system, arg
    visited[system] += arg
end
@test visited == [43, 43]
```

test/general/util.jl
```diff
@@ -1,3 +1,44 @@
+@testset verbose=true "foreach_noalloc" begin
```
efaulhaber, line 1: Same here. This is sufficient, considering that the args version is already tested through `foreach_system`:
```julia
collection1 = (1, 2)
visited = [0, 0]
TrixiParticles.foreach_noalloc(collection1) do collection
    visited[collection] += 1
end
@test visited == [1, 1]

collection2 = (3, 4)
visited = []
TrixiParticles.foreach_noalloc_zip(collection1, collection2) do (i, j)
    push!(visited, (i, j))
end
@test visited == [(1, 3), (2, 4)]
```

efaulhaber (changes requested):

efaulhaber (dismissed):

src/general/semidiscretization.jl
```diff
@@ -168,19 +201,53 @@ end
...
+
+@inline function foreach_system_wrapped(f, _semi::Union{NamedTuple, Semidiscretization},
+                                        v_ode, u_ode, semi_wrap)
+    return foreach_system(_foreach_system_wrapped, semi_wrap, f, semi_wrap, v_ode,
+                          u_ode)
+end
+
+# Define an explicit function to be passed to `foreach_system` instead of using a closure
+# because nested closures can cause allocations.
+@inline function _foreach_system_wrapped(system, f, semi_wrap, v_ode, u_ode)
```
efaulhaber:
```suggestion
    return foreach_system(_foreach_system_wrapped, semi_wrap, 
                          (f, semi_wrap, v_ode, u_ode))
end

# Define an explicit function to be passed to `foreach_system` instead of using a closure
# because nested closures can cause allocations.
@inline function _foreach_system_wrapped(system, data)
    f, semi_wrap, v_ode, u_ode = data
    
```
Here is a simple solution to avoid the allocations in Julia 1.10 when it doesn't specialize on 4 extra arguments for some reason.

src/general/semidiscretization.jl
```diff
@@ -168,19 +201,53 @@ end
...
+    return foreach_noalloc(f, semi.systems, args...)
 end
 
-@inline foreach_system(f, systems) = foreach_noalloc(f, systems)
+# Inference limits recursion through these methods on Julia 1.10, causing runtime
+# dispatch. Recursion here always terminates (`collection` shrinks each level),
+# so tell the compiler it may keep inferring through it.
+for m in methods(foreach_noalloc)
+    m.recursion_relation = (method, topmost, sig, topmost_sig) -> true
+end
```
efaulhaber: Then you can undo all of that sketchy stuff here.

src/util.jl
```diff
@@ -4,20 +4,23 @@
     return Base.FastMath.div_fast(x, y)
 end
 
-# Same as `foreach`, but it is unrolled by the compiler for small input tuples
-@inline function foreach_noalloc(func, collection)
+# Same as `foreach`, but it is unrolled by the compiler for small input tuples. Additional
+# arguments are passed unchanged to `func` for every element.
```
efaulhaber:
```suggestion
# arguments are passed unchanged to `func` for every element.
# Note that this might allocate for 3 or more `args` in Julia 1.10.
```
And a small warning here to prevent future debugging sessions.

efaulhaber (changes requested):

efaulhaber (approved):

src/general/semidiscretization.jl
```diff
@@ -168,19 +201,49 @@ end
...
-    foreach_system(semi) do system
-        @inline f(system, wrap_v(v_ode, system, semi), wrap_u(u_ode, system, semi))
-    end
+    return foreach_system_wrapped(f, semi, v_ode, u_ode, semi)
+end
+
+@inline function foreach_system_wrapped(f, _semi::Union{NamedTuple, Semidiscretization},
+                                        v_ode, u_ode, semi_wrap)
+    return foreach_system(_foreach_system_wrapped, semi_wrap,
+                          (f, semi_wrap, v_ode, u_ode))
```
efaulhaber:
```suggestion
    # Wrap extra arguments in a tuple to avoid allocations because Julia 1.10 avoids
    # specialization for too many arguments, which causes dynamic dispatch.
    # Make sure that `f` is passed separately (not inside the tuple) or Julia will
    # not specialize on the type of `f` and allocate through dynamic dispatch.
    return foreach_system(_foreach_system_wrapped, semi_wrap, f
                          (semi_wrap, v_ode, u_ode))
```

src/general/semidiscretization.jl
```diff
@@ -168,19 +201,49 @@ end
...
+end
+
+# Define an explicit function to be passed to `foreach_system` instead of using a closure
+# because nested closures can cause allocations.
+@inline function _foreach_system_wrapped(system, data)
+    f, semi_wrap, v_ode, u_ode = data
```
efaulhaber:
```suggestion
@inline function _foreach_system_wrapped(system, f, data)
    semi_wrap, v_ode, u_ode = data
```

efaulhaber (changes requested):

src/general/semidiscretization.jl
```diff
@@ -168,19 +201,53 @@ end
...
+    return foreach_system_wrapped(f, semi, v_ode, u_ode, semi)
+end
+
+@inline function foreach_system_wrapped(f, _semi::Union{NamedTuple, Semidiscretization},
+                                        v_ode, u_ode, semi_wrap)
+    # Wrap extra arguments in a tuple to avoid allocations because Julia 1.10 avoids
+    # specialization for too many arguments, which causes dynamic dispatch.
+    # Make sure that `f` is passed separately (not inside the tuple) or Julia will
+    # not specialize on the type of `f` and allocate through dynamic dispatch.
+    return foreach_system(_foreach_system_wrapped, semi_wrap, f
```
efaulhaber:
```suggestion
    return foreach_system(_foreach_system_wrapped, semi_wrap, f,
```

src/general/semidiscretization.jl
```diff
@@ -168,19 +201,53 @@ end
...
 # This is just for readability to loop over all systems without allocations
-@inline function foreach_system(f, semi::Union{NamedTuple, Semidiscretization})
-    return foreach_noalloc(f, semi.systems)
+@inline function foreach_system(f, semi::Union{NamedTuple, Semidiscretization}, args...)
+    return foreach_noalloc(f, semi.systems, args...)
 end
 
-@inline foreach_system(f, systems) = foreach_noalloc(f, systems)
+@inline foreach_system(f, systems, args...) = foreach_noalloc(f, systems, args...)
+
```
efaulhaber, line 225:
```suggestion
```

src/general/semidiscretization.jl
```diff
@@ -15,9 +15,23 @@ The semidiscretization couples the passed systems to one simulation.
                             and the examples below for more details.
                             To use a periodic domain, pass a [`PeriodicBox`](@ref) to the
                             neighborhood search.
-- `threaded_nhs_update=true`:   Can be used to deactivate thread parallelization in the neighborhood search update.
-                                This can be one of the largest sources of variations between simulations
-                                with different thread numbers due to particle ordering changes.
+- `parallelization_backend=PolyesterBackend()`: Backend used for thread-parallel loops.
+                            Pass `SerialBackend()` to disable thread parallelization.
+                            See [the docs](@ref gpu_support) on how to use GPU backends.
+- `interaction_matrix=trues(n_systems, n_systems)`: Matrix controlling ordered system-pair
```
LasNikas: Please also update the function signature above

src/general/semidiscretization.jl
```diff
@@ -62,20 +85,61 @@ struct Semidiscretization{BACKEND, S, RU, RV, NS, UCU, IT}
...
+
+    return Matrix{eltype(interaction_matrix)}(interaction_matrix)
+end
+
+@inline is_interaction_entry(entry::Bool) = true
+
+function is_interaction_entry(entry)
+    return !isempty(methods(entry))
+end
+
```
LasNikas: Is there a reason why you put this code between the struct definition and the outer constructor?

LasNikas (changes requested):

src/general/semidiscretization.jl
```diff
@@ -1,37 +1,37 @@
...
-                            the same arguments as `interact!(...; kwargs...)` to use a custom
-                            interaction function for this force computation. Disabled
-                            pairs are skipped in the forces computation and in auxiliary neighbor loops such as
-                            density summation, correction factors, surface normals, pressure
-                            extrapolation, and particle shifting. The semidiscretization still
-                            stores a full matrix of neighborhood searches for uniform indexing
-                            and for APIs such as point interpolation, which use neighborhood
-                            searches independently of force interactions.
+- `neighborhood_search=GridNeighborhoodSearch{NDIMS}()`: The neighborhood search used in
+  the simulation. Use `nothing` to loop over all particles without a neighborhood search.
```
efaulhaber: Please indent by two spaces for readability.

src/general/semidiscretization.jl
```diff
@@ -253,18 +253,19 @@ Create an `ODEProblem` from the semidiscretization with the specified `tspan`.
...
+- `reset_threads=true`: Reset Polyester.jl threads before the simulation. After an error
+  within a threaded loop, threading might be disabled; resetting the threads ensures that
+  threading is enabled again for the simulation.
+  See also
+  [trixi-framework/Trixi.jl#1583](https://github.com/trixi-framework/Trixi.jl/issues/1583).
+- `restart_with=nothing`: Restart the simulation from VTK solution files created by
+  [`SolutionSavingCallback`](@ref). Pass a filename for a semidiscretization with one
+  system, or a tuple containing one filename for each system. The tuple order must match
+  the system order in the [`Semidiscretization`](@ref). By default, `nothing` starts a new
+  simulation. When restarting, `semidiscretize` replaces the initial time (`tspan[1]`) with
```
efaulhaber: Why? Niklas specifically removed the distinction for one or multiple systems, so that a tuple is the documented default (easier to understand than special behavior for one system), but passing a filename directly is still supported for convenience.

Also, please indent.

src/general/semidiscretization.jl
```diff
@@ -411,16 +412,22 @@ function initialize!(semi::Semidiscretization, restart_with::Nothing)
...
 
 # Arguments
-- `semi`:   The semidiscretization
-- `sol`:    The `ODESolution` returned by `solve` of OrdinaryDiffEq.jl
+- `semi`: The semidiscretization to update.
+- `sol`:  The `ODESolution` returned by `solve` from OrdinaryDiffEq.jl.
+
+# Keywords
+- `reset_threads=true`: Reset Polyester.jl threads before updating the systems. After an
+  error within a threaded loop, threading might be disabled; resetting the threads ensures
```
efaulhaber: Please indent.

efaulhaber (changes requested):

efaulhaber (dismissed):

efaulhaber (approved):

PR #1058, opened 2026-01-22
Only small variable name renamings remaining

svchb:
It originally explored an alternative to neighborhood-search handlers. That alternative is no longer part of this changeset; the branch now retains the neighborhood-search-handler implementation from `main`.

Based on #1057.


svchb:
Checked it on 15.8.2026 and is current.

src/general/semidiscretization.jl
```diff
@@ -631,7 +631,7 @@ function update_systems_and_nhs(v_ode, u_ode, semi, t)
         update_quantities!(system, v, u, v_ode, u_ode, semi, t)
     end
 
-    update_implicit_sph!(semi, v_ode, u_ode, t)
+    update_inter_system_quantities!(semi, v_ode, u_ode, t)
```
efaulhaber, line 634: This function is explicitly only used by IISPH. There is also no way to dispatch it since it's working on `semi` and not a `system`. I don't see the point in changing the name to something generic if the function is not generic and cannot be used generically.

efaulhaber (changes requested):

src/general/semidiscretization.jl
svchb: I know but if we wanted to support WCSPH+IISPH for example for water-air simulations or something a dispatch would need to be added and pressure "exchanged" at this point. Anyway something like this was the idea behind the name. I just don't like to have non generic calls start popping up in semi.

efaulhaber: > I just don't like to have non generic calls start popping up in semi.

But it is a non-generic call. Renaming it doesn't change that. If it becomes generic at some point, then it does make sense to rename it, but not now. Now it is just confusing.

PR #1059, opened 2026-01-22
Extract source-term handling from Semidiscretization

svchb:
## Summary

Extract source-term types and dispatch from `semidiscretization.jl` into `source_terms.jl`, keeping the semidiscretization implementation focused on system coupling.

## Stack

This PR is stacked on #1058 and should be reviewed after it.


svchb:
no longer makes sense

PR #1060, opened 2026-01-22
Extract ODE right-hand side from Semidiscretization

svchb:
## Summary

Move the ODE right-hand-side implementation from `semidiscretization.jl` into `ode_rhs.jl`. No behavioral changes are intended.


svchb:
checked on the 15.8.2026 and is current.

efaulhaber (approved):

PR #1061, opened 2026-01-22
Ode reoder

svchb:
based on #1060


svchb:
superseeded by 1060 and ongoing changes

PR #1101, opened 2026-03-16
Small fixes to callbacks

efaulhaber:
This PR
- fixes wall velocity in the interpolation if `include_wall_velocity == true` and `cut_off_bnd == false`,
- fixes the conditions for callbacks,
- makes density reinit callback work when a WCSPH system with `SummationDensity` is in the simulation.


efaulhaber:
I removed the interpolation fix, which is now included in #1185 and the steady state callback fix, which is now included in #1180.

svchb (approved):

PR #1112, opened 2026-03-19
Move TLSPH and Rigid fluid-structure interaction to structure

svchb:
based on #1110


src/schemes/structure/structure.jl
```diff
@@ -0,0 +1,66 @@
...
+    dv_adhesion = adhesion_force(surface_tension, neighbor_system, particle_system,
+                                 neighbor, particle, pos_diff, distance)
+
+    # Keep the shared helper acceleration-like; callers convert it to the quantity they store.
+    dv_fs = dv_boundary + dv_viscosity_ + dv_adhesion
+
+    return dv_fs
+end
+
+@inline function structure_fluid_continuity!(dv, v_particle_system, v_neighbor_system,
```
efaulhaber: I don't understand why you need a different function here. Why not dispatch `continuity_equation!`?

efaulhaber (changes requested):
I think the structure RHS is more controversial, so it would be much faster if you remove that from this PR and we do it in a new one. Unless you agree with my suggestion and it actually works like that in the code.

efaulhaber (changes requested):

efaulhaber (changes requested):

efaulhaber (approved):

PR #1114, opened 2026-03-20
[Proof of Concept] Solution and performance comparison with DualSPHysics

efaulhaber:
Based on #1044


efaulhaber:
Superseded by #1265.

PR #1120, opened 2026-03-26
Test 1.13

svchb:


svchb:
1.13RC3:
dependency issue
unit tests fail with some huge numerical errors in some cases seems like a regression in 1.13

PR #1123, opened 2026-04-01
Extend shared rigid contact runtime model

svchb:
Keeps PR1 focused on active rigid-contact runtime behavior: rigid contact stays normal-only, while contact_count and max_contact_penetration are added as diagnostics in system data, VTK output, and regression tests. The PR also fixes their lifecycle so they are recomputed per interaction pass, remain thread-safe, and stay available for output callbacks.

These diagnostics are useful now for monitoring contact in tests and ParaView, and later for richer contact features in follow-up PRs.


docs/src/systems/rigid_body.md
```diff
@@ -17,6 +17,29 @@ Rigid contact is configured through the contact model. This is separate from the
...
 
+`RigidContactModel` currently defines a normal spring-dashpot contact law with the
+parameters `normal_stiffness`, `normal_damping`, and `contact_distance`.
+
+The current implementation uses the same model for rigid-wall and rigid-rigid contact:
+
+- rigid-wall contact groups penetrating wall neighbors into a small number of contact
+  manifolds per rigid particle and applies one normal contact force per manifold,
+- rigid-rigid contact evaluates direct pairwise normal contact forces between rigid
+  particles,
```
efaulhaber, line 28: No manifold stuff here?

docs/src/systems/rigid_body.md
```diff
@@ -17,6 +17,29 @@ Rigid contact is configured through the contact model. This is separate from the
...
+- and both paths are currently normal-only, i.e. there are no tangential/frictional
+  forces or contact-history terms yet.
+
+`contact_distance` defines when contact starts. If `contact_distance == 0`, the
+particle spacing of the `RigidBodySystem` is used when the contact model is adapted to
+the runtime system.
```
efaulhaber:
```suggestion
particle spacing of the `RigidBodySystem` is used.
```
I think this second part is confusing and irrelevant for the user. Is the default even useful here? Isn't it enough to list that in the docstring?

src/schemes/structure/rigid_body/contact_models.jl
```diff
@@ -5,10 +5,11 @@ abstract type AbstractRigidContactModel end
...
 
-Basic rigid contact model stored on a rigid body.
-It is currently used for both rigid-wall and rigid-rigid contact.
-The contact force consists of a linear normal spring-dashpot contribution only.
-If `contact_distance == 0`, the particle spacing of the `RigidBodySystem` will be used as contact distance.
+Shared rigid-contact model used by the active rigid-wall and rigid-rigid contact paths.
+The current contact force consists of a linear normal spring-dashpot contribution only.
+If `contact_distance == 0`, the particle spacing of the `RigidBodySystem` will be used
+as contact distance when the model is adapted via
+`copy_contact_model(model, particle_spacing, ELTYPE)`.
```
efaulhaber:
```suggestion
as contact distance.
```
Irrelevant implementation details.

src/general/semidiscretization.jl
```diff
@@ -491,7 +491,7 @@ end
 
 # Update the systems and neighborhood searches (NHS) for a simulation
 # before calling `interact!` to compute forces.
-function update_systems_and_nhs(v_ode, u_ode, semi, t)
+function update_systems_and_nhs(v_ode, u_ode, semi, t; reset_interaction_caches=true)
```
efaulhaber: I really dislike having an extra kwarg here that is specific not just to rigid bodies but is only used for rigid body postprocessing. Please find a less invasive way to do this postprocessing.

efaulhaber (changes requested):

efaulhaber (approved):

PR #1124, opened 2026-04-01
Make `dv_shifting`, `surface_tension_force` and `adhesion_force` inplace

efaulhaber:
This PR makes `dv_shifting`, `surface_tension_force` and `adhesion_force` inplace and is part of #1131. Instead of accumulating zero `SVector`s when no shifting/surface tension/adhesion is used, these functions are now no-ops. It also reduces the number of live variables, which reduces register pressure.

Note that these optimizations make almost no difference on modern Hopper GPUs, but are very noticable on older Ampere GPUs.

### DIFF
--- a/src/io/write_vtk.jl
+++ b/src/io/write_vtk.jl
@@ -350,18 +350,13 @@ function write2vtk!(vtk, v, u, t, system::AbstractFluidSystem)
             rho_b = current_density(v, system, neighbor)
             grad_kernel = smoothing_kernel_grad(system, pos_diff, distance, particle)
 
-            surface_tension[1:ndims(system),
-                            particle] .+= surface_tension_force(surface_tension_a,
-                                                                surface_tension_b,
-                                                                system,
-                                                                system,
-                                                                particle,
-                                                                neighbor,
-                                                                pos_diff,
-                                                                distance,
-                                                                rho_a,
-                                                                rho_b,
-                                                                grad_kernel)
+            dv_surface_tension = Ref(zero(pos_diff))
+            surface_tension_force!(dv_surface_tension,
+                                   surface_tension_a, surface_tension_b,
+                                   system, system, particle, neighbor,
+                                   pos_diff, distance, rho_a, rho_b, grad_kernel)
+
+            surface_tension[1:ndims(system), particle] .+= dv_surface_tension[]
         end
         vtk["surface_tension"] = surface_tension
 
--- a/src/io/write_vtk.jl
+++ b/src/io/write_vtk.jl
@@ -354,7 +354,7 @@ function write2vtk!(vtk, v, u, t, system::AbstractFluidSystem)
             surface_tension_force!(dv_surface_tension,
                                    surface_tension_a, surface_tension_b,
                                    system, system, particle, neighbor,
-                                   pos_diff, distance, rho_a, rho_b, grad_kernel)
+                                   pos_diff, distance, rho_a, rho_b, grad_kernel, 1)
 
             surface_tension[1:ndims(system), particle] .+= dv_surface_tension[]
         end
--- a/src/schemes/fluid/entropically_damped_sph/rhs.jl
+++ b/src/schemes/fluid/entropically_damped_sph/rhs.jl
@@ -52,26 +52,27 @@ function interact!(dv, v_particle_system, u_particle_system,
                                                sound_speed, m_a, m_b, rho_a, rho_b,
                                                grad_kernel)
 
+        dv_particle = Ref(dv_pressure + dv_viscosity_)
+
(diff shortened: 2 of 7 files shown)


src/schemes/fluid/surface_tension.jl
```diff
@@ -162,66 +162,83 @@ end
...
+                                        particle, neighbor, pos_diff, distance,
+                                        rho_a, rho_b, grad_kernel,
+                                        surface_tension_correction)
     # No cohesion with oneself. See `src/general/smoothing_kernels.jl` for more details.
-    distance^2 < eps(initial_smoothing_length(particle_system)^2) && return zero(pos_diff)
+    distance^2 < eps(initial_smoothing_length(particle_system)^2) && return dv_particle
 
     m_b = hydrodynamic_mass(neighbor_system, neighbor)
     support_radius = compact_support(smoothing_kernel,
                                      smoothing_length(particle_system, particle))
```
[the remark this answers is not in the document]
efaulhaber: Moved to #1132.

src/io/write_vtk.jl
```diff
@@ -350,18 +350,13 @@ function write2vtk!(vtk, v, u, t, system::AbstractFluidSystem)
...
-                                                                rho_a,
-                                                                rho_b,
-                                                                grad_kernel)
+            dv_surface_tension = Ref(zero(pos_diff))
+            surface_tension_force!(dv_surface_tension,
+                                   surface_tension_a, surface_tension_b,
+                                   system, system, particle, neighbor,
+                                   pos_diff, distance, rho_a, rho_b, grad_kernel, 1)
+
+            surface_tension[1:ndims(system), particle] .+= dv_surface_tension[]
```
svchb, line 359: a comment is needed why this is so weird

src/schemes/fluid/entropically_damped_sph/rhs.jl
```diff
@@ -52,26 +52,27 @@ function interact!(dv, v_particle_system, u_particle_system,
...
-                                       particle_system, neighbor_system,
-                                       v_particle_system, v_neighbor_system,
-                                       particle, neighbor, m_a, m_b, rho_a, rho_b,
-                                       pos_diff, distance, grad_kernel, correction)
-
-        dv_surface_tension = surface_tension_force(surface_tension_a, surface_tension_b,
-                                                   particle_system, neighbor_system,
-                                                   particle, neighbor, pos_diff, distance,
-                                                   rho_a, rho_b, grad_kernel)
+        dv_particle = Ref(dv_pressure + dv_viscosity_)
```
svchb, line 55: comment needed to explain what is done here and why

src/schemes/fluid/weakly_compressible_sph/rhs.jl
```diff
@@ -70,27 +70,28 @@ function interact!(dv, v_particle_system, u_particle_system,
...
-                                       v_particle_system, v_neighbor_system,
-                                       particle, neighbor, m_a, m_b, rho_a, rho_b,
-                                       pos_diff, distance, grad_kernel, correction)
-
-        dv_surface_tension = surface_tension_correction *
-                             surface_tension_force(surface_tension_a, surface_tension_b,
-                                                   particle_system, neighbor_system,
-                                                   particle, neighbor, pos_diff, distance,
-                                                   rho_a, rho_b, grad_kernel)
+        dv_particle = Ref(dv_pressure + dv_viscosity_)
```
svchb, line 73: comment needed to explain what is done here and why

svchb (changes requested):

src/io/write_vtk.jl
efaulhaber: This is temporary and will be removed again in #1116. This PR is just a preparation that allows us to accumulate `dv_particle` over all neighbors in #1116. Same for your other comments.

Should I add a comment stating that or is it enough like this if we remove it anyway within the next 2 weeks?

svchb: ok than its fine

svchb (approved):

src/schemes/fluid/entropically_damped_sph/rhs.jl
LasNikas, line 55: Why can't you do `dv_particle = Ref(zero(pos_diff))` and pass this to both `dv_pressure` and `dv_viscosity`, as you did with the other `dv_*`s?

efaulhaber: Because `dv_viscosity` is not inplace yet (this is #1125). In #1116, you can see how this will be used at the end, with one `dv_particle = Ref(zero(pos_diff))` before the neighbors loop and accumulation over all neigbors. This here is ugly but temporary.

PR #1128, opened 2026-04-01
Use fast divisions in performance-critical code

efaulhaber:
This PR switches all divisions inside the neighbor loop to fast divisions. The fluid-* interaction kernel now still contains regular divisions in the PointNeighbors.jl part, but not for each neighbor particle, so these don't make a noticeable difference.

As shown in the benchmarks in #1116, these fast divisions make a massive difference in the runtime of the fluid-* interaction kernel, but they don't work with `Float64`. @Mikolaj-A-Kowalski wrote this incredible `fast_inv_cuda` function to get fast divisions with `Float64` without losing much accuracy in https://github.com/CliMA/Oceananigans.jl/pull/5140.

Here is a simple demonstration of the accuracy of this function vs a simple `llvm.nvvm.rcp.approx.ftz.d` without the following iteration:
```julia
julia> y2 = CUDA.rand(Float64, 100_000) .+ 0.1; # avoid numbers close to zero

julia> maximum(inv.(y2) .- fast_inv_cuda_nofma.(y2))
7.105216770497691e-6

julia> maximum(inv.(y2) .- fast_inv_cuda.(y2))
8.881784197001252e-16
```

Here is a comparison of different definitions for `div_fast` with the generated LLVM and corresponding runtime of the fluid-* interact kernel on an H100.
| Variant                        | LLVM Operation (FP32)                                 | Runtime (FP32) | LLVM Operation (FP64)                                 | Runtime (FP64) |
|---------------------------------|------------------------------------------------------|----------------|------------------------------------------------------|----------------|
| `x / y`                        | `fdiv float %143, %144, !dbg !528`                   | 4.894 ms       | `fdiv double %143, %144, !dbg !528`                  | 10.159 ms      |
| `x * (1 / y)`                  | `fdiv float 1.000000e+00, %145, !dbg !530`<br>`fmul float %144, %146, !dbg !533` | 4.227 ms       | `fdiv double 1.000000e+00, %145, !dbg !530`<br>`fmul double %144, %146, !dbg !533` | 6.878 ms       |
| `x * inv(y)`                   | `call float @llvm.nvvm.rcp.rn.f(float %118), !dbg !413`<br>`fmul float %117, %119, !dbg !418` | 3.620 ms       | `fdiv double 1.000000e+00, %145, !dbg !532`<br>`fmul double %144, %146, !dbg !535` | 6.852 ms       |
| `x * fast_inv_cuda(y)`         | `call float @llvm.nvvm.rcp.approx.ftz.f(float %118), !dbg !413`<br>`fmul float %117, %119, !dbg !418` | 3.281 ms       | `call double @llvm.nvvm.rcp.approx.ftz.d(double %118), !dbg !413`<br>`fneg double %118, !dbg !418`<br>`call double @llvm.fma.f64(double %119, double %120, double 1.000000e+00), !dbg !420`<br>`call double @llvm.fma.f64(double %121, double %121, double %121), !dbg !422`<br>`call double @llvm.fma.f64(double %122, double %119, double %119), !dbg !424`<br>`fmul double %117, %123, !dbg !426` | 4.844 ms       |
| `Base.FastMath.div_fast(x, y)` | `call float @llvm.nvvm.div.approx.f(float %118, float %115), !dbg !413` | 3.114 ms       | `fdiv fast double %143, %144, !dbg !530`             | 10.114 ms      |
| `x * fast_inv_cuda_nofma(y)`   | `call double @llvm.nvvm.rcp.approx.ftz.d(double %118), !dbg !413`<br>`fmul double %117, %119, !dbg !418` | —              | —                                                    | 4.758 ms       |

With `Float32`, `1 / y` is faster than `x / y`, but it doesn't produce a reciprocal instruction.
`inv` does produce a reciprocal and is faster. The approx reciprocal is even faster, but still not as fast as the fast division.
With `Float64`, `x / y` is translated to a `fdiv fast`, but (as expected, since there is no double fast division in PTX), this is not faster. `1 / y` and `inv` are both translated to `fdiv double 1.0 ...` and are significantly faster than the regular division. `fast_inv_cuda` is even faster, and skipping the Newton-iteration makes it only 1.018x faster (at a significant loss of accuracy).

### DIFF
--- a/docs/src/development.md
+++ b/docs/src/development.md
@@ -66,3 +66,39 @@ To create a new release for TrixiParticles.jl, perform the following steps:
    version should be `v0.3.1-dev`. If you just released `v0.2.4`, the new development
    version should be `v0.2.5-dev`.
 
+## Writing GPU-compatible code (@id writing_gpu_code)
+
+When implementing new functionality that should run on both CPUs and GPUs,
+follow these rules:
+
+1. Data structures must be generic and parametric.
+   Do not hardcode concrete CPU array types like `Vector` or `Matrix` in fields.
+   Use type parameters, so the same structure can store CPU arrays and GPU arrays.
+2. Add an Adapt.jl rule in `src/general/gpu.jl`.
+   Register the new type with `Adapt.@adapt_structure ...`, so `adapt` can recursively
+   convert all arrays inside the structure to GPU arrays.
+   This conversion is then applied automatically inside `semidiscretize`.
+3. Use `@threaded` for all loops.
+   Accessing GPU arrays inside regular loops is not allowed.
+   With a GPU backend, `@threaded` loops are compiled to GPU kernels.
+4. Write type-stable code and do not allocate inside `@threaded` loops.
+   This is required for GPU kernels and is also essential for fast multithreaded CPU code.
+
+## Writing fast GPU code
+
+The following rules improve kernel performance and avoid common GPU pitfalls:
+
+1. Avoid exceptions and bounds errors inside kernels.
+   Perform all required checks before entering `@threaded` loops (that is, before GPU kernels).
+   Then use `@inbounds` directly at the loop where bounds are guaranteed.
+   In TrixiParticles.jl, we do not place `@inbounds` inside inner helper functions.
+   Instead, mark helper functions with `@propagate_inbounds` so the loop-level
+   `@inbounds` is propagated.
+2. Avoid implicit `Float64` literals in arithmetic.
+   For example, prefer `x / 2` over `0.5 * x` so `Float32` simulations stay in `Float32`.
+   Verify this with `@device_code`, or by confirming the kernel runs on an Apple GPU
+   (most Apple GPUs do not support `Float64`).
+3. Use `div_fast` only in performance-critical divisions and only after benchmarking.
+   It can significantly speed up kernels, but should not be applied indiscriminately.
+   When introducing `div_fast` in code, add a reference to [this section](@ref gpu_div_fast)
+   to document the rationale and benchmarking context.
--- a/docs/src/development.md
+++ b/docs/src/development.md
@@ -98,7 +98,13 @@ The following rules improve kernel performance and avoid common GPU pitfalls:
    For example, prefer `x / 2` over `0.5 * x` so `Float32` simulations stay in `Float32`.
    Verify this with `@device_code`, or by confirming the kernel runs on an Apple GPU
    (most Apple GPUs do not support `Float64`).
-3. Use `div_fast` only in performance-critical divisions and only after benchmarking.
+3. Use `div_fast` in performance-critical divisions, but only after benchmarking (!).
    It can significantly speed up kernels, but should not be applied indiscriminately.
    When introducing `div_fast` in code, add a reference to [this section](@ref writing_fast_gpu_code)
-   to document the rationale and benchmarking context.
+   to document the rationale and benchmarking context, e.g., like so:
+   ```julia
+   # Since this is one of the most performance critical functions, using fast divisions
+   # here gives a significant speedup on GPUs.
+   # See the docs page "Development" for more details on `div_fast`.
+   result = div_fast(divident, divisor)
+   ```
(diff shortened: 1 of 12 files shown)


efaulhaber:
To-Do:
- [x] Tests for `fast_inv_cuda`.
- [x] Figure out what happens on AMD and Metal.

sloede:
Very impressive results! I recommend to add a description to the performance section of the docs that briefly explains the rationale behind this optimization and references this PR for further details. It should also include a note that `fast_div` should only be used where it was deemed necessary and effective after making adequate performance comparisons.

In addition, I recommend to add a comment to each use of `fast_div` that points to said documentation, such that unwitting new developers do not think this is something to use at will.

docs/src/development.md
```diff
@@ -66,3 +66,39 @@ To create a new release for TrixiParticles.jl, perform the following steps:
...
+   Instead, mark helper functions with `@propagate_inbounds` so the loop-level
+   `@inbounds` is propagated.
+2. Avoid implicit `Float64` literals in arithmetic.
+   For example, prefer `x / 2` over `0.5 * x` so `Float32` simulations stay in `Float32`.
+   Verify this with `@device_code`, or by confirming the kernel runs on an Apple GPU
+   (most Apple GPUs do not support `Float64`).
+3. Use `div_fast` only in performance-critical divisions and only after benchmarking.
+   It can significantly speed up kernels, but should not be applied indiscriminately.
+   When introducing `div_fast` in code, add a reference to [this section](@ref writing_fast_gpu_code)
+   to document the rationale and benchmarking context.
```
sloede:
~~~suggestion
3. Use `div_fast` in performance-critical divisions, but only after benchmarking (!).
   It can significantly speed up kernels, but should not be applied indiscriminately.
   When introducing `div_fast` in code, add a reference to [this section](@ref writing_fast_gpu_code)
   to document the rationale and benchmarking context, e.g., like so:
   ```julia
   # Since this is one of the most performance critical functions, using fast divisions
   # here gives a significant speedup on GPUs.
   # See the docs page "Development" for more details on `div_fast`.
   result = div_fast(divident, divisor)
   ```
~~~
I slightly changed the first sentence to be more explanatory and added an example of the comment for consistency.

sloede (commented):
Very nice docs! Small suggestion from my side.

ext/TrixiParticlesCUDAExt.jl
```diff
@@ -0,0 +1,34 @@
+# TODO this might be integrated into CUDA.jl at some point, see
+# https://github.com/JuliaGPU/CUDA.jl/pull/3077
```
LasNikas, line 2: Do we need an issue to be reminded once https://github.com/JuliaGPU/CUDA.jl/pull/3077 is integrated?

svchb:
> To-Do:
> 
> * [x]  Tests for `fast_inv_cuda`.
> * [ ]  Figure out what happens on AMD and Metal.

and what is with AMD and metal?

ext/TrixiParticlesCUDAExt.jl
svchb: If this is **soonish** added to CUDA.jl does it even make sense to merge this?

efaulhaber: "Soonish" is an overstatement, I think. At this point, it is unclear if this will be added at all. But @vchuravy can probably give some more information if and when this will be added.

efaulhaber:
> and what is with AMD and metal?

It doesn't make sense that I wrote Metal. Metal doesn't support FP64, so this is irrelevant there. I just benchmarked with Niklas earlier today on an MI300A, and it turned out that FP64 fast divisions just work out of the box with `Base.FastMath.div_fast`, so this hacky workaround is not required there. I updated the benchmark results in #1131.

ext/TrixiParticlesCUDAExt.jl
efaulhaber: #1137

LasNikas (approved):

svchb (approved):

PR #1130, opened 2026-04-02
Make `continuity_equation!` and `density_diffusion!` work on Ref values

efaulhaber:
This PR makes `continuity_equation!` and `density_diffusion!` work on Ref values instead of directly writing to `dv`. Together with #1116, it makes it possible to accumulate `drho_particle` over all neighbors and only once write it to `dv`.
It is part of #1131.

### DIFF
--- a/src/schemes/fluid/weakly_compressible_sph/density_diffusion.jl
+++ b/src/schemes/fluid/weakly_compressible_sph/density_diffusion.jl
@@ -225,10 +225,8 @@ end
                                 particle_system, particle, neighbor)
     density_diffusion_term = dot(psi, grad_kernel) * volume_b
 
-    smoothing_length_avg = (smoothing_length(particle_system, particle) +
-                            smoothing_length(particle_system, neighbor)) / 2
-    drho_particle[] += delta * smoothing_length_avg * sound_speed *
-                       density_diffusion_term
+    h = smoothing_length(particle_system, particle)
+    drho_particle[] += delta * h * sound_speed * density_diffusion_term
 end
 
 # Density diffusion `nothing` or interaction other than fluid-fluid
--- a/src/schemes/fluid/weakly_compressible_sph/density_diffusion.jl
+++ b/src/schemes/fluid/weakly_compressible_sph/density_diffusion.jl
@@ -206,13 +206,12 @@ function update!(density_diffusion::DensityDiffusionAntuono, v, u, system, semi)
     return density_diffusion
 end
 
-@propagate_inbounds function density_diffusion!(dv,
+@propagate_inbounds function density_diffusion!(drho_particle,
                                                 density_diffusion::AbstractDensityDiffusion,
-                                                v_particle_system, particle, neighbor,
-                                                pos_diff, distance, m_b, rho_a, rho_b,
                                                 particle_system::Union{AbstractFluidSystem,
                                                                        OpenBoundarySystem{<:BoundaryModelDynamicalPressureZhang}},
-                                                grad_kernel)
+                                                particle, neighbor, pos_diff, distance,
+                                                m_b, rho_a, rho_b, grad_kernel)
     # Density diffusion terms are all zero for distance zero.
     # See `src/general/smoothing_kernels.jl` for more details.
     distance^2 < eps(initial_smoothing_length(particle_system)^2) && return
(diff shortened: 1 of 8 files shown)


src/schemes/fluid/weakly_compressible_sph/density_diffusion.jl
```diff
@@ -226,15 +225,13 @@ end
                                 particle_system, particle, neighbor)
     density_diffusion_term = dot(psi, grad_kernel) * volume_b
 
-    smoothing_length_avg = (smoothing_length(particle_system, particle) +
-                            smoothing_length(particle_system, neighbor)) / 2
-    dv[end, particle] += delta * smoothing_length_avg * sound_speed *
-                         density_diffusion_term
+    h = smoothing_length(particle_system, particle)
+    drho_particle[] += delta * h * sound_speed * density_diffusion_term
```
efaulhaber: I did not find any information in any paper saying that this should be the average smoothing length, so I simplified it. @svchb, @LasNikas do you know if there was a reason why we implemented it this way?

svchb: Its the average smoothing length for refinement/variable smoothing length. It was introduced when smoothing_length was made a per particle value in preparation of the variable smoothing length/refinement implementation.

src/schemes/fluid/weakly_compressible_sph/density_diffusion.jl
```diff
@@ -226,15 +225,13 @@ end
                                 particle_system, particle, neighbor)
     density_diffusion_term = dot(psi, grad_kernel) * volume_b
 
-    smoothing_length_avg = (smoothing_length(particle_system, particle) +
-                            smoothing_length(particle_system, neighbor)) / 2
-    dv[end, particle] += delta * smoothing_length_avg * sound_speed *
-                         density_diffusion_term
+    h = smoothing_length(particle_system, particle)
```
svchb: return to average definition since this was already changed to a per particle value in preparation for refinement/variable smoothing length

svchb (changes requested):

src/schemes/fluid/weakly_compressible_sph/density_diffusion.jl
efaulhaber: Where does it say that? I couldn't find a source for that.

svchb: I think it always is

efaulhaber: Since delta is a kinda arbitrary parameter, I guess it doesn't matter here, so we can remove the extra load and average computation here, no?

svchb:
I found **one blocker** in the current head.

## Blocker: open-boundary continuity path lost its dispatch target

In `src/schemes/boundary/open_boundary/rhs.jl`, the PR now calls:

```julia
continuity_equation!(drho_particle, ContinuityDensity(),
                     particle_system, neighbor_system, ...)
```

where `particle_system` is an `OpenBoundarySystem{<:BoundaryModelDynamicalPressureZhang}`. 

But after the refactor, the relevant overload in `src/schemes/fluid/fluid.jl` is only defined for

```julia
particle_system::AbstractFluidSystem
```

and `OpenBoundarySystem` is **not** an `AbstractFluidSystem`; it is an `AbstractSystem{NDIMS}`.  

Before this PR, there was still a generic helper

```julia
continuity_equation!(dv, particle_system, neighbor_system, ...)
```

with no `AbstractFluidSystem` restriction, and the open-boundary code reused that path. This PR removes that generic path by folding the logic into the typed overload, so this looks like a regression that should surface as a `MethodError` once the open-boundary interaction is exercised. 

## Why I think this is the intended fix direction

The helpers used inside the continuity update already have open-boundary support: `OpenBoundarySystem` defines `density_diffusion` and `shifting_technique`, and `density_diffusion!` already accepts `OpenBoundarySystem{<:BoundaryModelDynamicalPressureZhang}`. So the missing piece is the `continuity_equation!` dispatch, not the underlying math.   

## Suggested fix

Either:

1. **Restore the generic helper** that computes the continuity contribution for any system supporting the required accessors, then let both fluid and open-boundary code call it, or
2. Add a dedicated overload for `OpenBoundarySystem{<:BoundaryModelDynamicalPressureZhang}` mirroring the previous generic behavior.

## What looks fine now

The earlier Copilot concern about `write_drho_particle!` for structure systems seems already addressed: `structure.jl` now has a generic no-op `write_drho_particle!(dv, ::AbstractSystem, ...)`, so that specific `MethodError` no longer looks live. 

So my review would be: **request changes for the open-boundary continuity dispatch regression**.

svchb (approved):

LasNikas (approved):

PR #1139, opened 2026-04-13
Apply GPU optimizations to TLSPH

efaulhaber:
Based on https://github.com/trixi-framework/PointNeighbors.jl/pull/154. Tests will pass once PointNeighbors 0.6.6 is released.


src/schemes/structure/total_lagrangian_sph/rhs.jl
```diff
@@ -29,48 +31,61 @@ end
...
+                                                       initial_distance, particle)
+
+            rho_b = @inbounds system.material_density[neighbor]
+            m_b = @inbounds system.mass[neighbor]
+            # PK1 / rho^2
+            pk1_rho2_b = @inbounds pk1_rho2(system, neighbor)
+            current_coords_b = @inbounds current_coords(system, neighbor)
+
+            # The compiler is smart enough to optimize this away if no penalty force is used
+            F_b = @inbounds deformation_gradient(system, neighbor)
```
LasNikas, line 71: Does this have to be placed here in order to be optimized by the compiler?
```suggestion
            # F_b is used here only for the penalty force. The compiler is smart enough to optimize this away if no penalty force is used.
            F_b = @inbounds deformation_gradient(system, neighbor)
```
Does this have to be placed here in order to be optimized by the compiler?

src/schemes/structure/total_lagrangian_sph/system.jl
```diff
@@ -489,30 +489,48 @@ end
...
+            # Now that we know that `distance` is not zero, we can safely call the unsafe
+            # version of the kernel gradient to avoid redundant zero checks.
+            grad_kernel = smoothing_kernel_grad_unsafe(system, initial_pos_diff,
+                                                       initial_distance, particle)
+
+            volume = @inbounds mass[neighbor] / material_density[neighbor]
+            current_coords_b = @inbounds current_coords(system, neighbor)
+            pos_diff_ = current_coords_a - current_coords_b
+            # On GPUs, convert `Float64` coordinates to `Float32` after computing the difference
+            pos_diff = convert.(eltype(system), pos_diff_)
```
LasNikas, line 526: Would it perhaps make sense at this point to create a function that developers who aren't very familiar with GPU implementation can use? Something like `calculate_pos_diff(pos_a, pos_b, system)`. Or is that just unnecessary maintenance work?

src/schemes/structure/total_lagrangian_sph/rhs.jl
efaulhaber: Unfortunately, yes. Moving this into the penalty force makes it slower when penalty force is used. Having it here does not make it slower when no penalty force is used.

src/schemes/structure/total_lagrangian_sph/system.jl
efaulhaber: I think it's more readable like this because it clearly states what is happening here instead of replacing it by a cryptic function.

src/schemes/structure/total_lagrangian_sph/penalty_force.jl
```diff
@@ -14,38 +14,51 @@ struct PenaltyForceGanzenmueller{ELTYPE}
...
 
-    inv_current_distance = 1 / current_distance
+    E_a = young_modulus(system, particle)
+    E_b = young_modulus(system, neighbor)
 
-    # Use the symmetry of epsilon to simplify computations
-    eps_sum = (F_a + F_b) * initial_pos_diff - 2 * current_pos_diff
-    delta_sum = dot(eps_sum, current_pos_diff) * inv_current_distance
+    eps_a = F_a * initial_pos_diff - current_pos_diff
+    eps_b = -(F_b * initial_pos_diff - current_pos_diff)
```
svchb: This is not the same as the line that is replaced? Why now with a minus for eps_b?

svchb (changes requested):

src/schemes/structure/total_lagrangian_sph/penalty_force.jl
efaulhaber: You're right. The new version is the correct epsilon from the paper, but then the delta is missing a minus.
I'm surprised that the tests are not failing because of this, only the validation. Apparently, this incorrect penalty force and be countered by a small time step. I added another oscillating beam example test with more penalty force and a tightly tuned CFL as a regression test, which correctly catches this error.

svchb (approved):

LasNikas (approved):

PR #1143, opened 2026-04-14
Make pressure clipping in `BoundaryModelDummyParticles` optional

efaulhaber:
Pressure clipping is not a correct formulation, it is just a workaround to avoid sticking artifacts.
Away from free surfaces it leads to artificially increased boundary pressures that cause larger gaps between fluid and boundary in areas of low pressure, against which the particle shifting technique is fighting.

This PR adds an option to disable clipping, which should be enabled for closed systems.


svchb:
What is supposed to happen when clipping in the state equation is activated and deactivated here?

efaulhaber:
When clipping in the state equation is activated, then pressure is clipped everywhere and it cannot become negative anyway, except perhaps with Adami when the boundary is accelerated away from the fluid.

svchb:
But shouldn't this be an error than or warning?

efaulhaber:
> But shouldn't this be an error than or warning?

Why? Everything is working as expected.

svchb:
Because the inconsistency might be confusing.

efaulhaber:
I don't see an inconsistency. Enable pressure clipping for the boundary model to clip boundary pressure. Enable pressure clipping of the state equation to enable global pressure clipping.

svchb:
Ok than just adjust the docstring also for state equation pressure clipping.

src/schemes/boundary/wall_boundary/dummy_particles.jl
```diff
@@ -22,6 +23,15 @@ Boundary model for [`WallBoundarySystem`](@ref).
 - `correction`:                 Correction method of the adjacent fluid system (see [Corrections](@ref corrections)).
 - `viscosity`:                  Slip (default) or no-slip condition. See description below for further
                                 information.
+- `clip_negative_pressure=true`: Clip negative boundary pressures to avoid sticking
```
LasNikas: As discussed on Slack, I would set the default to `false` if it isn't explicitly mentioned anywhere in the literature.

LasNikas (changes requested):

src/schemes/boundary/wall_boundary/dummy_particles.jl
efaulhaber: Done.

src/schemes/boundary/wall_boundary/dummy_particles.jl
```diff
@@ -22,6 +23,15 @@ Boundary model for [`WallBoundarySystem`](@ref).
...
                                 information.
+- `clip_negative_pressure=false`: Clip negative boundary pressures to avoid sticking
+                                artifacts from attractive fluid-boundary forces at free
+                                surfaces. Note that this is not a correct formulation
+                                at interior boundaries (away from free surfaces).
+                                Disable this option when simulating closed systems without
+                                free surfaces to avoid artificially increased boundary
+                                pressures that cause larger gaps between fluid and boundary
+                                in areas of low pressure, against which the particle
+                                shifting technique is fighting.
```
svchb, line 34:
```suggestion
- `clip_negative_pressure=false`: Clip negative boundary pressures to avoid sticking
                                artifacts from attractive fluid-boundary forces at free
                                surfaces. Note that this is not a correct formulation
                                at interior boundaries (away from free surfaces).

                                This option only controls clipping performed inside
                                `BoundaryModelDummyParticles` itself. If the selected
                                `state_equation` already clips negative pressures,
                                then boundary pressures computed directly from that
                                state equation can still be non-negative regardless
                                of this setting.

                                Disable this option when simulating closed systems without
                                free surfaces to avoid artificially increased boundary
                                pressures that cause larger gaps between fluid and boundary
                                in areas of low pressure, against which the particle
                                shifting technique is fighting.
```

svchb (changes requested):
- Add a similar note to state equation
- add a test for the mixed stateequation/boundary clipping

LasNikas (approved):

svchb (approved):

PR #1147, opened 2026-04-17
Use combined vector loads on GPUs

efaulhaber:
Based on #1116.


src/general/semidiscretization.jl
```diff
@@ -94,12 +94,34 @@ function Semidiscretization(systems::Union{AbstractSystem, Nothing}...;
...
-                     for i in eachindex(sizes_v))
+
+    ELTYPE = eltype(first(systems))
+    if parallelization_backend isa KernelAbstractions.GPU
+        # Align ranges to 64 bytes by adding padding if necessary.
+        # This ensures that aligned loads can be used on the wrapped integration arrays,
+        # which can significantly improve performance on GPUs.
+        block_size = div(64, sizeof(ELTYPE))
+    else
+        # There is no performance benefit to aligning ranges for CPU backends.
```
sloede: Is this guaranteed? IIRC, "back in the days" we had our own "fast allocation memory pool" implemented in C++ and we did see some performance improvements on the CPU when going to aligned load/store.

sloede: Orthogonal to the question above: Why not also do the alignment on the CPU? Unless there is a significant benefit from not doing this on the CPU, I'd recommend to always use the same alignment everywhere, every time. This makes it much easier to reason about differences between CPU and GPU code, and ensures that you do not accidentally screw this up on the GPU (but miss it, because your development happens on the CPU).

efaulhaber: I benchmarked it and didn't find a difference between regular loads and `vloada`. Also note that alignment is not trivial on the CPU, as the start of the array may not be aligned to pages like on the GPU, so we would have to use an initial offset as well. That also means we can't just use aligned loads on matrices like the deformation gradient of elastic structures.

efaulhaber: Adding padding here means we add dummy values to the time integration that have to be computed as well. If there is no benefit on the CPU, why would I increase the workload?

sloede: Thanks for the explanation. Out of curiosity: Have you also tested with an initial offset such that on the GPU everything is properly aligned?

sloede: Given that you add at most 15 additional FP32 values to the time integrator per system, I'd say that's negligible.

> If there is no benefit on the CPU, why would I increase the workload?

Maybe I was not clear above: From experience, I'd prefer a consistent memory layout among nearly all backends (CPU, GPU, whatever) over a potential (but very likely negligible) increase in the overall workload. I've spend too much time on OOB errors/Valgrind runs/Heisenbug chases, thus making the code simpler and harder to use wrongly is very high on my priority list 😅

Adding at most 15 FP32 values (for 64 byte alignment), which corresponds to - at most - 5 additional particles for the time integrator, I'd say that's a pretty good deal. But I'll leave the decision up to you.

efaulhaber: Yes. In my benchmarks, I only have one system, so everything is aligned. I added an offset of 8 or 16 bytes to this array, and there was no difference in the benchmark on the CPU.

efaulhaber: @LasNikas @svchb what are your thoughts?

sloede: 👍

efaulhaber:
@sloede I now ran a benchmark with the TLSPH kernel, which reads two 2x2 matrices per particle-neighbor pair that can be aligned, so we expect a larger difference in performance here. As opposed to the integration array with the padding, these are individual arrays that are always aligned on the GPU, but not guaranteed on the CPU (although in practice always aligned for large enough arrays). Note that for the non-aligned version I just added an offset of 8 bytes at the beginning.

| Method             | aligned memory (ARM) | non-aligned memory (ARM) | aligned memory (RAMSES) | non-aligned memory (RAMSES) |
|--------------------|:-------------:|:------------------:|:--------------------:|:------------------------:|
| linear indexing    | 16.467 ms     | 16.480 ms          | 15.325 ms            | 15.244 ms                |
| Cartesian indexing | 15.631 ms     | 15.623 ms          | 15.543 ms            | 15.528 ms                |
| vload              | 14.845 ms     | 14.871 ms          | 14.274 ms            | 14.263 ms                |
| vloada             | 14.837 ms     | —                  | 14.272 ms            | —                        |

- "linear indexing" is
  ```julia
  SMatrix{N, N}(ntuple(@inline(j->@inbounds A[(i - 1) * N^2 + j]), Val(N^2)))
  ```
- "Cartesian indexing" is
  ```julia
  SMatrix{N, N}(ntuple(@inline(j->@inbounds A[mod(j - 1, N) + 1, div(j - 1, N) + 1, i]), Val(N^2)))
  ```
  This is slower on the GPU (adds two integer add instructions) and in micro benchmarks on the ARM CPU, but for some reason, it is faster than linear indexing in the full interactions benchmark on the ARM CPU. I don't understand why and the Julia Slack wasn't helpful yet. On RAMSES, it is slower as expected.
- "vload" is using `SIMD.vload` for a vector load. This is noticably faster here on the CPU, but on the GPU it produces the same instructions as the linear indexing.
- "vloada" is using `SIMD.vloada` for **aligned** vector loads (and hence requires alignment). On the GPU, this is the only version that produces combined load instructions that are faster. This has the same performance as `vload` on the CPU.

Since `vload` is just as fast as `vloada` and alignment doesn't make a difference in performance, I conclude that we don't need to worry about alignment on the CPU and should use `vload` instead. The only question is if we want to use the same padding on CPUs for consistency reasons only.

svchb:
> @sloede I now ran a benchmark with the TLSPH kernel, which reads two 2x2 matrices per particle-neighbor pair that can be aligned, so we expect a larger difference in performance here. As opposed to the integration array with the padding, these are individual arrays that are always aligned on the GPU, but not guaranteed on the CPU (although in practice always aligned for large enough arrays). Note that for the non-aligned version I just added an offset of 8 bytes at the beginning.
> 
> | Method             | aligned memory (ARM) | non-aligned memory (ARM) | aligned memory (RAMSES) | non-aligned memory (RAMSES) |
> |--------------------|:-------------:|:------------------:|:--------------------:|:------------------------:|
> | linear indexing    | 16.467 ms     | 16.480 ms          | 15.325 ms            | 15.244 ms                |
> | Cartesian indexing | 15.631 ms     | 15.623 ms          | 15.543 ms            | 15.528 ms                |
> | vload              | 14.845 ms     | 14.871 ms          | 14.274 ms            | 14.263 ms                |
> | vloada             | 14.837 ms     | —                  | 14.272 ms            | —                        |
> 
> - "linear indexing" is
>   ```julia
>   SMatrix{N, N}(ntuple(@inline(j->@inbounds A[(i - 1) * N^2 + j]), Val(N^2)))
>   ```
> - "Cartesian indexing" is
>   ```julia
>   SMatrix{N, N}(ntuple(@inline(j->@inbounds A[mod(j - 1, N) + 1, div(j - 1, N) + 1, i]), Val(N^2)))
>   ```
>   This is slower on the GPU (adds two integer add instructions) and in micro benchmarks on the ARM CPU, but for some reason, it is faster than linear indexing in the full interactions benchmark on the ARM CPU. I don't understand why and the Julia Slack wasn't helpful yet. On RAMSES, it is slower as expected.
> - "vload" is using `SIMD.vload` for a vector load. This is noticably faster here on the CPU, but on the GPU it produces the same instructions as the linear indexing.
> - "vloada" is using `SIMD.vloada` for **aligned** vector loads (and hence requires alignment). On the GPU, this is the only version that produces combined load instructions that are faster. This has the same performance as `vload` on the CPU.
> 
> Since `vload` is just as fast as `vloada` and alignment doesn't make a difference in performance, I conclude that we don't need to worry about alignment on the CPU and should use `vload` instead. The only question is if we want to use the same padding on CPUs for consistency reasons only.

Have you tested this on CPU on a sufficiently large benchmark? CPUs do alot of hiding latency when you are not escaping the L1/L2 Cache size.

sloede:
> Since `vload` is just as fast as `vloada` and alignment doesn't make a difference in performance, I conclude that we don't need to worry about alignment on the CPU and should use `vload` instead. The only question is if we want to use the same padding on CPUs for consistency reasons only.

Yes, that's the question. Especially in performance-critical sections, if I can get by with just a single implementation for everything, I'd prefer that. You always need to (or should) keep in mind that you're not writing this code just for yourself, but also the next generation of researchers who might have much less experience in performance engineering and probably highly value simplicity in these "infrastructure" regions of the code. But as I said, it's a question of prioritizing different objectives 🤷‍♂️



> Have you tested this on CPU on a sufficiently large benchmark? CPUs do alot of hiding latency when you are not escaping the L1/L2 Cache size.

I agree, this would be interesting to know.

efaulhaber:
> Have you tested this on CPU on a sufficiently large benchmark? CPUs do alot of hiding latency when you are not escaping the L1/L2 Cache size.

The same 4M particles benchmark that I run on the GPUs.

svchb:
> > Have you tested this on CPU on a sufficiently large benchmark? CPUs do alot of hiding latency when you are not escaping the L1/L2 Cache size.
> 
> The same 4M particles benchmark that I run on the GPUs.

Hmm than modern CPUs are just so optimized that it doesn't matter.

efaulhaber:
The alignment check should fail on Metal because https://github.com/JuliaGPU/Metal.jl/pull/768 is not yet released. I guess it makes sense to wait for the next release of Metal.jl and then require that version in the compat entries.

PR #1150, opened 2026-04-22
Remove Publications section from README

LasNikas:
This section is empty anyway.


sloede:
No! I'd much rather you start adding the TrixiParticles.jl publications there! This is a great way to point others to the scientific output that has been generated with our package, and maybe encourages further use for research & development by others.

LasNikas:
But why should we have an empty section in the README? The Trixi.jl README doesn't have this section either. That's why I decided to remove it.

efaulhaber:
You already have the packing paper. Why don't you list it there?

PR #1171, opened 2026-04-30
support scimlbase 3

svchb:


efaulhaber:
This PR breaks compatibility with all old versions. Do we want that? Trixi.jl keeps backwards compatibility:
https://github.com/trixi-framework/Trixi.jl/pull/2996/changes

svchb:
> This PR breaks compatibility with all old versions. Do we want that? Trixi.jl keeps backwards compatibility: https://github.com/trixi-framework/Trixi.jl/pull/2996/changes

yes but they don't actually test the old stack. I tried with an older version before and than it just broke because of the dependencies that are too wide in the sciml world.

svchb:
So if we want to have compatability with the older sciml stack we would also need to add a test in my opinion for that.

efaulhaber:
That is a good point. I guess since we're not as big as Trixi, we can simply drop support for the old versions. @sloede what do you say is the best approach here?

sloede:
> That is a good point. I guess since we're not as big as Trixi, we can simply drop support for the old versions. @sloede what do you say is the best approach here?

I don't think backwards compatibility is strictly needed. If you find that for some reason SciMLBase v3 is not working out, you can always re-add backwards compatibility later.

src/callbacks/info.jl
```diff
@@ -67,8 +67,8 @@ function (info_callback::InfoCallback)(integrator)
                 @sprintf("│ run time: %.4e s", runtime_absolute))
     end
 
-    # Tell OrdinaryDiffEq that u has not been modified
-    u_modified!(integrator, false)
+    # This callback only reports progress and does not introduce a derivative discontinuity.
```
efaulhaber: I think this is not very self-explanatory. How about something like this?
```suggestion
    # This callback only reports progress and does not change the result of the right-hand side.
```
This comment is for all occurrences of `derivative_discontinuity!`.

efaulhaber (changes requested):

src/callbacks/split_integration.jl
```diff
@@ -207,17 +207,13 @@ function (split_integration_callback::SplitIntegrationCallback)(integrator)
     data.t_ref[] = new_t
 
     if isapprox(new_t, old_t)
-        # Tell OrdinaryDiffEq that `u` has NOT been modified.
         # Usually, `u` is modified in the split integration above, but if the split
         # integrator has already been advanced to the new step time in the last stage of the
         # previous step, the split integration above is skipped and `u` is not modified.
-        # (Technically, the split integration `u` is copied to the large `u` to account for
-        # potential caching errors, but the RHS of the last stage of the previous step
-        # can be reused for FSAL methods, which is what `u_modified!` is for.)
```
efaulhaber, line 216: This comment was essential. How about rephrasing it?
```julia
        # Usually, `u` is modified in the split integration above, but if the split
        # integrator has already been advanced to the new step time in the last stage of the
        # previous step, the split integration above is skipped and only the split integration `u`
        # is copied to the large `u` to account for potential caching errors.
        # This does not affect the RHS of the large integrator, so no discontinuity is introduced.
```

src/callbacks/split_integration.jl
```diff
@@ -571,8 +569,8 @@ function update_averaged_velocity_callback!(integrator)
         compute_averaged_velocity!(system, v_ode, semi, t_new)
     end
 
-    # Tell OrdinaryDiffEq that `integrator.u` has not been modified
-    u_modified!(integrator, false)
+    # No derivative discontinuity is introduced since `integrator.u` has not been modified and hence does not change the result of the right-hand side.
```
efaulhaber:
```suggestion
    # This callback does not modify `integrator.u` and hence does not change the result
    # of the right-hand side.
```

src/callbacks/stepsize.jl
```diff
@@ -72,8 +72,8 @@ function (stepsize_callback::StepsizeCallback)(integrator)
     integrator.opts.dtmax = dt
     integrator.dtcache = dt
 
-    # Tell OrdinaryDiffEq that `u` has not been modified
-    u_modified!(integrator, false)
+    # This callback only processes results and does not change the result of the right-hand side.
```
efaulhaber:
```suggestion
    # This callback only updates the step size and does not change the result of the right-hand side.
```

src/callbacks/update.jl
```diff
@@ -83,10 +83,9 @@ function (update_callback!::UpdateCallback)(integrator)
     semi = integrator.p.semi
     v_ode, u_ode = integrator.u.x
 
-    # Tell OrdinaryDiffEq that `integrator.u` has NOT been modified.
-    # This will be set to `true` in any of the update functions below that modify
-    # either `v_ode` or `u_ode`.
-    u_modified!(integrator, false)
+    # The update callback starts without a derivative discontinuity and hence don't modify
+    # the results of the right hand side. Individual update functions below mark one when they change `v_ode` or `u_ode`.
```
efaulhaber:
```suggestion
    # An empty update without calling any of the functions below does not modify
    # the results of the right-hand side.
    # The functions that add a discontinuity call `derivative_discontinuity!` themselves.
```

efaulhaber (changes requested):

JoshuaLampert:
> yes but they don't actually test the old stack.

We _do_ test the old stack via the [downgrade CI job](https://github.com/trixi-framework/Trixi.jl/blob/97ab50636e6c4751227f94be3f2dd57bae947b50/.github/workflows/Downgrade.yml) in Trixi.jl.

svchb:
ok thanks I didn't know.

I still would be for the less complicated option of only supporting scimlbase 3 moving forward.

efaulhaber (approved):

PR #1178, opened 2026-05-08
Fix `SortingCallback` and add tests

efaulhaber:
The `SortingCallback` is currently broken and untested.

### DIFF
--- a/src/callbacks/sorting.jl
+++ b/src/callbacks/sorting.jl
@@ -42,7 +42,7 @@ function SortingCallback(; interval::Integer=-1, dt=0.0, initial_sort=true)
         interval = Float64(dt)
 
         # Sort every time step (default)
-    elseif interval == -1
+    elseif interval <= 0
         interval = 1
     end
 
@@ -77,6 +77,7 @@ end
 function (sorting_callback!::SortingCallback{Int})(u, t, integrator)
     (; interval) = sorting_callback!
 
+    # Don't sort when the simulation is finished.
     return !isfinished(integrator) && condition_integrator_interval(integrator, interval)
 end
 
@@ -84,7 +85,9 @@ end
 function (sorting_callback!::SortingCallback)(u, t, integrator)
     (; interval, last_t) = sorting_callback!
 
-    return (t - last_t) > interval
+    # Sort in the next time step after `dt` has elapsed since the last sorting.
+    # Don't sort when the simulation is finished.
+    return !isfinished(integrator) && (t - last_t) > interval
 end
 
 # `affect!`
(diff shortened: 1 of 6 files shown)


src/callbacks/sorting.jl
svchb:
```suggestion
    return !isfinished(integrator) && (t - last_t) > interval
```

svchb, line 169: add initial_sort state to output

test/callbacks/sorting.jl
```diff
@@ -0,0 +1,137 @@
...
+        # Trigger after 3 accepted steps.
+        integrator = (; t=0.1, stats=(; naccept=3), isfinished=false)
+        @test callback.condition(nothing, 0.1, integrator) == true
+
+        # Don't trigger if the integrator is finished.
+        integrator = (; t=0.1, stats=(; naccept=3), isfinished=true)
+        @test callback.condition(nothing, 0.1, integrator) == false
+
+        callback = SortingCallback(dt=0.02)
+        @test callback.condition(nothing, 0.019, nothing) == false
```
svchb:
```suggestion
        integrator = (; t=0.021, isfinished=false)
        @test callback.condition(nothing, 0.019, integrator) == false
```

test/callbacks/sorting.jl
```diff
@@ -0,0 +1,137 @@
...
+        @test callback.condition(nothing, 0.021, nothing) == true
```
svchb:
```suggestion
         integrator = (; t=0.021, isfinished=true)
        @test callback.condition(nothing, 0.021, integrator) == true
```

src/callbacks/sorting.jl
```diff
@@ -31,8 +32,8 @@ See [#1044](https://github.com/trixi-framework/TrixiParticles.jl/pull/1044) for
                        slow down the first time steps, since a perfect grid is even better
                        than sorting by NHS cell index.
 """
-function SortingCallback(; interval::Integer=-1, dt=0.0)
-    if dt > 0 && interval !== -1
+function SortingCallback(; interval::Integer=-1, dt=0.0, initial_sort=true)
+    if dt > 0 && interval != -1
         throw(ArgumentError("Setting both interval and dt is not supported!"))
     end
```
svchb, line 39:
```suggestion
interval >= -1 || throw(ArgumentError("`interval` must be positive or -1"))
dt >= 0 || throw(ArgumentError("`dt` must be non-negative"))
interval != 0 || throw(ArgumentError("`interval` must be positive or -1"))
```

svchb (changes requested):

src/callbacks/sorting.jl
```diff
@@ -41,47 +42,57 @@ function SortingCallback(; interval::Integer=-1, dt=0.0)
         interval = Float64(dt)
 
         # Sort every time step (default)
-    elseif interval == -1
+    elseif interval <= 0
```
svchb, line 45: this is different to the behavior in the other callbacks

svchb (changes requested):

src/callbacks/sorting.jl
efaulhaber: I changed it to throw an error when calling `SortingCallback()`. For solution saving that makes sense if one wants to store initial and final solution only, but for sorting it doesn't.

svchb (dismissed):

LasNikas (approved):

PR #1180, opened 2026-05-11
Fix steady state callback interval handling

svchb:
## Summary
- respect `interval` before evaluating the steady-state kinetic-energy condition
- normalize integer tolerances to floating-point values so the callback can be constructed with integer inputs
- reject non-positive `interval_size` values

## Root Cause
The discrete callback condition called `steady_state_condition!` directly, so `interval` was never checked. The constructor also chose the previous kinetic-energy vector element type before converting tolerances to a floating type, which made integer tolerances try to store `Inf` as an integer.

## Validation
- `julia --project=. -e 'using Test, TrixiParticles; include("test/callbacks/steady_state_reached.jl")'`


src/callbacks/steady_state_reached.jl
```diff
@@ -26,13 +26,16 @@ end
...
                                     interval_size::Integer=10, abstol=1.0e-8, reltol=1.0e-6)
-    ELTYPE = eltype(abstol)
-    abstol, reltol = promote(abstol, reltol)
-
     if dt > 0 && interval > 0
         throw(ArgumentError("setting both `interval` and `dt` is not supported"))
     end
 
+    interval_size > 0 ||
+        throw(ArgumentError("`interval_size` must be positive"))
```
efaulhaber:
```suggestion
    if interval_size <= 0
        throw(ArgumentError("`interval_size` must be positive"))
    end
```
I think that's more readable.

src/callbacks/steady_state_reached.jl
```diff
@@ -67,10 +70,22 @@ function (cb::SteadyStateReachedCallback{Int})(integrator)
...
+end
+
 function (steady_state_callback::SteadyStateReachedCallback)(vu_ode, t, integrator)
     return steady_state_condition!(steady_state_callback, integrator)
 end
 
+@inline function condition_steady_state_interval(cb::SteadyStateReachedCallback{Int},
+                                                 integrator)
+    return cb.interval == 0 ||
+           condition_integrator_interval(integrator, cb.interval; save_final_solution=false)
```
efaulhaber: Why should this callback behave differently than all other callbacks? Because it would break existing behavior with `SteadyStateReachedCallback()` without arguments?

src/callbacks/steady_state_reached.jl
```diff
@@ -67,10 +70,22 @@ function (cb::SteadyStateReachedCallback{Int})(integrator)
 end
 
 # `condition` (`DiscreteCallback`)
+function (steady_state_callback::SteadyStateReachedCallback{Int})(vu_ode, t, integrator)
+    condition_steady_state_interval(steady_state_callback, integrator) || return false
```
efaulhaber:
```suggestion
    # TODO interval = 0 and interval = 1 should not be equivalent, but this would be a breaking change
    if steady_state_callback.interval == 0 || condition_steady_state_interval(steady_state_callback, integrator)
        return steady_state_condition!(...)
    end
    
    return false
```
Or change the default behavior if we don't consider it a breaking change since it is not documented anywhere that this CB should work with interval zero.

efaulhaber (changes requested):

src/callbacks/steady_state_reached.jl
```diff
@@ -8,11 +8,13 @@ where `ekin` is the total kinetic energy of the simulation.
 
 # Keywords
 - `interval=0`:     Check steady state condition every `interval` time steps.
+                    A value of `0` disables step-interval checks.
```
efaulhaber:
```suggestion
                    Use either `interval` or `dt`.
```

src/callbacks/steady_state_reached.jl
```diff
@@ -8,11 +8,13 @@ where `ekin` is the total kinetic energy of the simulation.
...
 - `dt=0.0`:         Check steady state condition in regular intervals of `dt` in terms
                     of integration time by adding additional `tstops`
                     (note that this may change the solution).
-- `interval_size`:  The interval in which the change of the kinetic energy is considered.
-                    `interval_size` is a (integer) multiple of `interval` or `dt`.
+- Either `interval` or `dt` must be set to something larger than 0.
```
efaulhaber:
```suggestion
                    Use either `interval` or `dt`.
```

src/callbacks/steady_state_reached.jl
```diff
@@ -26,51 +28,107 @@ end
...
+function initialize_steady_state_callback!(cb, u, t, integrator)
+    # The `SteadyStateReachedCallback` is either `cb.affect!` (with `DiscreteCallback`)
+    # or `cb.affect!.affect!` (with `PeriodicCallback`).
+    # Let recursive dispatch handle this.
+    initialize_steady_state_callback!(cb.affect!, u, t, integrator)
+end
+
+function initialize_steady_state_callback!(cb::SteadyStateReachedCallback, u, t, integrator)
+    semi = integrator.p.semi
+    set_callbacks_used!(semi, integrator)
```
efaulhaber: Why?

src/callbacks/steady_state_reached.jl
```diff
@@ -26,51 +28,107 @@ end
...
+
+    empty!(cb.previous_ekin)
+    push!(cb.previous_ekin, convert(eltype(cb.previous_ekin), Inf))
+
+    return nothing
```
efaulhaber:
```suggestion
    return cb
```

src/callbacks/steady_state_reached.jl
```diff
@@ -26,51 +28,107 @@ end
...
+end
+
 # `affect!` (`PeriodicCallback`)
 function (cb::SteadyStateReachedCallback)(integrator)
-    steady_state_condition!(cb, integrator) || return nothing
+    if !steady_state_condition!(cb, integrator)
+        u_modified!(integrator, false)
```
efaulhaber, line 88:
```suggestion
    u_modified!(integrator, false)
    
    if !steady_state_condition!(cb, integrator)
```
Do we even need `u_modified!` in the condition? Have you checked this?

src/callbacks/steady_state_reached.jl
```diff
@@ -26,51 +28,107 @@ end
...
+        return nothing
+    end
 
     print_summary(integrator)
 
     terminate!(integrator)
+
+    u_modified!(integrator, false)
+
```
efaulhaber:
```suggestion
```

src/callbacks/steady_state_reached.jl
```diff
@@ -26,51 +28,107 @@ end
...
+    return nothing
```
efaulhaber:
```suggestion
    return cb
```

src/callbacks/steady_state_reached.jl
```diff
@@ -26,51 +28,107 @@ end
...
 
 # `affect!` (`DiscreteCallback`)
 function (cb::SteadyStateReachedCallback{Int})(integrator)
     print_summary(integrator)
 
     terminate!(integrator)
+
+    u_modified!(integrator, false)
+
+    return nothing
```
efaulhaber:
```suggestion
    return cb
```

src/callbacks/steady_state_reached.jl
```diff
@@ -26,51 +28,107 @@ end
...
 end
 
 # `condition` (`DiscreteCallback`)
+function (steady_state_callback::SteadyStateReachedCallback{Int})(vu_ode, t, integrator)
+    condition_steady_state_interval(steady_state_callback, integrator) || return false
```
efaulhaber:
```suggestion
    if condition_integrator_interval(integrator, cb.interval; save_final_solution=false)
        return false
    end
```
No need for an extra function.

efaulhaber (changes requested):

src/callbacks/steady_state_reached.jl
svchb: ?

src/callbacks/steady_state_reached.jl
svchb: added comment

src/callbacks/steady_state_reached.jl
svchb: By default before affect! u_modified is set to true:

```
@inline function apply_discrete_callback!(integrator, callback::DiscreteCallback)
    saved_in_cb = false
    if callback.condition(integrator.u, integrator.t, integrator)
        # handle saveat
        _, savedexactly = savevalues!(integrator)
        saved_in_cb = true
        @inbounds if callback.save_positions[1]
            # if already saved then skip saving
            savedexactly || savevalues!(integrator, true)
        end
        integrator.u_modified = true
        callback.affect!(integrator)
```

efaulhaber: But this is the condition. If you don't enter the `if` branch, `u_modified` is not set. The code you posted confirms that `u_modified!` is only required in the `affect!`.

src/callbacks/steady_state_reached.jl
efaulhaber: You added another list entry.

src/callbacks/steady_state_reached.jl
efaulhaber: I do not understand your comment.

`set_callbacks_used!` only needs to be called when `semi.update_callback_used[]` or `semi.integrate_tlsph[]` are read within the callback initialization (e.g. when `kick!` is called). This is not the case here. This callback initialization is only emptying an array, so there is no need to call this function.

src/callbacks/steady_state_reached.jl
efaulhaber: Now I'm confused. This comment is for the `affect!`?

efaulhaber: Okay, nvm this comment. I got confused.

src/callbacks/steady_state_reached.jl
```diff
@@ -8,11 +8,13 @@ where `ekin` is the total kinetic energy of the simulation.
 
 # Keywords
 - `interval=0`:     Check steady state condition every `interval` time steps.
+                    Use either `interval` or `dt`.
 - `dt=0.0`:         Check steady state condition in regular intervals of `dt` in terms
                     of integration time by adding additional `tstops`
                     (note that this may change the solution).
-- `interval_size`:  The interval in which the change of the kinetic energy is considered.
-                    `interval_size` is a (integer) multiple of `interval` or `dt`.
+                    Either `interval` or `dt` must be set to something larger than 0.
```
efaulhaber:
```suggestion
                    Use either `interval` or `dt`.
```
Your sentence sounds more like "use at least one".

efaulhaber (changes requested):

efaulhaber (approved):

PR #1181, opened 2026-05-11
[codex] Fix density reinitialization callback

svchb:
## Summary
- store the particle system passed to `DensityReinitializationCallback` and reinitialize only that system
- trigger `dt`-based reinitialization when the requested interval is reached, not only after it is exceeded
- mark the integrator state as modified after density and pressure fields are changed
- add focused unit coverage for show output, `dt` triggering, and selected-system reinitialization

## Root Cause
The callback constructor accepted a specific particle system, but the affect path called the global `reinit_density!(vu_ode, semi)` helper, which loops over every system in the semidiscretization. The time-based condition also used a strict `>` comparison, delaying callbacks that land exactly on the configured interval.

## Validation
- `julia --project=. -e 'using Test, TrixiParticles; include("test/callbacks/density_reinit.jl")'`\n- `julia --project=. -e 'using Test, TrixiParticles, Printf; include("test/callbacks/callbacks.jl")'`


efaulhaber (changes requested):
Doesn't work because the system might be replaced in the semidiscretization.

src/callbacks/density_reinit.jl
```diff
@@ -1,16 +1,25 @@
 """
-    DensityReinitializationCallback(; interval::Integer=0, dt=0.0)
+    DensityReinitializationCallback(particle_system; interval::Integer=0, dt=0.0)
+    DensityReinitializationCallback(system_index::Integer; interval::Integer=0, dt=0.0)
+    DensityReinitializationCallback(system_indices; interval::Integer=0, dt=0.0)
+    DensityReinitializationCallback(; system_index::Integer, interval::Integer=0, dt=0.0)
+    DensityReinitializationCallback(; system_indices, interval::Integer=0, dt=0.0)
```
efaulhaber:
```suggestion
    DensityReinitializationCallback(system...; interval::Integer=0, dt=0.0)
```
This should be the only public API. Working with indices as a user is fragile and should always be avoided. See `MechanicalWorkCalculatorCallback`.

efaulhaber: Actually, I don't see the point in allowing multiple systems. This just adds additional complexity, while a user could just use two callbacks instead. So I think the API should stay the same.

examples/fluid/dam_break_2d.jl
```diff
@@ -111,8 +111,9 @@ extra_callback = nothing
 extra_callback2 = nothing
 
 use_reinit = false
+# Use the system stored in `ode.p.semi`, since `semidiscretize` can replace systems.
 density_reinit_cb = use_reinit ?
-                    DensityReinitializationCallback(semi.systems[1], interval=10) :
+                    DensityReinitializationCallback(ode.p.semi.systems[1]; interval=10) :
```
efaulhaber: Still too complex. This should be the API:
```suggestion
                    DensityReinitializationCallback(fluid_system, semi; interval=10) :
```
Like I said before, please implement it like the `MechanicalWorkCalculatorCallback`, which takes system and semi as arguments and then only stores the system index.

src/callbacks/density_reinit.jl
```diff
@@ -1,45 +1,50 @@
...
-# Keywords
-- `interval=0`:              Reinitialize the density every `interval` time steps.
-- `dt`:                      Reinitialize the density in regular intervals of `dt` in terms
-                             of integration time. This callback does not add extra time
-                             steps / `tstops`; instead, reinitialization is triggered at
-                             the first solver step after each `dt` interval has elapsed.
-- `reinit_initial_solution`: Reinitialize the initial solution (default=true)
-"""
-mutable struct DensityReinitializationCallback{I}
+mutable struct DensityReinitializationCallbackAffect{I}
```
efaulhaber: Why did you change the name? We don't append `Affect` for the other callbacks either.

src/callbacks/density_reinit.jl
```diff
@@ -83,27 +89,68 @@ function initialize_reinit_cb!(cb::DensityReinitializationCallback, u, t, integr
...
+                               vu_ode, semi)
+    v_ode, u_ode = vu_ode.x
+
+    particle_system = current_reinit_system(reinit_callback.system_index, semi)
+    check_density_reinit_system(particle_system)
+
+    v = wrap_v(v_ode, particle_system, semi)
+    u = wrap_u(u_ode, particle_system, semi)
+
+    reinit_density!(particle_system, v, u, v_ode, u_ode, semi)
```
efaulhaber, line 138: In the WCSPH system,
```julia
v[end, :] ./= kernel_correction_coefficient
```
doesn't work when there are inactive particles (e.g. with open boundaries). Please change this to
```julia
    @threaded semi for particle in eachparticle(system)
        v[end, particle] /= kernel_correction_coefficient[particle]
    end
```
to make it work.

efaulhaber (changes requested):
The big question is: is this a breaking change that requires a new major release? Or was the density reinit callback unusably broken before, so that we can consider it a fix instead?

test/callbacks/density_reinit.jl
```diff
@@ -0,0 +1,232 @@
...
+        TrixiParticles.initialize_reinit_cb!(callback, vu_ode, 0.0, integrator)
+        @test isempty(density_reinit_calls)
+
+        callback = DensityReinitializationCallback(system, semi; interval=1,
+                                                   reinit_initial_solution=true).affect!
+        TrixiParticles.initialize_reinit_cb!(callback, vu_ode, 0.0, integrator)
+        @test density_reinit_calls == [:fluid]
+    end
+
+    @testset verbose=true "selected system" begin
```
efaulhaber:
```suggestion
    @testset verbose=true "only selected systems are affected" begin
```

test/callbacks/density_reinit.jl
```diff
@@ -0,0 +1,232 @@
...
+        callback = DensityReinitializationCallback(system1; interval=1).affect!
+
+        @test callback.system_index == 0
+
+        TrixiParticles.reinitialize_density!(callback, vu_ode, semi)
+
+        @test density_reinit_calls == [:fluid1, :fluid2]
+    end
+
+    @testset verbose=true "selected semidiscretized system" begin
```
efaulhaber, line 111: I don't understand the difference between this test and the newly added next test. Please choose more descriptive test names and add more comments where necessary.

efaulhaber (changes requested):
Please also create a new PR to remove the legacy constructor in the next breaking release.

efaulhaber (approved):

PR #1182, opened 2026-05-12
Fix InitialCondition restart velocity wrapping

svchb:


efaulhaber (commented):
This seems like an oddly specific test for one very specific typo. I don't think that makes sense. This fix is also included in #1101.

PR #1184, opened 2026-05-12
Fix custom quantity reductions

svchb:


test/general/custom_quantities.jl
```diff
@@ -116,4 +116,104 @@
...
+        end
+        function TrixiParticles.current_density(v, system::ActiveFluidQuantityMock)
+            return TrixiParticles.current_density(v, ContinuityDensity(), system)
+        end
+        function TrixiParticles.current_pressure(v, system::ActiveFluidQuantityMock)
+            return system.pressure
+        end
+        function TrixiParticles.write_u0!(u0, system::ActiveFluidQuantityMock)
+            u0 .= system.initial_condition.coordinates
+            return u0
```
efaulhaber, line 185: At this point, wouldn't it be easier to use a WCSPH system?

src/general/custom_quantities.jl
```diff
@@ -39,7 +38,7 @@ end
 Returns the total mass of all particles in a system.
```
efaulhaber, line 38:
```suggestion
Returns the total mass of all active particles in a system.
```
I guess we should add this to all the docstrings to be explicit.

efaulhaber (changes requested):
Please add a NEWS.md entry.

efaulhaber (approved):

PR #1189, opened 2026-05-12
Use interpolation radius for packing SDF constraints

svchb:


test/systems/packing_system.jl
```diff
@@ -54,6 +55,26 @@
         │ place_on_shell: ……………………………………… no                                                               │
         │ boundary: ……………………………………………………… no                                                               │
         └──────────────────────────────────────────────────────────────────────────────────────────────────┘"""
-        @test repr("text/plain", system) == show_box
+    end
+
+    @trixi_testset "signed distance interpolation radius" begin
```
efaulhaber, line 60: Does this fail without the fix?

svchb: WIthout fix:
Test Failed at test/systems/packing_system.jl:78
Expression: u[1, 1] < initial_condition.coordinates[1, 1]
Evaluated: 0.0 < 0.0

efaulhaber (changes requested):
I don't understand the test without having to think about it extensively. Why is this testing the interpolation smoothing length? Why is u expected to be smaller than initially? Please add more comments explaining what is happening and why.

efaulhaber (approved):

PR #1195, opened 2026-05-17
Fix open boundary shifting

efaulhaber:
With #1043, the hardcoded continuity equation for dynamical pressure open boundaries was changed to use `continuity_equation!`, which contains shifting terms. These shifting terms in the continuity equation don't preserve volume when shifting is used at the free surface. This led to a gradual volume increase over time, which eventually destabilized my simulations.
This PR disables shifting entirely for all particles within one compact support of the free surface. The ramping then starts behind this region as usual.

- **Before:** Ramping within `[0, 1 * compact_support]`.
- **Now:** No shifting within `[0, 1 * compact_support]`. Ramping within `[1 * compact_support, 2 * compact_support]`.

### Numerical examples

I noticed this bug in my carbon fiber fin simulation (#844). At ~0.7s simulation time, the simulation is crashing. The pictures below are at 0.6s simulation time.

#### main

grafik
Notice the wrinkly texture. A bit later, this becomes so pronounced that it crashes the simulation. The reason for this is the volume-non-preservation that we can see in the density plot:
grafik
At this point, the total volume increased by 0.45% compared to the initial state.

#### Before #1043

grafik
Almost no wrinkles are visible. Only very slightly in the dark blue region at the bottom and in the top left.
grafik
Density is much closer to 1000, with a total volume increase of 0.1%.

#### This PR

grafik
The wrinkles are basically gone now. There are some lines visible in the top left, but that could just as well be visualization artifacts because ParaView uses 256 discrete colors by default.
grafik
Density is nicely centered around 1000, with a total volume increase of 0.005%.
Even after 5s simulation time, the total volume increases only by 0.027%.

### DIFF
--- a/src/schemes/boundary/open_boundary/system.jl
+++ b/src/schemes/boundary/open_boundary/system.jl
@@ -81,6 +81,16 @@ function OpenBoundarySystem(boundary_zones::Union{BoundaryZone, Nothing}...;
                                               density_diffusion(fluid_system) : nothing)
     boundary_zones_ = filter(bz -> !isnothing(bz), boundary_zones)
 
+    if boundary_model isa BoundaryModelDynamicalPressureZhang &&
+       !isnothing(shifting_technique)
+        # When dynamical pressure is used with shifting, the shifting is ramped up until
+        # it reaches the full value at two compact supports away from the free surface
+        # of the boundary zone.
+        # In order to prevent discontinuities in the shifting velocity, the boundary zone
+        # must be wide enough to accommodate this ramping region.
+        check_boundary_zone_widths(boundary_zones_, fluid_system)
+    end
+
     initial_conditions = union((bz.initial_condition for bz in boundary_zones_)...)
 
     buffer = SystemBuffer(nparticles(initial_conditions), buffer_size)
@@ -704,8 +714,7 @@ end
 
 @inline use_open_boundary_interpolation_neighbor(system) = false
 
-function check_configuration(system::OpenBoundarySystem, systems,
-                             neighborhood_search::PointNeighbors.AbstractNeighborhoodSearch)
+function check_configuration(system::OpenBoundarySystem, systems, neighborhood_search)
     (; boundary_model, boundary_zones) = system
 
     # Store index of the fluid system. This is necessary for re-linking
@@ -724,4 +733,20 @@ function check_configuration(system::OpenBoundarySystem, systems,
                             "that does not require an update for the first set of coordinates (e.g. `GridNeighborhoodSearch`). " *
                             "See the PointNeighbors.jl documentation for more details."))
     end
+end
+
+function check_boundary_zone_widths(boundary_zones, fluid_system)
+    support = compact_support(system_smoothing_kernel(fluid_system),
+                              initial_smoothing_length(fluid_system))
+    minimum_width = 2 * support
+
+    for (i, zone) in enumerate(boundary_zones)
+        if zone.zone_width < minimum_width
+            throw(ArgumentError("boundary zone $i has width $(zone.zone_width), " *
+                                "but must be at least two compact supports " *
+                                "($(minimum_width))"))
+        end
+    end
+
+    return nothing
 end
--- a/src/schemes/boundary/open_boundary/system.jl
+++ b/src/schemes/boundary/open_boundary/system.jl
@@ -605,14 +605,25 @@ end
                                  -boundary_zone.face_normal)
         dist_free_surface = boundary_zone.zone_width - dist_to_transition
 
-        if dist_free_surface < 2 * compact_support(fluid_system, fluid_system)
-            # Ramp shifting velocity near the free surface using a kernel-weighted transition.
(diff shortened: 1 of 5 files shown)


src/schemes/boundary/open_boundary/system.jl
```diff
@@ -605,14 +605,25 @@ end
                                  -boundary_zone.face_normal)
         dist_free_surface = boundary_zone.zone_width - dist_to_transition
 
-        if dist_free_surface < compact_support(fluid_system, fluid_system)
-            # Ramp shifting velocity near the free surface using a kernel-weighted transition.
+        free_surface_region_width = compact_support(fluid_system, fluid_system)
+        free_surface_ramp_width = 2 * free_surface_region_width
```
svchb: I think these names are not ideal. Isn't the ramp a width of 1*free_surface_region_width

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -605,14 +605,25 @@ end
...
+
+        if dist_free_surface <= free_surface_region_width
+            # The free-surface region is the layer within one compact support of the
+            # free surface. Shifting is disabled there to avoid shifting particles with
+            # incomplete support.
+            for dim in 1:ndims(system)
+                cache.delta_v[dim, particle] = zero(eltype(system))
+            end
+        elseif dist_free_surface < free_surface_ramp_width
```
svchb:
```suggestion
        elseif dist_free_surface < free_surface_region_width + free_surface_ramp_width
```

svchb (changes requested):

svchb (dismissed):

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -605,15 +605,30 @@ end
...
+            # The free-surface region is the layer within one compact support of the
+            # free surface. Shifting is disabled there to avoid shifting particles with
+            # incomplete support.
+            for dim in 1:ndims(system)
+                cache.delta_v[dim, particle] = zero(eltype(system))
+            end
+        elseif dist_free_surface < free_surface_region_width + ramp_width
+            # Between one and two compact supports from the free surface, shifting is
+            # ramped from zero to the unmodified shifting velocity with a kernel-weighted
+            # transition.
```
LasNikas, line 634: In that case, we need to add a check to verify that the boundary zone width is large enough (at least 2 * compact_support).

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -605,15 +605,30 @@ end
...
+        ramp_width = free_surface_region_width
+
+        if dist_free_surface <= free_surface_region_width
+            # The free-surface region is the layer within one compact support of the
+            # free surface. Shifting is disabled there to avoid shifting particles with
+            # incomplete support.
+            for dim in 1:ndims(system)
+                cache.delta_v[dim, particle] = zero(eltype(system))
+            end
+        elseif dist_free_surface < free_surface_region_width + ramp_width
```
LasNikas, line 631: Why do you need that extra check here? The kernel value is zero anyway outside of compact support. So the weighting is zero anyway. Or did you add the check just for clarity?

efaulhaber: The check is
1. for clarity,
2. for performance, to reduce the number of operations in case it's zero anyway,
3. for robustness, since we might want to use the unsafe version of the kernel in the future for performance reasons.

src/schemes/boundary/open_boundary/system.jl
```diff
@@ -605,15 +605,30 @@ end
                                  -boundary_zone.face_normal)
         dist_free_surface = boundary_zone.zone_width - dist_to_transition
 
-        if dist_free_surface < compact_support(fluid_system, fluid_system)
-            # Ramp shifting velocity near the free surface using a kernel-weighted transition.
+        # Width of the free-surface region where shifting is disabled.
+        free_surface_region_width = compact_support(fluid_system, fluid_system)
+        # Width of the ramp region behind the free-surface region where shifting is ramped
+        # from zero to the unmodified value.
+        ramp_width = free_surface_region_width
```
LasNikas, line 622:
```suggestion
       if dist_free_surface < 2 * compact_support(fluid_system, fluid_system)
            # Ramp shifting velocity near the free surface using a kernel-weighted transition.
```
Wouldn't that be enough?

src/schemes/boundary/open_boundary/system.jl
LasNikas: ah okay I see

src/schemes/boundary/open_boundary/system.jl
efaulhaber: Yes, this was my first version (see https://github.com/trixi-framework/TrixiParticles.jl/pull/1195/changes/18edf5af6cbcdc27ffeedb1a1556acae4dbec83a). I changed it to make it easier to understand.

svchb (approved):

LasNikas (approved):

PR #1203, opened 2026-05-17
Fix open-boundary characteristic zone handling

svchb:


test/schemes/boundary/open_boundary/characteristic_variables.jl
```diff
@@ -18,6 +18,72 @@
...
+                                             boundary_model=BoundaryModelCharacteristicsLastiwka())
+
+        error_str = "`BoundaryModelCharacteristicsLastiwka` needs a directed boundary zone. " *
+                    "Please specify `InFlow()` or `OutFlow()` instead of `BidirectionalFlow()`."
+        @test_throws ArgumentError(error_str) Semidiscretization(fluid_system,
+                                                                 boundary_system)
+    end
+
+    @testset "Fallback is zone-local" begin
+        face_vertices = ([0.0, 0.0], [0.0, 0.5])
```
efaulhaber, line 45: I don't think this test is trivial at all. It deserves several comments to make clear what is happening without studying the code.
```suggestion
        # This test creates two inflows far away from each other, the "near" one with fluid neighbors,
        # the "far" one without. When evaluating characteristics of the far inflow, no fluid neighbors
        # are found, so it should fall back to using previous values of other inflow particles.
        # This test checks that this fallback does not use previous values from the other inflow zone.
        face_vertices = ([0.0, 0.0], [0.0, 0.5])
```

test/schemes/boundary/open_boundary/characteristic_variables.jl
```diff
@@ -18,6 +18,72 @@
...
+        boundary_system = OpenBoundarySystem(inflow, inflow_far;
+                                             fluid_system, buffer_size=0,
+                                             boundary_model=BoundaryModelCharacteristicsLastiwka())
+        semi = Semidiscretization(fluid_system, boundary_system)
+        ode = semidiscretize(semi, (0.0, 5.0))
+
+        v0_ode, u0_ode = ode.u0.x
+        v = TrixiParticles.wrap_v(v0_ode, boundary_system, semi)
+        u = TrixiParticles.wrap_u(u0_ode, boundary_system, semi)
+
```
efaulhaber, line 74:
```suggestion

# Evaluate the characteristics to make sure that particles of the near inflow have non-zero
# characteristics. Then, evaluate again, and test that the far inflow particles don't use the
# non-zero characteristics of the particles in the other inflow zone.
```

test/schemes/boundary/open_boundary/characteristic_variables.jl
```diff
@@ -18,6 +18,72 @@
...
+        zone_1_particles = findall(==(1), boundary_system.boundary_zone_indices)
+        zone_2_particles = findall(==(2), boundary_system.boundary_zone_indices)
+
+        @test any(!isapprox(characteristic, 0.0)
+                  for characteristic in boundary_system.cache.characteristics[:,
+                                                                              zone_1_particles])
+        @test all(isapprox(characteristic, 0.0)
+                  for characteristic in boundary_system.cache.characteristics[:,
+                                                                              zone_2_particles])
```
efaulhaber:
```suggestion
        @test !all(iszero, boundary_system.cache.characteristics[:, zone_1_particles])
        @test all(iszero, boundary_system.cache.characteristics[:, zone_2_particles])
```
This is not numerically computed but lieterally set to zero, so we don't have to use `iszero`.

efaulhaber (changes requested):
Also add a NEWS.md entry.

efaulhaber (approved):

PR #1205, opened 2026-05-18
Fix Julia 1.10 CI fails due to hydrostatic tank validation

efaulhaber:
I could not figure out where the non-determinism comes from. This is the stupid fix now.

### DIFF
--- a/test/validation/validation.jl
+++ b/test/validation/validation.jl
@@ -104,8 +104,14 @@
         ]
 
         # We compare the relative error to the analytical solution
-        @test isapprox(errors[:edac][2], 0.0, atol=0.033)
-        @test isapprox(errors[:wcsph][2], 0.0, atol=0.045)
+        @test isapprox(errors[:edac][2], 0.0, atol=0.031)
+        if VERSION < v"1.11"
+            # Julia 1.10 CI tests produce larger errors non-deterministically
+            # for unknown reasons.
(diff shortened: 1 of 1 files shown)


efaulhaber:
@svchb I don't understand why this is so inconsistent. macOS is passing because it's ARM, but why is Julia 1.11 Windows failing?

svchb:
I kept the tspan and resolution as small as possible to reduce the runtime. This variation should reduce with the longer sampling time.
The simulation result is basically swinging around the analytical value slight shifts through adaptive time stepping modify the frequency of this swinging which have a larger impact on the smaller sampling time.

svchb:
With the longer sampling time:
Errors for hydrostatic water column 2D:Dict(:wcsph => (1.9001644312676932e-6, 0.026961378735545445), :edac => (1.531217523045389e-6, 0.021726401613458318))

Before:
Errors for hydrostatic water column 2D:Dict(:wcsph => (3.086363804366519e-6, 0.043792327693287766), :edac => (2.1778726591173615e-6, 0.0309017728329325))

so the error reduced by 20-40%.

efaulhaber:
My concern is the variation between CI runs, not the error itself.

svchb:
We kept the error bounds currently the same so the error is now well below the current bounds in the validation check.

efaulhaber:
We still have a problem with inconsistent and especially non-deterministic differences.

PR #1207, opened 2026-05-21
Validation Carreau Yasuda Model 

sherin-joseph:
Please note that this is more of a steady-profile preservation check than a full convergence-from-rest validation, because the default and high-resolution runs use analytical initial conditions. . For a stronger transient validation, we need to use initial_condition_mode = zero or newtonian and verify convergence of the velocity profile toward the analytical solution. 

The relative L2 velocity-profile error remains below 3% for all tested Carreau-Yasuda indices at ny = 200. The largest error occurs for n = 0.25, the strongest shear-thinning case, while the Newtonian limit n = 1 is recovered with about 1.4% relative L2 error.

carreau_final_profiles

The reference Poiseuille validation from Section 3.1 of Coclite et al. (2020) is shown below for comparison.

image


svchb:
For validation cases we want to have the following structure:
1) `example file` that is trixi_include'd into the `validation_run` file
2) the validation run file contain some kind of error bounds
3) Additionally a `plot file` to plot the results of the validation run.

So you just need to create this additional example file and figure out an error bound for the validation to get to that structure.

sherin-joseph:
@svchb 
The PR now includes:

examples/fluid/poiseuille_carreau_2d.jl as the example file
validation/poiseuille_carreau_2d/validation_poiseuille_carreau_2d.jl as the validation runner using trixi_include
relative L2 error bounds in the validation runner
plot_carreau_comparison.jl for plotting the validation results

I hope this follows the expected structure now. Please let me know if any further changes or corrections are needed.

Thanks for the guidance!

examples/fluid/poiseuille_carreau_2d.jl
```diff
@@ -0,0 +1,286 @@
+using TrixiParticles
+using OrdinaryDiffEqLowStorageRK
+
+# ==========================================================================================
+# 2D Periodic Poiseuille Flow with Carreau-Yasuda Viscosity
+#
```
efaulhaber, line 6: Please add the reference paper that this simulation is based on. Take a look at the dam break example file for reference.

examples/fluid/poiseuille_carreau_2d.jl
```diff
@@ -0,0 +1,286 @@
...
+# ==========================================================================================
+# ==== Resolution
+ny = 50
+
+# ==========================================================================================
+# ==== Experiment Setup
+t_end_factor = 0.1
+eps_factor = 1.0
+sound_speed_factor = 60.0
+initial_condition_mode = "analytical"
```
efaulhaber, line 21: Please use symbols here:

```suggestion
initial_condition_mode = :analytical
```

examples/fluid/poiseuille_carreau_2d.jl
```diff
@@ -0,0 +1,286 @@
...
+lambda_exponent = 2.0
+reynolds_number = 200.0
+
+reference_velocity = reynolds_number * nu0 / channel_height
+pressure_gradient = 8.0 * fluid_density * reference_velocity^2 /
+                    (reynolds_number * channel_height)
+acceleration_x = pressure_gradient / fluid_density
+carreau_time_constant = channel_height / max(reference_velocity, eps())
+
+t_end = t_end_factor * channel_height / max(reference_velocity, eps())
```
efaulhaber, line 41:
```suggestion
carreau_time_constant = channel_height / reference_velocity

t_end = t_end_factor * channel_height / reference_velocity
```
Or is there any reason why the `reference_velocity` should be zero?

examples/fluid/poiseuille_carreau_2d.jl
```diff
@@ -0,0 +1,286 @@
...
+tspan = (0.0, t_end)
+
+if !(initial_condition_mode in ("newtonian", "analytical", "zero"))
+    throw(ArgumentError("initial condition mode must be \"newtonian\", " *
+                        "\"analytical\", or \"zero\""))
+end
+
+# ==========================================================================================
+# ==== Analytical Solution
```
efaulhaber, line 50: A few thoughts on this whole block:
1. We usually try to keep the example files as simple as possible. So no postprocessing, no comparison to the analytical solution. This should all live in the validation file. Take a look at the dam break validation. There, the example is simple, it really only defines the setup without a postprocess callback. Then, in `validation/dam_break_2d/setup_marrone_2011.jl`, the sensors and the corresponding postprocess callback are defined. This example file here should only contain the definition of the initial condition.
2. You are currently evaluating the errors based on the particle values. In SPH, the actual solution approximation does not consist of the particle values at the particle positions. Instead, the particles are merely interpolation points, and the interpolated quantity is the approximate result produced by the simulation. That means you should use the SPH interpolation with `interpolate_line` and compare this against the analytical solution. This will greatly improve the quality of the surface layers and probably give you significantly smaller errors. See here how Niklas computed the interpolated solution for the Poiseuille flow.
3. It is usually preferred to keep the postprocessing as simple as possible, meaning you don't want to compute the whole error during the simulation. Instead, write out the interpolated velocity and then later compute the error against the analytical solution in the validation script.

validation/poiseuille_carreau_2d/plot_carreau_comparison.jl
```diff
@@ -0,0 +1,253 @@
...
+        m = match(r"timestep=\"([^\"]+)\".*file=\"([^\"]+)\"", line)
+        if m !== nothing
+            push!(times, parse(Float64, m.captures[1]))
+            push!(files, m.captures[2])
+        end
+    end
+    return times, files
+end
+
+function read_vtu_velocity(file::AbstractString)
```
efaulhaber, line 20: You should not need to read PVD and VTU files. Use the `PostprocessCallback` to extract all the quantities you need, then only read the JSON file here.

validation/poiseuille_carreau_2d/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+- relative L2 velocity error against the analytical profile
+- maximum absolute velocity error against the analytical profile
+
+## Running
+
+Default run:
+
+```bash
+julia --project=. validation/poiseuille_carreau_2d/validation_poiseuille_carreau_2d.jl
```
efaulhaber, line 57: We discourage the use of `--project=.` because that requires adding additional dependencies like plotting packages directly to TrixiParticles. See https://trixi-framework.org/TrixiParticles.jl/stable/install/#for-developers.

validation/poiseuille_carreau_2d/validation_poiseuille_carreau_2d.jl
```diff
@@ -0,0 +1,80 @@
...
+# power-law indices and checks the final relative L2 velocity error against the
+# analytical steady profile.
+# ==========================================================================================
+
+# ==========================================================================================
+# ==== Resolution
+# The default resolution is intentionally modest so this validation is practical to run
+# during review. For the high-resolution run inspired by Coclite et al. (2020), use
+# `ny = 200`.
+ny = length(ARGS) >= 1 ? parse(Int, ARGS[1]) : 50
```
efaulhaber, line 16: We don't usually use `ARGS`. Instead, just hardcode `ny = 50`, and then let users use `trixi_include(file, ny=100` to change the resolution. This allows users to work from an interactive REPL.

efaulhaber (changes requested):
I apologize for the long review turnaround. We have a lot of open PRs at the moment.

The results look very promising. Could you please also add the analytical solution in the plots to be able to get a better feeling of how accurate the results are?

sherin-joseph:
Thanks for the detailed review. I added the analytical profiles to the plots. These figures show the transient relaxation from the Newtonian initial profile toward the Carreau-Yasuda analytical profiles. The n = 0.25 case is much slower to converge, so I am rerunning the remaining cases for a longer final time and will update the figures once those runs finish.
image
image


> I apologize for the long review turnaround. We have a lot of open PRs at the moment.
> 
> The results look very promising. Could you please also add the analytical solution in the plots to be able to get a better feeling of how accurate the results are?

sherin-joseph:
image

PR #1208, opened 2026-05-22
Add Newtonian gravity force model

svchb:
## Summary

Adds a small gravity module centered on `NewtonianGravity` as a force model rather than a standalone astrophysics solver. The model provides pairwise Newtonian acceleration helpers, default no-op dispatch hooks, and validation for gravitational constant, softening length, and cutoff radius.

Also updates the local N-body example system so it can store a `NewtonianGravity` model while preserving the existing `G` field used by the benchmark path. A new `n_body_newtonian_gravity.jl` example exercises the explicit force-model construction, and the gravity docs page is wired into the manual.

## Performance

The benchmark file `examples/n_body/n_body_benchmark_trixi.jl` is intentionally unchanged.

`NBodySystem` now has one gravity interaction implementation. It asks the gravity model for the scalar acceleration factor, and the concrete `NewtonianGravity` type keeps optional softening/cutoff behavior type-stable without duplicating N-body methods. For the common unsoftened all-to-all case (`softening_length == 0`, `cutoff_radius == Inf`), that specializes to the original direct `G * m / r^3` formulation; softened/cutoff gravity still uses the general model helper.

Fair side-by-side timings compare an old-shape system using the original `TrixiParticles.interact!` formula against the current `NBodySystem` path. Both paths had zero allocations.

| particles | old median | new median | ratio new/old |
|---:|---:|---:|---:|
| 5 | `0.076551466s` | `0.081578193s` | `1.0657` |
| 9 | `0.227526529s` | `0.233741349s` | `1.0273` |
| 64 | `0.583772636s` | `0.577071592s` | `0.9885` |

The focused n-body system test guards the generic RHS path with a zero-allocation check, avoiding noisy benchmark assertions while still catching practical hot-path regressions.

## Testing

- `test/general/gravity.jl`
- `test/examples/n_body_system.jl`
- `examples/n_body/n_body_newtonian_gravity.jl`
- `examples/n_body/n_body_solar_system.jl`


efaulhaber (commented):
What is the point of this PR?

A pointwise gravity model can never be used in practice in a real simulation because looping over all particles is prohibitively expensive. Real N-body solvers use a tree-based approach, where particles don't interact with clusters instead of individual particles when they are far away.

examples/n_body/n_body_system.jl
```diff
@@ -1,24 +1,45 @@
 using TrixiParticles
 using LinearAlgebra
 
-struct NBodySystem{NDIMS, ELTYPE <: Real, IC} <: TrixiParticles.AbstractSystem{NDIMS}
+struct NBodySystem{NDIMS, ELTYPE <: Real, IC, GR} <: TrixiParticles.AbstractSystem{NDIMS}
     initial_condition :: IC
     mass              :: Array{ELTYPE, 1} # [particle]
-    G                 :: ELTYPE
-    buffer            :: Nothing
+    # Kept for compatibility with n-body benchmark code that reads `system.G`.
```
efaulhaber, line 7: This PR also looks like it still contains some AI artifacts and is not yet ready for review. Plus the tests are failing.

svchb:
Tests failed because of an issue with github by the way.

examples/n_body/n_body_system.jl
svchb: This is of course only a first step to enable some play examples. For larger systems one would have to develop a tree based method.

Follow up is #1209 and #1210. After #1210 systems with 1000s of particles can still be run for some play examples. Anything further would than require a tree based method.

PR #1215, opened 2026-05-26
Fix IISPH density error accounting

svchb:
## Summary
- Use absolute density errors when accumulating the IISPH pressure-solver convergence metric.
- Clear stale per-particle density errors when pressure is clamped to zero.
- Update the regression coverage for the corrected accounting.

## Root Cause
Signed density errors could cancel each other, and zero-pressure particles could retain stale error values from earlier iterations.

## Validation
- Direct Julia pressure-update regression check passed.
- Full targeted test files were blocked locally by `OrdinaryDiffEqCore` / `OrdinaryDiffEqSymplecticRK` precompilation failures.


efaulhaber:
This PR requires an in-depth analysis, clearly demonstrating the differences in behavior and runtime.
It should also be an option that can be turned on and off, so that we can still run simulations with the original IISPH behavior from Ihmsen et al. (2014).

The "bug" label is also incorrect. There was nothing broken, so I would consider this an "enhancement".

PR #1216, opened 2026-05-26
Fix IISPH neighbor coordinate wrapping

svchb:
## Summary
- Wrap neighbor-system coordinates from the ODE vector before calculating IISPH pressure contributions.
- Add a regression with two IISPH systems that have different particle counts.

## Root Cause
`calculate_sum_d_ij_pj!` used the current system wrapper for neighbor coordinates, which can index the wrong coordinate layout when systems differ.

## Validation
- Direct Julia multi-system IISPH regression check passed.
- Full targeted test files were blocked locally by `OrdinaryDiffEqCore` / `OrdinaryDiffEqSymplecticRK` precompilation failures.


test/systems/iisph_system.jl
```diff
@@ -459,6 +459,47 @@
             @test isapprox(system_pressure.density_error, [-3.0, 0.0])
         end
 
+        @testset "Cross-system pressure sums use neighbor coordinates" begin
```
efaulhaber, line 462: I think this test is too specific and unnecessary. It tests for this very specific single bug, which is now fixed anyway. In the future, we should have a simulation with two IISPH fluids, but for now, I would leave this PR as a two-liner.

efaulhaber (changes requested):

PR #1222, opened 2026-05-27
Implement NHS handlers to avoid storing redundant copies of the same grid NHS

efaulhaber:
Extracted from #1088.

This is the first part of improving NHS efficiency. It builds the handler structure and avoids storing multiple identical neighborhood searches without touching the NHS update process. This means instead of updating multiple redundant searches, we now update one search multiple times. The next PR avoids the redundant updates.


src/general/neighborhood_search.jl
```diff
@@ -220,13 +206,185 @@ function create_neighborhood_search(neighborhood_search, system::TotalLagrangian
...
+"""
+    SharedNHSHandler
+
+Neighborhood search handler optimized for reusable neighborhood searches.
+
+`SharedNHSHandler` stores one neighborhood search for each neighbor system and
+required search radius, instead of one search for every ordered pair of systems.
+For a query involving `(system, neighbor)`, it selects the stored search for the
+neighbor system with a radius large enough for that pair's compact support.
+
```
svchb, line 316:
```suggestion
`SharedNHSHandler` reuses the same neighborhood search for all interactions
that have the same neighbor system and search radius. This is only valid if the
neighborhood search stores information about the neighbor particles only.

For example, a grid-based search can be shared because it stores the neighbor
particles in grid cells; particles from any other system can then query that
grid. In contrast, a precomputed neighbor list cannot be shared because its
stored neighbors are tied to the specific query particles used during
initialization.
```

src/general/neighborhood_search.jl
```diff
@@ -220,13 +206,185 @@ function create_neighborhood_search(neighborhood_search, system::TotalLagrangian
...
+                            all(row -> isconcretetype(eltype(row)), searches)
+    @assert searches_are_concrete "neighborhood searches are not type-stable"
+
+    return SharedNHSHandler(search_radii, searches)
+end
+
+function get_neighborhood_search(handler::SharedNHSHandler, system_index, neighbor_index,
+                                 search_radius)
+    radii = handler.search_radii[neighbor_index]
+    radius_index = searchsortedfirst(radii, search_radius - eps(search_radius))
```
svchb, line 387:
```suggestion
  # `search_radius` is recomputed from the same compact-support values used to
  # build `radii`. Subtract one ulp to tolerate roundoff, while still throwing
  # below for genuinely too-large radii.
    radius_index = searchsortedfirst(radii, search_radius - eps(search_radius))
```

test/general/semidiscretization.jl
```diff
@@ -28,15 +28,51 @@
...
+                  semi_grid.neighborhood_search_handler.search_radii)
+        @test all(searches -> searches isa Vector,
+                  semi_grid.neighborhood_search_handler.neighborhood_searches)
+        @test all(searches -> isconcretetype(eltype(searches)),
+                  semi_grid.neighborhood_search_handler.neighborhood_searches)
+        @test TrixiParticles.get_neighborhood_search(semi_grid.neighborhood_search_handler,
+                                                     1, 1, 0.1) isa GridNeighborhoodSearch
+        @test_throws ArgumentError TrixiParticles.get_neighborhood_search(semi_grid.neighborhood_search_handler,
+                                                                          1, 1, 1.0)
+
```
svchb, line 77:
```suggestion
handler = semi_grid.neighborhood_search_handler
radii = handler.search_radii[1]
r = first(radii)

# Exact lookup.
@test TrixiParticles.get_neighborhood_search(handler, 1, 1, r) ===
      handler.neighborhood_searches[1][1]

# Tiny roundoff above a stored radius is tolerated.
@test TrixiParticles.get_neighborhood_search(handler, 1, 1, r + eps(r) / 2) ===
      handler.neighborhood_searches[1][1]

# But this is not a general "pick a too-small radius" mechanism.
@test_throws ArgumentError TrixiParticles.get_neighborhood_search(handler, 1, 1,
                                                                  last(radii) +
                                                                  10eps(last(radii)))
```

test/general/semidiscretization.jl
```diff
@@ -28,15 +28,51 @@
...
+                                        neighborhood_search_handler=PairsNHSHandler)
+        @test semi_pairs.neighborhood_search_handler isa TrixiParticles.PairsNHSHandler
+
+        # Need to pass a type, not an NHS handler object.
+        nhs = TrixiParticles.TrivialNeighborhoodSearch{3}()
+        handler = TrixiParticles.PairsNHSHandler(nhs, (system1, system2))
+        @test_throws ArgumentError Semidiscretization(system1, system2,
+                                                      neighborhood_search=nothing,
+                                                      neighborhood_search_handler=handler)
     end
```
svchb, line 91:
```suggestion
@testset verbose=true "SharedNHSHandler matches PairsNHSHandler neighbor queries" begin
    # Use a real two-system setup: this test guards the shared-storage invariant,
    # not just constructor/type plumbing.
    trixi_include(@__MODULE__,
                  joinpath(examples_dir(), "fluid", "hydrostatic_water_column_2d.jl"),
                  sol=nothing)

    systems = (fluid_system, boundary_system)

    # Reference: old pairwise storage model.
    semi_pairs = Semidiscretization(systems...;
                                    neighborhood_search=GridNeighborhoodSearch{2}(),
                                    neighborhood_search_handler=PairsNHSHandler,
                                    parallelization_backend=SerialBackend())

    # New shared storage model; observable neighbor queries must remain identical.
    semi_shared = Semidiscretization(systems...;
                                     neighborhood_search=GridNeighborhoodSearch{2}(),
                                     neighborhood_search_handler=SharedNHSHandler,
                                     parallelization_backend=SerialBackend())

    TrixiParticles.initialize_neighborhood_searches!(semi_pairs)
    TrixiParticles.initialize_neighborhood_searches!(semi_shared)

    function collect_neighbor_pairs(semi, system, neighbor)
        pairs = Tuple{Int, Int}[]

        system_coords = TrixiParticles.initial_coordinates(system)
        neighbor_coords = TrixiParticles.initial_coordinates(neighbor)

        PointNeighbors.foreach_point_neighbor(system, neighbor,
                                              system_coords, neighbor_coords,
                                              semi) do particle, neighbor_particle,
                                                       pos_diff, distance
            push!(pairs, (particle, neighbor_particle))
        end

        return sort!(pairs)
    end

    for system in systems, neighbor in systems
        pairs_pairs = collect_neighbor_pairs(semi_pairs, system, neighbor)
        pairs_shared = collect_neighbor_pairs(semi_shared, system, neighbor)

        @test pairs_shared == pairs_pairs
    end
end
```

test/examples/gpu.jl
```diff
@@ -315,6 +325,7 @@ end
 
             hydrostatic_water_column_tests = Dict(
                 "WCSPH default" => (),
+                "WCSPH with PairsNHSHandler" => (neighborhood_search_handler=PairsNHSHandler,),
```
svchb: SharedNHSHandler?

svchb (changes requested):
It should be checked if this should be adjusted after #1057 is merged semantically it might make sense to implement this as an handler as a followup? E.g. a CustomInteractionHandler or something. Not sure if this makes sense though?

NEWS.md entry is missing

test/examples/gpu.jl
efaulhaber: Should be the default with Grid NHS.

test/general/semidiscretization.jl
efaulhaber: Idk about this. That's another significant portion of code that we have to maintain, and I don't think it provides much value. There is only one search radius used in this simulation, so this is a rather trivial test.

This is already tested in the example tests, and we will later have harder tests with particle refinement.

test/examples/gpu.jl
efaulhaber: Correction: It is the default for this test.

svchb (dismissed):

svchb (dismissed):

svchb (approved):

PR #1221, opened 2026-05-27
Add vortex street example file

efaulhaber:
Ported from #781.

grafik

PR #781, opened 2025-04-25
Validate open boundaries

LasNikas:
Comparison of the Mirroring and Dynamical Pressure Models

# Tafuni et al. (2018)
 
| d / 100  (St = 0.206)|
|--------|
| image |

# `BoundaryModelMirroringTafuni`

### PST

```
shifting_technique = ParticleShiftingTechnique(; sound_speed_factor=0.1, v_max_factor=0)
```

| d / 12.5 | d / 25 | d / 50 
|------|-----|-----|
| image | image | image |

### TVF

```
shifting_technique = TransportVelocityAdami(background_pressure=5 * fluid_density * sound_speed^2)
```

| d / 12.5 | d / 25 | d / 50 
|------|-----|-----|
| image | image | image |

#  `BoundaryModelDynamicalPressureZhang`

### PST

```
shifting_technique = ParticleShiftingTechnique(; sound_speed_factor=0.1, v_max_factor=0)
```

| d / 12.5 | d / 25 | d / 50 
|------|-----|-----|
| image | image | image|

### TVF

```
shifting_technique = TransportVelocityAdami(background_pressure=5 * fluid_density * sound_speed^2)
```

| d / 12.5 | d / 25 | d / 50 
|------|-----|-----|
| image | image | image |

PR #1223, opened 2026-05-27
Vortex street validation

efaulhaber:
Second part of #781. Based on #1221.

This is the plot based on the reference files, which I created by running the validation file at the default resolution of 0.05:
grafik


efaulhaber:
@LasNikas @svchb I'm unsure how to validate this example from CI. The highest resolution with acceptable runtime would be 0.2. This runs in 1 minute on a large CPU, so we might be able to get it to run to t=2 in 5-10 minutes in CI, which is already longer than I'd prefer for a CI test. However, the result with this resolution looks like this:
grafik
```
┌ Info: Strouhal number
└   round(strouhal_number, digits = 3) = 0.029
┌ Info: Fraction of the dominant frequency band in the total spectrum
└   round(dominant_band_fraction, digits = 3) = 0.06
┌ Info: C_L_max for the periodic shedding
└   round(maximum(f_lift_cut), digits = 3) = 0.846

┌ Info: C_D_max for the periodic shedding
└   round(maximum(f_drag_cut), digits = 3) = 1.521
```
Of course, we can validate against these results to detect changes in the codebase, but I'd much prefer testing against a useful Strouhal number.

efaulhaber:
The simulation at this resolution looks very good though:
grafik

efaulhaber:
Update: I put a velocity probe in the wake of the cylinder, and that looks much nicer for the same low resolution:
grafik
I will use this for the validation.

PR #1224, opened 2026-05-27
IISPH pressure solver performance: cached pressure loop

svchb:
### Summary

This PR speeds up the IISPH pressure solver by caching the per-step pressure equation coefficients and reusing them across all Jacobi iterations. This also covers wall systems with `PressureBoundaries`, which participate in the pressure solve with their own pressure unknowns.

**Optimization: cached pressure loop**
- Per time step, the PPE coefficients (`d_ij`, `d_ji·∇W`, `m_j∇W`, boundary sums) are computed once and stored in CSR-style arrays, then reused in every pressure solver iteration. This requires exactly one `ImplicitIncompressibleSPHSystem`.
- The Jacobi iterations no longer recompute kernel gradients / `d_ij` values per iteration, removing essentially all of the pressure solver cost for the standard setup.
- Wall systems with `PressureBoundaries` are fully supported by the cached loop: the fluid–wall neighbor pairs are cached in both directions, so the boundary pressure contributions to the fluid `sum_d_ij_pj` and the wall `sum_term` reuse the cached coefficients while still reading the current (iteration-dependent) pressures. The wall pressures keep being updated by the same relaxed Jacobi scheme, so the coupled system is solved exactly as before.
- The pressure cache is built at the beginning of `update_implicit_sph!` (i.e., right before the solve), when the quantities of all systems — in particular the boundary densities recomputed in `update_quantities!` — are up to date.
- The generic neighbor-loop path is kept as a fallback (multiple IISPH systems) and is performance-neutral w.r.t. `main`.
- The cached coefficients are owned by the `Semidiscretization` (`semi.iisph_pressure_cache`) instead of the system struct. Storing 9 extra arrays inline in the (immutable) system struct measurably degrades codegen of the neighbor-loop closures (~8–15% slower `sum_term`/`predict_advection`), so the cache lives in the semidiscretization, which is insensitive to this.
- `inv_a_ii` is precomputed once per step and used by the relaxed Jacobi pressure update (also on the fallback path).

**Pressure solver instrumentation**
- `iisph_pressure_iteration_stats(semi)` / `reset_iisph_pressure_iteration_stats!(semi)` expose cumulative Jacobi iteration counts and pressure solve wall times.

### Performance

Full dam break (`fluid/dam_break_2d_iisph.jl` setup, `t = 1.41`, ≈1410 steps, 3200 fluid + 3912 boundary particles, 4 threads, Julia 1.11):

| Variant | Metric | main | this PR | Speedup |
|---|---|---:|---:|---:|
| IISPH (default) | total runtime | 3.67 s | 2.53 s | **1.45×** |
| IISPH (default) | pressure solver | 1.92 s | 0.094 s | **~20×** |
| IISPH + PressureBoundaries | total runtime | 5.05 s | 3.87 s | **1.30×** |
| IISPH + PressureBoundaries | pressure solver | 2.34 s | 0.12 s | **~19×** |

At 4× resolution (`fluid_particle_spacing = 0.6/80`, 12800 fluid + 7760 boundary particles) with the same `dt = 1e-3` as the 1× setup. Note that this time step is under-resolved at this resolution: the default setup hits `max_iterations = 30` (avg. 7.3 iterations per solve, i.e. some solves terminate unconverged):

| Variant | Metric | main | this PR | Speedup |
|---|---|---:|---:|---:|
| IISPH (default) | total runtime | 33.3 s | 11.9 s | **2.80×** |
| IISPH (default) | pressure solver | 25.2 s | 0.97 s | **~26×** |
| IISPH + PressureBoundaries | total runtime | 29.4 s | 15.4 s | **1.92×** |
| IISPH + PressureBoundaries | pressure solver | 19.1 s | 0.92 s | **~21×** |

With a CFL-scaled time step (`dt = 5e-4`, i.e. 2822 steps), the pressure solves converge again (avg. 2.2 / 2.0 iterations, no `max_iterations` hits) and the comparison is the properly resolved one. The pressure solver speedup is unchanged, while the total-runtime speedup is smaller since the remaining time is dominated by the (unmodified) generic neighbor loops, which scale with the number of steps:

| Variant | Metric | main | this PR | Speedup |
|---|---|---:|---:|---:|
| IISPH (default) | total runtime | 32.3 s | 22.5 s | **1.44×** |
| IISPH (default) | pressure solver | 16.7 s | 0.63 s | **~27×** |
| IISPH + PressureBoundaries | total runtime | 40.0 s | 29.7 s | **1.35×** |
| IISPH + PressureBoundaries | pressure solver | 20.1 s | 0.97 s | **~21×** |

Since the pressure solver cost is almost independent of the time step on this PR (0.6–1.0 s in all 4× runs), properly resolved time steps become affordable: halving `dt` mainly costs the re-evaluation of the remaining per-step quantities, not the pressure solve.

Further checks:
- Identical number of Jacobi solves and iteration statistics between main and this PR for every setup; results agree to floating-point rounding (trajectories are bit-identical for the first ~30 steps and then diverge only from accumulated roundoff).
- Zero allocations per RHS call (`count_rhs_allocations(sol) == 0` for all four dam-break IISPH example variants, also with split integration).
- Cache footprint: ~561 bytes per fluid particle (1.79 MB for 3200 particles), rebuilt once per time step.

### Testing

- 128 unit tests in `test/systems/iisph_system.jl` (including cached-vs-generic equivalence testsets with and without `PressureBoundaries`, and pressure cache ownership checks).
- All dam-break IISPH example tests pass with zero RHS allocations.
- Callback test suite (194 tests, incl. density reinit and split integration) passes.


svchb:
Some things are not fully fixed. Review again.

PR #1235, opened 2026-06-18
Fix interpolation with `VelocityAveraging`

efaulhaber:
Based on #1037.

When interpolating the initial state or after restarting from files (especially when interpolating from VTK files in postprocessing), velocity averaging is not initialized correctly, which leads to zero wall velocity even if the structure is moving. With this PR, interpolation from VTK files in postprocessing works as intended.


src/general/semidiscretization.jl
```diff
@@ -318,6 +321,11 @@ function semidiscretize(semi, tspan; reset_threads=true, restart_with=nothing)
     # Initialize all particle systems
     initialize!(semi_new, restart_with)
 
+    # Initialize the averaged velocity for TLSPH systems.
+    foreach_system(semi) do system
+        initialize_averaged_velocity!(system, v0_ode, semi, tspan[1])
+    end
```
svchb, line 327:
```suggestion
    foreach_system(semi_new) do system
        initialize_averaged_velocity!(system, v0_ode, semi_new, tspan[1])
    end
```
This initializes the adapted semidiscretization that is stored in `ode.p.semi`.

src/callbacks/update.jl
```diff
@@ -59,14 +59,6 @@ function initial_update!(cb::UpdateCallback, vu_ode, t, integrator)
...
 
-    # If TLSPH is not integrated, the averaged velocity will be initialized in the
-    # split integration.
-    if semi.integrate_tlsph[]
-        foreach_system(semi) do system
-            initialize_averaged_velocity!(system, v_ode, semi, t)
-        end
-    end
-
     return cb(integrator)
```
svchb, line 62:
```suggestion
    if semi.integrate_tlsph[]
        foreach_system(semi) do system
            initialize_averaged_velocity!(system, v_ode, semi, t)
        end
    end

    return cb(integrator)
```
This resets the averaged velocity when each integrator starts; [the regression test](https://github.com/trixi-framework/TrixiParticles.jl/pull/1235#issuecomment-5062693601) shows that a second regular integration otherwise keeps data from the first run.

src/callbacks/split_integration.jl
```diff
@@ -538,22 +538,9 @@ function UpdateAveragedVelocityCallback()
     # The first one is the `condition`, the second the `affect!`
     return DiscreteCallback(update_averaged_velocity_callback!,
                             update_averaged_velocity_callback!,
-                            initialize=(initialize_averaged_velocity_callback!),
                             save_positions=(false, false))
 end
```
svchb, line 542:
```suggestion
    return DiscreteCallback(update_averaged_velocity_callback!,
                            update_averaged_velocity_callback!,
                            initialize=(initialize_averaged_velocity_callback!),
                            save_positions=(false, false))
end

# `initialize`
function initialize_averaged_velocity_callback!(cb, vu_ode, t, integrator)
    v_ode, u_ode = vu_ode.x
    semi = integrator.p.semi

    foreach_system(semi) do system
        initialize_averaged_velocity!(system, v_ode, semi, t)
    end

    return cb
end
```
This resets the averaged velocity when each split integrator starts; [the regression test](https://github.com/trixi-framework/TrixiParticles.jl/pull/1235#issuecomment-5062693601) shows that a second split integration otherwise keeps the old velocity and time.

svchb:
Add `using OrdinaryDiffEqLowStorageRK` at the top of `test/schemes/structure/total_lagrangian_sph/velocity_averaging.jl`, then add this inside the existing `Velocity Averaging` test set:

```julia
@testset "Reinitialize averaged velocity for integration" begin
    algorithm = CarpenterKennedy2N54(williamson_condition=false)
    callbacks = (("update", () -> UpdateCallback()),
                 ("split", () -> SplitIntegrationCallback(algorithm; dt=0.01)))

    for (name, callback) in callbacks
        @testset "$name integration" begin
            velocity_averaging = VelocityAveraging(time_constant=0.5)
            system = velocity_averaging_test_system(; velocity_averaging)
            semi = Semidiscretization(system, neighborhood_search=nothing)
            ode = semidiscretize(semi, (0.0, 0.1))
            system = ode.p.semi.systems[1]

            TrixiParticles.SciMLBase.solve(ode, algorithm; dt=0.01, adaptive=false,
                                          callback=callback(), save_everystep=false)

            @test system.cache.t_last_averaging[] == last(ode.tspan)

            TrixiParticles.SciMLBase.init(ode, algorithm; dt=0.01, adaptive=false,
                                         callback=callback(), save_everystep=false)

            @test system.cache.t_last_averaging[] == first(ode.tspan)
            @test system.cache.averaged_velocity == [1.0 3.0 5.0
                                                     2.0 4.0 6.0]
        end
    end
end
```

This reproduces stale data for both regular and split integration: three assertions fail with the current changes, while all twenty-one tests pass when both initialization hooks are restored.

PR #1244, opened 2026-07-14
Fix `StepsizeCallback` for inviscid simulations

efaulhaber:
When no viscosity is used, there should be no `dt_viscosity` limitation. Instead, on main, the time step for inviscid simulations is often dominated by this spurious condition.

### DIFF
--- a/src/schemes/fluid/fluid.jl
+++ b/src/schemes/fluid/fluid.jl
@@ -202,9 +202,9 @@ function calculate_dt(v_ode, u_ode, cfl_number, system::AbstractFluidSystem, sem
     # TODO variable smoothing length
     smoothing_length_ = initial_smoothing_length(system)
 
-    dt_viscosity = 0.125 * smoothing_length_^2
+    dt_viscosity = Inf
     if !isnothing(system.viscosity)
-        dt_viscosity = dt_viscosity /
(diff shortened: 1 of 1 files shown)


src/schemes/fluid/fluid.jl
```diff
@@ -202,9 +202,9 @@ function calculate_dt(v_ode, u_ode, cfl_number, system::AbstractFluidSystem, sem
     # TODO variable smoothing length
     smoothing_length_ = initial_smoothing_length(system)
 
-    dt_viscosity = 0.125 * smoothing_length_^2
+    dt_viscosity = Inf
```
efaulhaber, line 205: **Low:** This changes the no-viscosity branch from a finite `0.125 * h^2` cap to `Inf`, but the diff adds no regression test that would fail if this regressed. Add a `calculate_dt` test for a fluid system with `viscosity === nothing` where the acceleration/sound-speed limits are larger than the old viscosity cap, so the expected timestep proves the viscosity limit is ignored only for inviscid systems.

efaulhaber (commented):
## Codex Review

The patch changes fluid timestep calculation so the viscosity constraint is ignored when no viscosity model is configured. I did not identify correctness or regression issues in the diff.

svchb (approved):

PR #1271, opened 2026-08-09
Add balanced CSS surface tension

svchb:
## Summary

- rebuild the balanced continuum-surface-stress (CSS) portion of the #1253 prototype as a focused layer
- evaluate `delta_s * (I - n * nᵀ)` directly in the symmetric fluid pair interaction instead of caching a per-particle stress tensor and subtracting a global maximum
- reuse the C1 color-gradient and complete-support activity introduced by #1269 for `SurfaceTensionMomentumMorris`
- accumulate the continuous support moment during existing fluid and boundary normal traversals and apply the symmetric correction `2 / (q_a + q_b)`
- retain exact pairwise linear-momentum conservation for constant smoothing length, including unequal particle masses, densities, and interface activities
- preserve VTK stress diagnostics by constructing the physical tensor only when writing output
- document the balanced operator, migration of `ideal_density_threshold`, and one-phase surface-delta normalization

## Stack

Depends on #1269, which depends on #1268 and #1252. This is the correct branch point for the CSS substack because #1269 provides the shared smooth interface activity and support-moment plumbing. It intentionally does not depend on #1270, whose C-CSF planar boundary-integral geometry is unrelated; #1270 and this PR are sibling layers.

Fork branches cannot be selected as bases for pull requests into the upstream repository, so this PR targets `main` and temporarily shows the preceding commits.

Review-only diff: https://github.com/svchb/TrixiParticles.jlOpen/compare/rebuild/morris-csf-interface...rebuild/balanced-css

## Scope

This first CSS layer covers one-phase colorfield geometry, smooth interface activity, the balanced fluid-fluid stress divergence, and wall/rigid support completion. Wetted-area/contact-line models, normal smoothing, interface-aware shifting/TIC, and other later CSS prototype features remain outside this PR. C-CSF boundary-integral geometry remains isolated in #1270.

## Verification

- aggregate fluid tests: all passed
- surface-tension tests: 116/116, including balanced CSS 21/21 and static Laplace balance 4/4
- CSS flat-pool wall geometry: 6/6
- RHS conservation tests: 1197/1197
- WCSPH system tests: 224/224
- EDAC system tests: 96/96
- VTK tests: 24/24 plus focused CSS tensor diagnostics
- Aqua passed
- documentation build passed
- complete unit target: 22875 passed; only the established baseline set remained (1 failed, 3 errored)
- JuliaFormatter 2.1.1 and `git diff --check` passed

PR #1270, opened 2026-08-09
Add C-CSF planar boundary geometry

svchb:
## Summary

- extend `CorrectedCSFSurfaceNormal` with an optional finite contact angle
- add the C-CSF planar boundary-integral moment, eigenvalue-gradient, color-gradient, and curvature terms from Vergnaud et al. (2022)
- apply the published distance-weighted contact-normal correction near the wall
- add explicit per-particle `surface_measure` quadrature data to dummy-particle boundaries and validate the supported configuration
- retain the existing no-contact free-surface path without boundary-only cache allocations
- add metadata, documentation, migration notes, and planar/hemispherical geometry regressions

## Stack

Depends on #1269, which depends on #1268 and #1252. Fork branches cannot be selected as bases for pull requests into the upstream repository, so this PR targets `main` and temporarily shows the preceding commits.

Review-only diff: https://github.com/svchb/TrixiParticles.jlOpen/compare/rebuild/morris-csf-interface...rebuild/ccsf-boundary-geometry

## Scope

This PR implements C-CSF boundary geometry only. It supports one fluid system with stationary planar dummy-particle `WallBoundarySystem`s, explicit face measures and normal offsets, and a 3D `WendlandC2Kernel`. Hydrodynamic continuity, pressure, and viscosity coupling still use the configured dummy-particle model; the full hydrodynamic BIM formulation is not included. Moving walls, rigid boundaries, general curved BIM faces, and ghost-particle C-CSF geometry are intentionally excluded.

## Verification

- C-CSF free-surface geometry tests: 18/18
- C-CSF planar boundary-geometry tests: 22/22
- C-CSF hemispherical-contact tests: 4/4
- C-CSF 3D curvature tests: 4/4
- aggregate surface-tension tests: 97/97
- WCSPH system tests: 224/224
- EDAC system tests: 96/96
- rigid-system tests: 182/182
- RHS tests: 1197/1197
- particle-packing tests: 8/8
- Aqua passed
- documentation build passed
- complete unit target: 22883 passed; only the established baseline set remained (1 failed, 3 errored)
- JuliaFormatter 2.1.1 and `git diff --check` passed

PR #1269, opened 2026-08-09
Add smooth Morris interface activity

svchb:
## Summary

- add C1 interface activation for Morris CSF with `ColorfieldSurfaceNormal`
- combine a tapered color-gradient indicator with a continuous complete-support kernel moment
- accumulate fluid, wall, and rigid-body support moments during the existing normal traversal without another neighbor pass
- weight the curvature stencil continuously by neighboring interface activity and handle vanishing normalization safely
- add configurable taper controls, metadata, documentation, migration notes, and focused regressions

## Stack

Depends on #1268, which in turn depends on #1252. Fork branches cannot be selected as bases for pull requests into the upstream repository, so this PR targets `main` and temporarily shows the preceding commits.

Review-only diff: https://github.com/svchb/TrixiParticles.jlOpen/compare/rebuild/ccsf-free-surface-core...rebuild/morris-csf-interface

## Scope

The smooth activity applies only to Morris CSF with `ColorfieldSurfaceNormal`; it does not alter the separate C-CSF geometry. Normal smoothing, shifting/TIC coupling, and a CSS stress-operator rewrite are intentionally excluded.

## Verification

- smooth-interface activity tests: 52/52
- rigid/wall support-moment tests: 19/19
- CSS/CSF sphere-normal tests: 199/199
- Akinci sphere-normal tests: 30/30
- rectangular-normal tests: 103/103
- aggregate surface-tension tests: 97/97
- WCSPH system tests: 224/224
- EDAC system tests: 96/96
- rigid-system tests: 182/182
- RHS tests: 1197/1197
- Aqua passed
- documentation build passed
- complete unit target: 22850 passed; only the established baseline set remained (1 failed, 3 errored)
- JuliaFormatter 2.1.1 and `git diff --check` passed

PR #1268, opened 2026-08-09
Add C-CSF free-surface core

svchb:
## Summary

- add `CorrectedCSFSurfaceNormal`, an explicit free-surface implementation of the corrected continuous-surface-force interface geometry from Vergnaud et al. (2022)
- compute normals from the renormalized minimum-eigenvalue gradient, curvature from the corrected divergence and thin-jet filter, and the surface delta with Shepard correction
- correct `SurfaceTensionMorris` to apply its local CSF acceleration once per particle while retaining the one-phase surface delta
- wire the local Morris acceleration into WCSPH and EDAC and validate the supported model combination
- add focused 2D/3D regressions, metadata, documentation, migration notes, and the Vergnaud citation

## Stack

Depends on #1252. Fork branches cannot be selected as bases for pull requests into the upstream repository, so this PR targets `main` and temporarily shows the preceding commits.

Review-only diff: https://github.com/svchb/TrixiParticles.jlOpen/compare/bug_fixed...rebuild/ccsf-free-surface-core

## Scope

This implements the single-fluid free-surface C-CSF core. Boundary-integral and contact-angle terms are intentionally not included.

## Verification

- C-CSF 2D geometry tests: 11/11
- C-CSF 3D curvature tests: 4/4
- surface-tension tests: 37/37
- WCSPH system tests: 224/224
- EDAC system tests: 96/96
- RHS tests: 1197/1197
- Aqua passed
- documentation build passed
- complete unit target: 22814 passed; only the established baseline set remained (1 failed, 3 errored)
- JuliaFormatter 2.1.1 and `git diff --check` passed

PR #1272, opened 2026-08-09
Add normal smoothing and VTK diagnostics

svchb:
## Summary

- rebuild the opt-in normal-smoothing layer from the #1253 prototype on top of balanced CSS #1271
- add `ColorfieldSurfaceNormal(normal_smoothing=true)` with an activity-weighted, same-fluid Shepard pass over unit normals
- skip inactive targets and neighbors, normalize finite weighted directions, and fall back to the raw normal when the weighted direction is undefined
- route Morris curvature/local force and balanced CSS stress evaluation through the capillary normal while retaining the raw normal, interface activity, and surface delta unchanged
- allocate smoothing caches and run the additional traversal only when explicitly enabled
- expand VTK output with the raw and capillary normals, surface delta, interface activity, model force, support moment or divergence correction, and an output-only physical stress tensor
- correct Morris VTK force output to use its particle-local acceleration rather than the now-empty pair-force path
- persist the smoothing choice in system metadata and document its scope

## Stack

Depends directly on #1271, the first balanced CSS layer. #1271 depends on #1269, #1268, and #1252. C-CSF planar boundary geometry #1270 remains a sibling substack and is not required here.

Fork branches cannot be selected as bases for pull requests into the upstream repository, so this PR targets `main` and temporarily shows the preceding commits.

Review-only diff: https://github.com/svchb/TrixiParticles.jlOpen/compare/rebuild/balanced-css...rebuild/css-normal-smoothing

## Scope

Smoothing is an explicit Morris CSF/CSS stabilization option. It does not change raw colorfield geometry, C-CSF geometry, interface activation, surface-delta magnitude, particle regularization, wetting/contact models, shifting, or tensile-instability control. The default path has no smoothing cache or neighbor traversal.

## Verification

- Shepard-smoothed CSS normal tests: 20/20
- surface-tension tests: 132/132
- CSS flat-pool geometry: 6/6
- RHS conservation tests: 1197/1197
- WCSPH system tests: 224/224
- EDAC system tests: 96/96
- VTK tests: 24/24 plus focused Morris/CSS diagnostic checks
- constant-normal invariance, inactive-particle preservation, zero-vector fallback, raw-geometry preservation, Morris force routing, and smoothed CSS stress reconstruction passed
- Aqua passed
- documentation build passed
- complete unit target: 22911 passed; only the established baseline set remained (1 failed, 3 errored)
- JuliaFormatter 2.1.1 and `git diff --check` passed


svchb:
Dependency note: after the lower stack is rebased onto #1282, this draft should use the shared raw normal and `surface_activity` data and retain only its capillary-normal smoothing caches and diagnostics. No additional intermediary stack branch is planned.

PR #1253, opened 2026-07-31
Surface Tension Prototype V1 

svchb:
## Status

This draft is the source umbrella for the original surface-tension prototype. Reviewable Akinci work has been rebuilt as a focused, physical-model-free stack directly on upstream `main`.

## Focused Stack

- #1258 surface-tension configuration hardening
- #1279 surface-normal scoping, explanation, and visualization
- #1280 multicolor gradients and threshold handling
- #1259 dimensionally consistent 2D Akinci kernels
- #1261 WCSPH and EDAC free-surface correction support
- #1266 neighbor-volume normalization for the Akinci normal force
- #1267 Akinci surface tension and correction support in IISPH
- #1262 aggregate implementation view
- #1263 cube-to-sphere validation and resolution study

PR #1260 is retired because the experimental physical-coefficient model is not part of the rebuilt stack.

## Review Topology

All focused pull requests target upstream `main` because GitHub cannot use branches from the contributor fork as upstream PR bases. Each stacked PR links a fork comparison that isolates its own layer. As earlier PRs merge, GitHub will automatically remove those commits from later cumulative diffs.

This umbrella remains open as historical context and is not the intended merge path.

PR #1252, opened 2026-07-30
Draft: combined pending PRs

svchb:
This draft branch currently merges:

- #1024
- ~#1058~
- #1060
- #1086
- #1087
- #1094
- #1145
- #1183 
- #1184
- #1187
- #1188
- #1190
- #1191
- #1196
- #1198
- #1199
- #1200
- #1202
- #1203
- #1213
- #1214
- #1215
- #1216
- #1217
- #1218

PR #1273, opened 2026-08-09
Add free-surface tangential shifting

svchb:
## Summary

- rebuild the opt-in free-surface shifting layer from the #1253 prototype on top of normal smoothing and diagnostics #1272
- add exported `FreeSurfaceTangentialShifting` support to callback-based Sun (2017) and consistent Sun (2019) particle shifting
- use the smooth interface activity to progressively remove the component along the raw surface normal while retaining full interior shifting
- clamp invalid activity ranges and preserve the original shifting velocity for non-finite activity or undefined normals
- validate the treatment for Morris CSF/CSS colorfield geometry and Morris C-CSF geometry in WCSPH and EDAC systems
- persist the selected treatment in system metadata and document the formulation, supported combinations, and limitations

## Stack

Depends directly on #1272. #1272 depends on #1271, #1269, #1268, and #1252. C-CSF planar boundary geometry #1270 remains a sibling substack and is not required here.

Fork branches cannot be selected as bases for pull requests into the upstream repository, so this PR targets `main` and temporarily shows the preceding commits.

Review-only diff: https://github.com/svchb/TrixiParticles.jlOpen/compare/rebuild/css-normal-smoothing...rebuild/free-surface-tangential-shifting

## Scope

Tangential shifting is an explicit regularization option. It does not change raw or smoothed capillary geometry, interface activity, surface-delta magnitude, surface-tension forces, TVF, wetting/contact models, or tensile-instability control. The default `free_surface_treatment=nothing` path retains the existing closed-system behavior.

## Verification

- shifting-technique tests: 52/52
- WCSPH system tests: 224/224
- EDAC system tests: 96/96
- RHS conservation tests: 1197/1197
- aggregate fluid suite passed
- disabled behavior, transition blending, non-unit and zero normals, invalid values, raw-versus-smoothed normal selection, constructor validation, and integrated treated-versus-untreated shifting passed
- Aqua passed
- documentation build and doctests passed
- complete unit target: 22950 passed; only the established baseline set remained (1 failed, 3 errored)
- JuliaFormatter 2.1.1 and `git diff --check` passed


svchb:
Dependency note: #1282 makes `surface_activity` a generic surface-method output. On rebase, tangential shifting should consume that signal and the normal capability directly instead of depending on Morris-specific activity plumbing. No additional intermediary stack branch is planned.

PR #1284, opened 2026-08-15
Fix SPH correction operator lifecycle

svchb:
## Summary

- split correction updates into global density, pressure, boundary, gradient, and surface phases
- route density, gradient, and force correction roles independently
- update WCSPH, EDAC, wall, rigid-body, and TLSPH systems through the shared lifecycle
- ensure correction results are independent of system declaration order
- split correction tests into focused lifecycle and method files

## Scope

This foundation intentionally excludes method-specific numerical hardening and conservative asymmetric pressure changes. Those are isolated in the draft follow-up PRs starting with #1292.

Isolated layer: 16 files, 520 additions, 40 deletions.

## Verification

- complete unit suite passed
- lifecycle ordering and boundary density/pressure tests passed
- JuliaFormatter and `git diff --check` passed


svchb:
Tested locally and reviewed. No performance regression. On 17.8.2026.

PR #1256, opened 2026-08-05
Add parallel intitialization for TLSPH arrays to ensure NUMA awareness

efaulhaber:
Benchmark on 2 x AMD EPYC 9965 (192 cores each) with 8 NUMA domains (48 cores each):
grafik
We can nicely see how this massive CPU is slow for small problems (when using all 384 cores), reaches maximum performance around 250k particles, then drops again when the problem size exceeds the cache, and then stagnates limited by the memory bandwidth.

The effect is much more pronounced for the deformation gradient:
grafik

With serial initialization, all data lives on one NUMA domain, decreasing memory locality and bandwidth. With parallel initialization, data lives on the NUMA domain belonging to the core that handles the corresponding particle.

PR #1294, opened 2026-08-17
Add parallel intitialization for TLSPH arrays to ensure NUMA awareness [Alternative]

efaulhaber:
This is a simpler alternative to #1256. I dislike that we use `PolyesterBackend` in this PR even if the user specified `SerialBackend` or a GPU backend (and doesn't expect us to use multiple threads), but the alternative is to initialize the arrays in `initialize!`, and that leads to a multitude of problems because `initial_coordinates` is accessed in several places before `initialize!` is called (see #1256). So I think this PR is still the cleaner solution.

Benchmark on 2 x AMD EPYC 9965 (192 cores each) with 8 NUMA domains (48 cores each):
grafik
We can nicely see how this massive CPU is slow for small problems (when using all 384 cores), reaches maximum performance around 250k particles, then drops again when the problem size exceeds the cache, and then stagnates limited by the memory bandwidth.

The effect is much more pronounced for the deformation gradient:
grafik

With serial initialization, all data lives on one NUMA domain, decreasing memory locality and bandwidth. With parallel initialization, data lives on the NUMA domain belonging to the core that handles the corresponding particle.


src/schemes/structure/total_lagrangian_sph/system.jl
```diff
@@ -145,10 +145,24 @@ function TotalLagrangianSPHSystem(initial_condition; smoothing_kernel, smoothing
...
+    initial_coordinates = similar(initial_condition_sorted.coordinates)
+    current_coordinates = similar(initial_condition_sorted.coordinates)
+    mass = similar(initial_condition_sorted.mass)
+    material_density = similar(initial_condition_sorted.density)
+
+    # Initialize the runtime arrays in parallel so that first-touch allocation places
+    # their memory close to the threads that will access it during the simulation.
+    # The semidiscretization backend is not available in the constructor yet, and using
+    # Polyester here is harmless for serial and GPU simulations.
+    parallelization_backend = PolyesterBackend()
```
efaulhaber, line 160: **Medium:** The parallel constructor initialization has no integration coverage, so replacing these calls with serial copies would still pass the existing value-based tests. Add an integration test or regression benchmark verifying that all four runtime arrays are initialized through the Polyester backend.

src/schemes/structure/total_lagrangian_sph/system.jl
```diff
@@ -145,10 +145,24 @@ function TotalLagrangianSPHSystem(initial_condition; smoothing_kernel, smoothing
...
-    initial_coordinates = copy(initial_condition_sorted.coordinates)
-    current_coordinates = copy(initial_condition_sorted.coordinates)
-    mass = copy(initial_condition_sorted.mass)
-    material_density = copy(initial_condition_sorted.density)
+    initial_coordinates = similar(initial_condition_sorted.coordinates)
+    current_coordinates = similar(initial_condition_sorted.coordinates)
+    mass = similar(initial_condition_sorted.mass)
+    material_density = similar(initial_condition_sorted.density)
+
+    # Initialize the runtime arrays in parallel so that first-touch allocation places
```
efaulhaber: **Low:** The comment overstates the NUMA-locality benefit because it applies specifically to compatible multithreaded CPU simulations, while serial execution uses one thread and GPU execution subsequently uses device memory. Qualify the comment to identify the benefiting backend and note that the initialization remains harmless elsewhere.

src/schemes/structure/total_lagrangian_sph/system.jl
efaulhaber, line 160: **Low:** This user-visible TLSPH initialization and NUMA-performance improvement is not documented in `NEWS.md`. Add an entry under an appropriate performance section, or explain why the change is intentionally not considered notable.

efaulhaber (commented):
## Codex Review

The PR replaces sequential copies of TLSPH runtime arrays with parallel initialization through `ThreadedBroadcastArray` and `PolyesterBackend` to improve NUMA first-touch placement. No concrete correctness or regression issues were found in this pass.

src/util.jl
```diff
@@ -301,3 +301,23 @@ function Base.similar(::Broadcast.Broadcasted{ThreadedBroadcastStyle{P}},
...
+    @threaded parallelization_backend for particle in axes(dest, 2)
+        for dimension in axes(dest, 1)
+            @inbounds dest[dimension, particle] = src[dimension, particle]
+        end
+    end
+
+    return dest
+end
+
+@inline function copyto_threaded!(dest::Vector, src, parallelization_backend)
```
efaulhaber, line 321: **High:** The helper only accepts `dest::Vector`, but `InitialCondition` preserves supported `AbstractVector` storage types such as `SVector`, for which `similar` may return a non-`Vector` destination and cause a `MethodError`. Accept appropriate abstract arrays or explicitly allocate ordinary arrays.

src/util.jl
```diff
@@ -301,3 +301,23 @@ function Base.similar(::Broadcast.Broadcasted{ThreadedBroadcastStyle{P}},
     # TODO we only have the type `P` here and just assume that we can do `P()`
     return ThreadedBroadcastArray(similar(Array{T}, dims), parallelization_backend=P())
 end
+
+# Like `copyto!`, but parallelize over particles to use the same memory access pattern
+# as the particle loops in the right-hand side.
+@inline function copyto_threaded!(dest::Matrix, src, parallelization_backend)
+    @threaded parallelization_backend for particle in axes(dest, 2)
```
efaulhaber, line 312: **High:** The loops prove bounds only for `dest`, while `@inbounds` also disables bounds checks for `src`; a smaller source can therefore turn a dimension mismatch into unsafe memory access. Validate that the relevant source and destination axes match before both threaded loops.

efaulhaber, line 312: **Medium:** Existing constructor tests check only copied values and would pass if the helpers reverted to serial `copyto!`, leaving the PR's threaded behavior unverified. Add matrix and vector tests with a counting or failing backend, similar to the existing `ThreadedBroadcastArray` tests, to assert that `@threaded` is invoked and the output remains correct.

efaulhaber (commented):
## Codex Review

The PR intends to improve NUMA performance by parallelizing first-touch initialization of TLSPH runtime arrays.

svchb (approved):

PR #1301, opened 2026-08-24
Add option to update `DensityDiffusionAntuono` between time steps

efaulhaber:
This strategy is explained in Equation (4.29) of Antuono et al., "Numerical diffusive terms in weakly-compressible SPH schemes" (2012).


src/callbacks/update.jl
```diff
@@ -111,8 +111,8 @@ function (update_callback!::UpdateCallback)(integrator)
         end
 
         foreach_system(semi) do system
-            particle_shifting_from_callback!(u_ode, shifting_technique(system), system,
-                                             v_ode, semi, integrator)
+            update_density_diffusion_from_callback!(system, density_diffusion(system),
```
efaulhaber, line 114: **High:** When `update_everystage=false`, this callback changes data used by the RHS but does not report a derivative discontinuity, allowing FSAL and derivative-caching integrators to reuse a derivative computed with the previous or initial gradient. Call `derivative_discontinuity!(integrator, true)` after updating the cache.

src/schemes/fluid/weakly_compressible_sph/density_diffusion.jl
```diff
@@ -112,17 +119,30 @@ L_a := \left( -\sum_{b} V_b r_{ab} \otimes \nabla W_{ab} \right)^{-1} \in \R^{d
...
 See [`AbstractDensityDiffusion`](@ref TrixiParticles.AbstractDensityDiffusion)
 for an overview and comparison of implemented density diffusion terms.
 """
-struct DensityDiffusionAntuono{ELTYPE} <: AbstractDensityDiffusion
+struct DensityDiffusionAntuono{ELTYPE, UPDATE_EVERYSTAGE} <: AbstractDensityDiffusion
     delta::ELTYPE
 
-    function DensityDiffusionAntuono(; delta)
-        new{typeof(delta)}(delta)
+    function DensityDiffusionAntuono(; delta, update_everystage=true)
```
efaulhaber: **Medium:** A non-Boolean value such as `update_everystage=1` is accepted as a type parameter but matches neither the `{true}` nor `{false}` methods, silently disabling the normalized-gradient correction. Constrain the keyword to `Bool` so invalid modes are rejected.

test/schemes/fluid/weakly_compressible_sph/density_diffusion.jl
```diff
@@ -4,7 +4,15 @@
...
 
-            @test repr(density_diffusion) == "DensityDiffusionAntuono(0.1)"
+            @test repr(density_diffusion) ==
+                  "DensityDiffusionAntuono(delta=0.1, update_everystage=true)"
+
+            callback_density_diffusion = DensityDiffusionAntuono(delta=0.1,
+                                                                 update_everystage=false)
+            @test repr(callback_density_diffusion) ==
+                  "DensityDiffusionAntuono(delta=0.1, update_everystage=false)"
+            @test !TrixiParticles.requires_update_callback(density_diffusion)
```
efaulhaber, line 14: **Medium:** These assertions test `requires_update_callback` only on the diffusion object, while runtime validation invokes it on the complete system; the integration test always supplies an `UpdateCallback`, so broken system delegation would go unnoticed. Add a configuration with `update_everystage=false` and no `UpdateCallback`, then assert that solving raises the expected `ArgumentError`.

test/examples/examples_fluid.jl
```diff
@@ -553,6 +553,25 @@
...
+                                         joinpath(examples_dir(), "fluid",
+                                                  "vortex_street_2d.jl");
+                                         tspan=(0.0, 0.015), density_diffusion,
+                                         saving_callback=nothing)
+        @test sol.retcode == ReturnCode.Success
+        @test count_rhs_allocations(sol) == 0
+
+        # The final time t = 0.015 is right after activating inactive particles.
+        # The following test fails if the DD is not updated AFTER the open boundaries.
+        normalized_density_gradient = fluid_system.cache.density_diffusion_normalized_density_gradient
```
efaulhaber, line 569: **Medium:** The nonzero-gradient assertion confirms that the callback computes the cache but does not verify that ordinary RHS-stage updates are skipped when `update_everystage=false`. Add a focused test showing that the cache remains unchanged through the stage-update path and changes only when the callback runs.

efaulhaber (commented):
## Codex Review

The patch adds an option to update Antuono density-diffusion data between time steps through `UpdateCallback`, including documentation, tests, and initialized caches.

PR #1305, opened 2026-09-01
Fix docs @ref for Documenter 1.18

svchb:
Unbackticked `[WindingNumberHormann](@ref)` / `[WindingNumberJacobson](@ref)` in `docs/src/preprocessing/preprocessing.md:18-19` no longer resolve to docstrings under Documenter 1.18 (header-only).


svchb:
merged via #1303

PR #1308, opened 2026-09-02
Fix example `complex_shape_3d.jl`

LasNikas:


svchb:
Please add the example to the CI.

